PR #22, opened 2021-12-20
Add to cart functionality

Flanker72:
ST-139: Add to cart functionality from Category page view.
Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/client-app-0.0.1-pr-22-374af597.zip


client-app/pages/catalog/catalog.vue
```diff
@@ -375,6 +384,14 @@ watch(isMobile, async () => {
     }
   }
 });
+
+const isCartAddInfoOpen = ref(false);
+const cartAddInfoLineItem = ref();
+
+const onAddToCart = function (lineItem?: LineItemType) {
```
ilyawzrd, line 391: Can we have it as an arrow function? Just to keep everything the same way

ilyawzrd (changes requested):

PR #44, opened 2022-02-01
ST-293: Implement Checkout Defaults user settings

trueboroda:
Description:

--

QA-test:

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/client-app-0.0.1-pr-44-8be3b279.zip


client-app/components/molecules/select/select.vue
```diff
@@ -48,7 +48,7 @@
               @click="select(item)"
             >
               <slot name="item" v-bind="{ item, index, selected }">
-                {{ itemTextKey ? item[itemTextKey] : item }}
+                {{ itemTextKey && item ? item[itemTextKey] : item || "&nbsp;" }}
```
artmakarov, line 51: Remove ` || "&nbsp;"`
This logic is at the mercy of the developer. To help `item` slot
Example in lines 43-46 `checkout-defaults.vue`

client-app/pages/account/checkout-defaults.vue
```diff
@@ -0,0 +1,110 @@
...
+    result.push(...cart.value.availablePaymentMethods);
+  }
+
+  return result;
+});
+
+const { getUserCheckoutDefaults, setUserCheckoutDefaults } = useUserCheckoutDefaults();
+const { openPopup } = usePopup();
+
+let checkoutDefaults: Ref<CheckoutDefaults> = ref<CheckoutDefaults>({});
```
Flanker72, line 95: Why this is `let`? Refs are objects and it's better to use `const` instead.

client-app/shared/account/composables/useUserCheckoutDefaults.ts
```diff
@@ -0,0 +1,20 @@
...
+  const { me } = useUser();
+  const keyPrefix = "checkout_defaults_";
+
+  function setUserCheckoutDefaults(defaults: CheckoutDefaults) {
+    localStorage.setItem(`${keyPrefix}${me.value.id}`, JSON.stringify(defaults));
+  }
+
+  function getUserCheckoutDefaults(): CheckoutDefaults | null {
+    const value = localStorage.getItem(`${keyPrefix}${me.value.id}`);
+    return value == null ? null : (JSON.parse(value) as CheckoutDefaults);
```
Flanker72, line 13: It's better to wrap JSON.parse with try-catch block to avoid unhandled exceptions.

client-app/components/atoms/radio-button/radio-button.vue
```diff
@@ -0,0 +1,40 @@
...
+    />
+    <label :for="id" class="text-base font-medium" :class="{ 'text-gray-500': !isChecked }">{{ label }}</label>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { computed } from "vue";
+
+const props = defineProps({
+  id: {
```
Flanker72, line 19: Do we really need to pass an id here? Is this used only for linking input with label?

client-app/components/atoms/radio-button/radio-button.vue
```diff
@@ -0,0 +1,40 @@
...
+  <div class="flex flex-row items-center space-x-2">
+    <input
+      :id="id"
+      class="appearance-none rounded-full h-6 w-6 border-2 border-gray-300 bg-white checked:border-8 checked:border-yellow-500 focus:outline-none cursor-pointer"
+      type="radio"
+      :value="value"
+      :checked="isChecked"
+      @change="$emit('update:modelValue', value)"
+    />
+    <label :for="id" class="text-base font-medium" :class="{ 'text-gray-500': !isChecked }">{{ label }}</label>
```
Flanker72, line 11: Maybe it's better to wrap label in named slot to allow passing icons and other dynamic content?

client-app/components/atoms/radio-button/radio-button.vue
trueboroda: Only for that.

Flanker72: Maybe its better to generate unique id dynamically inside the `setup` to simplify component usage

Flanker72 (approved):

artmakarov (approved):

PR #60, opened 2022-02-22
ST-283: Add profile page

ilyawzrd:
Description:

--

QA-test: ST-286

Demo-test: ST-287

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/client-app-0.0.4-pr-60-2072825e.zip

### DIFF
--- a/client-app/pages/account/profile.vue
+++ b/client-app/pages/account/profile.vue
@@ -16,9 +16,6 @@
             <form @submit="onSubmit" class="lg:w-1/2 flex flex-col">
               <!-- Errors block -->
               <VcAlert v-if="updateProfileError" class="mb-2">There was an error during profile update</VcAlert>
-              <VcAlert v-if="!_.isEmpty(errors)" class="mb-2">
-                <strong>First name</strong> and <strong>Last name</strong> are required
-              </VcAlert>
 
               <VcInput
                 v-model="firstName"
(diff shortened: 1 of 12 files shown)


client-app/pages/account/profile.vue
```diff
@@ -0,0 +1,132 @@
...
+                :error-message="errors.lastName"
+              ></VcInput>
+
+              <VcInput
+                v-model="email"
+                name="email"
+                class="mb-4"
+                label="Email"
+                placeholder="Email"
+                :is-disabled="true"
```
Flanker72, line 49: This can be done shortly just using `is-disabled` - if prop exists, it's `true` by default. Just like `is-required`, `is-submit`, etc.

Flanker72 (approved):

client-app/shared/layout/components/header/_internal/mobile-menu-link.vue
```diff
@@ -5,7 +5,7 @@
       class="uppercase text-xl font-extrabold text-white flex items-center"
       @click="submenuVisible = !submenuVisible"
     >
-      <div>
+      <div class="overflow-hidden overflow-ellipsis">
```
Flanker72, line 8: I think here you can use `truncate` class instead of these two

PR #70, opened 2022-03-04
ST-180: Added bulk order pad page

artmakarov:
Description:

--

QA-test: ST-183

Demo-test: ST-184

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/client-app-0.0.4-pr-70-0f28e938.zip


client-app/components/atoms/alert/alert.vue
```diff
@@ -1,6 +1,84 @@
...
+      <slot />
+    </div>
   </div>
 </template>
+
+<script setup lang="ts">
+import { computed, PropType } from "vue";
+
+const props = defineProps({
+  text: Boolean,
```
Flanker72, line 14: It's too strange to make boolean `text` property. Maybe it's better to make `text` one of `type` enum.

client-app/components/molecules/text-area/text-area.vue
```diff
@@ -21,8 +23,11 @@ export default {
 import { computed } from "vue";
 
 const props = defineProps({
+  counter: Boolean,
```
Flanker72, line 26: Maybe better to use `hasCounter` or `showCounter` to be more intuitive.

Flanker72 (approved):

PR #87, opened 2022-03-23
ST-495: Compare products

artmakarov:
Description:

--

QA-test: ST-498

Demo-test: ST-499

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/client-app-0.0.4-pr-87-ea2558b2.zip

### DIFF
--- a/client-app/core/utilities/helpers/index.ts
+++ b/client-app/core/utilities/helpers/index.ts
@@ -1,3 +1,7 @@
 export function sleep(ms: number, resolvedValue?: any): Promise<any> {
   return new Promise((resolve) => setTimeout(resolve, ms, resolvedValue));
 }
+
+export function truncate(str: string, length: number): string {
+  return str.length > length ? `${str.slice(0, length)}...` : str;
+}
--- a/client-app/shared/compare/components/add-to-compare.vue
+++ b/client-app/shared/compare/components/add-to-compare.vue
@@ -1,0 +1,38 @@
+<template>
+  <VcCheckbox v-if="workaround" :model-value="isInCompareList" @change="toggle">
+    {{ $t("common.labels.compare") }}
+  </VcCheckbox>
+</template>
+
+<script setup lang="ts">
+import { nextTick, PropType, ref } from "vue";
+import { VcCheckbox } from "@/components";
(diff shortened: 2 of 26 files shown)


client-app/core/utilities/helpers/index.ts
```diff
@@ -1,3 +1,7 @@
 export function sleep(ms: number, resolvedValue?: any): Promise<any> {
   return new Promise((resolve) => setTimeout(resolve, ms, resolvedValue));
 }
+
+export function truncate(str: string, length: number): string {
+  return str.length > length ? `${str.slice(0, length)}...` : str;
```
Flanker72, line 6: Maybe it's better to use `&hellip;` instead of three dots?

client-app/shared/compare/components/add-to-compare.vue
```diff
@@ -0,0 +1,38 @@
...
+const props = defineProps({
+  product: {
+    type: Object as PropType<Product>,
+    required: true,
+  },
+});
+
+const { productsIds, addToCompareList, removeFromCompareList } = useCompareProducts();
+
+const workaround = ref(true); // See lines 32-34
```
Flanker72, line 23: Wrong line numbers, should be 34-36 instead

client-app/core/utilities/helpers/index.ts
artmakarov: `&hellip;` will only work when HTML is output (`v-html`)

trueboroda (approved):

client-app/shared/notification/composables/useNotifications.ts
```diff
@@ -0,0 +1,78 @@
+import { readonly, shallowRef, triggerRef } from "vue";
+import _ from "lodash";
+import { INotification, INotificationExtended } from "@/shared/notification";
+
+const stack = shallowRef<INotificationExtended[]>([]);
+
+async function show(options: INotification) {
```
Flanker72, line 7: If we have "show" method then we assume to have a "hide" method in opposite.
However we have "close" method, so we assume to have "open" method in opposite.

Maybe we should change naming to be more convenient?

Flanker72 (approved):

PR #145, opened 2022-05-19
ST-1154: Using  slugs for both products and categories

artmakarov:
Description:

--

QA-test: ST-1156

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.1.0-pr-145-56515456.zip


client-app/pages/search/search.vue
```diff
@@ -119,6 +120,13 @@ const searchParams = computed<ProductsSearchParams>(() => ({
   keyword: keywordQueryParam.value,
 }));
 
+const productsRoutes = computed(() =>
```
Flanker72, line 123: Why not to move this snippet to shared file? We reuse this code in every product-related route.

Flanker72 (approved):

PR #179, opened 2022-06-13
ST-2072: Refactor theme context functionality

artmakarov:
Description:

--

QA-test: ST-2075

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.3.0-pr-179-627fdf46.zip


client-app/shared/layout/composables/useNavigations.ts
```diff
@@ -17,7 +15,7 @@ const mainMenuLinks = computed<MenuLink[]>(() =>
       ({
         id: item.id,
         route: item.route,
-        title: globals.i18n!.global.t(item.title),
+        title: globals.i18n.global.t(item.title),
```
Flanker72, line 18: I think this should be a bit different, it's not clear how to use this. Maybe we should have an ability to use `useI18n` here, or use i18n globals from App instance, or pass data as-is and localize it within the template later.

client-app/shared/layout/composables/useNavigations.ts
```diff
@@ -17,7 +15,7 @@ const mainMenuLinks = computed<MenuLink[]>(() =>
...
         children: (menuLinkLists.value?.[item.id] || []).map((childrenItem) => ({
           id: childrenItem.url?.split("/").pop(),
           title: childrenItem.title,
```
Flanker72, line 21: Addition to the above comment: we have translation for parent item, however child item is passed as non-translated. This should be somehow synchronized.

PR #233, opened 2022-07-11
feat(ST-1728): implement wishlists popup

goldenmaya:
Description:
https://virtocommerce.atlassian.net/browse/ST-1728

--

QA-test:

Demo-test:

Download artifact URL:


client-app/shared/wishlists/components/add-to-wishlists-dialog.vue
```diff
@@ -1,70 +1,135 @@
...
+            <VcInput
+              class="flex-grow ml-2.5 mr-3.5"
+              :class="{ 'mb-4.5': !input.errorMessage }"
+              v-model="input.listName"
+              :is-disabled="loading"
+              :maxlength="25"
+              is-required
+              :error-message="input.errorMessage"
+            ></VcInput>
+            <button class="pt-3.5" @click="removeInput(index)">
```
Flanker72, line 67: With padding the button has a strange clickable area: the X mark itself and some empty space above. Maybe it's better to use margin instead?

goldenmaya: done

client-app/shared/wishlists/components/add-to-wishlists-dialog.vue
```diff
@@ -73,44 +138,131 @@ const props = defineProps({
...
 
 const loading = ref(false);
-const selectedListsIds = ref<string[]>([]);
+const selectedListsOtherIds = ref<string[]>([]);
+const listsRemove = ref<string[]>([]);
+const inputs = ref<WishlistInputType[]>([]);
+
+const productId = props.product.id;
+const config = inject(configInjectionKey);
+const listsLimit = config?.wishlists_limit || 10;
```
Flanker72, line 160: It's preferred to move this to core/constants - magic numbers hidden in the code are not so clear.

client-app/shared/wishlists/components/add-to-wishlists-dialog.vue
```diff
@@ -73,44 +138,131 @@ const props = defineProps({
...
   loading.value = false;
 
-  openPopup({
-    component: AddedToWishlistsDialog,
-    props: {
-      product: props.product,
-      listIds: selectedListsIds.value,
-    },
+  notifications.success({
+    duration: 15000,
```
Flanker72, line 250: Looks like this should be moved to core/constants - this interval should be tuned from the single place.

tailwind.config.js
```diff
@@ -130,13 +130,34 @@ module.exports = {
...
+            lineHeight: "1.5rem",
+          },
+        ],
+      },
+      maxHeight: {
+        "screen-60": "60vh",
+        "screen-75": "75vh",
+      },
+      margin: {
+        4.5: "1.125rem", //18px
```
Flanker72, line 160: Need to recheck the design. Do we really need this margin? Can we use 16 or 20px instead?

Flanker72 (changes requested):
Some minor changes needed

tailwind.config.js
goldenmaya: we have 3.5 by default, so why don't we also have 4.5 by default.

goldenmaya: this margin is used to prevent jerking when an error message appears

client-app/shared/wishlists/components/add-to-wishlists-dialog.vue
goldenmaya: done

client-app/shared/wishlists/components/add-to-wishlists-dialog.vue
goldenmaya: done

PR #285, opened 2022-09-01
fix(ST-2707): trim text inputs value. use VcInput everywhere it posible

trueboroda:
Description:

Find all input elements without VcInput component usage. 

use VcInput if it is possible.

use the ‘trim’ modifier in other cases. 

Fix VcInput. Don’t use trimming to type 'password'

check the trimming algoritm on possible improvements 

--

QA-test:  ST-2709

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.8.0-pr-285-6ac1dc97.zip

### DIFF
--- a/client-app/ui-kit/components/atoms/input/vc-input.vue
+++ b/client-app/ui-kit/components/atoms/input/vc-input.vue
@@ -7,8 +7,8 @@
 
     <div class="relative h-11">
       <input
-        class="appearance-none rounded h-full px-3 text-base leading-none box-border border border-gray-300 w-full outline-none focus:border-gray-400 min-w-0"
-        :class="{ 'pr-12': isPasswordIconVisible }"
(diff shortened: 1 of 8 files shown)


client-app/ui-kit/components/atoms/input/vc-input.vue
```diff
@@ -50,6 +53,7 @@ const props = defineProps({
   isReadonly: Boolean,
   isDisabled: Boolean,
   isRequired: Boolean,
+  isBordered: { type: Boolean, default: true },
```
artmakarov, line 56: Maybe `noBorder: Boolean` or `withoutBorder: Boolean`

To make it so:
  `<VcInput no-border />` or `<VcInput without-border />`

artmakarov (approved):

PR #508, opened 2023-04-03
feat: leave feedback

asvishnyakov:
Description:
Depends on
- [ ] #505

--

QA-test:

Demo-test:

Download artifact URL:


client-app/xapi/graphql/vendor/mutations/createCustomerReview/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  CreateCustomerReviewCommandType,
+  "storeId" | "userId" | "userName"
+>;
+
+export async function createCustomerReview(
+  commandArguments: CreateCustomerReviewCommandArgumentsType
+): Promise<CustomerReview> {
+  const { storeId, userId } = globals;
+
+  const { user } = useUser();
```
artmakarov, line 21: It shouldn't be here

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vc-theme-b2b-vue",
-  "version": "1.24.0",
+  "version": "1.25.0",
```
artmakarov, line 3: Roll back

client-app/xapi/graphql/vendor/mutations/createCustomerReview/index.ts
asvishnyakov: But I need to pass user name (full real name, not email). So what you will propose? Pass it as parameter? Why?

PR #569, opened 2023-05-23
PT-12181, PT-12175: add search results for suggestions and static pages

ksavosteev:
Description:
https://virtocommerce.atlassian.net/browse/ST-4480

--

QA-test:

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.32.0-pr-569-8b8f-8b8fe876.zip


client-app/shared/layout/components/search-bar/search-bar.vue
```diff
@@ -29,6 +29,39 @@
       >
         <!-- Results -->
         <template v-if="categories.length || products.length">
```
artmakarov, line 31: `v-if="categories.length || products.length || suggestions.length || pages.length"`
or for example
`v-if="isExistResults"`
`...`
`const isExistResults = computed(() => categories.value.length || products.value.length || suggestions.value.length || pages.value.length)`

PR #659, opened 2023-08-01
fix: quote form validation

goldenmaya:
Description:
https://virtocommerce.atlassian.net/browse/ST-4299

--

QA-test:

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.35.0-pr-659-873f-873f537a.zip


client-app/pages/account/edit-quote.vue
```diff
@@ -209,6 +200,22 @@ const quoteValid = computed<boolean>(
 
 const userHasAddresses = computedEager<boolean>(() => !!accountAddresses.value.length);
 
+const isQuoteCanBeSaved = computed<boolean>(
```
asvishnyakov, line 203: I think you can reuse `quoteValid` here to simplify this overcomplicated logic

asvishnyakov (changes requested):

Andrew-Orlov (approved):

client-app/pages/account/edit-quote.vue
goldenmaya: @asvishnyakov we can not reuse quoteValid, because we have to detect changes and we can save even invalid quote.

asvishnyakov (approved):

PR #720, opened 2023-08-31
feat(company-info): implement new settings dropdown design

goldenmaya:
Description:
https://virtocommerce.atlassian.net/browse/ST-4661

--

QA-test:

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.39.0-pr-720-fcb2-fcb2ca3f.zip

### DIFF
--- a/client-app/pages/company/info.vue
+++ b/client-app/pages/company/info.vue
@@ -133,7 +133,12 @@
                 </div>
 
                 <div v-if="userCanEditOrganization" class="absolute right-4 top-3">
-                  <VcActionDropdownMenu placement="left-start" :actions="getActionsForItem(item)" />
+                  <CompanyInfoDropdownMenu
+                    :address="item"
+                    placement="left-start"
+                    @edit="openAddOrUpdateCompanyAddressModal(item)"
+                    @delete="openDeleteAddressModal(item)"
+                  />
                 </div>
               </div>
             </template>
@@ -195,7 +200,11 @@
                 </td>
 
                 <td v-if="userCanEditOrganization" class="relative px-5 py-3 text-right">
-                  <VcActionDropdownMenu :actions="getActionsForItem(address)" />
+                  <CompanyInfoDropdownMenu
+                    :address="address"
+                    @edit="openAddOrUpdateCompanyAddressModal(address)"
+                    @delete="openDeleteAddressModal(address)"
+                  />
                 </td>
               </tr>
             </template>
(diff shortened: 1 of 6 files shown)


client-app/pages/company/info.vue
```diff
@@ -195,27 +220,24 @@
...
-                        :class="['mr-1', { 'text-[--color-danger-500]': !address.isDefault }]"
-                        name="x"
-                        :size="18"
-                      />
-                      <span class="text-sm">{{ $t("common.buttons.delete") }}</span>
+                      <VcIcon :class="{ 'text-[--color-danger-500]': !address.isDefault }" name="delete-2" size="sm" />
+
+                      {{ $t("common.buttons.delete") }}
                     </button>
                   </VcActionDropdownMenu>
```
asvishnyakov, line 242: Reuse this code and don't forget about permissions checking

asvishnyakov (changes requested):

client-app/pages/company/info.vue
```diff
@@ -76,7 +76,6 @@
         <div v-else class="flex flex-col md:rounded md:border">
           <VcTable
             :loading="loadingAddresses"
-            :item-actions-builder="itemActionsBuilder"
```
asvishnyakov, line 79: Remove no longer used variable `itemActionsBuilder` from typescript

tailwind.config.ts
```diff
@@ -1,5 +1,6 @@
 import { BREAKPOINTS } from "./client-app/core/constants/tailwind";
 import type { Config } from "tailwindcss";
+import {deprecate} from "util";
```
asvishnyakov, line 3: You're using jsdoc `@deprecated` and don't need to import this node.js function

asvishnyakov (changes requested):

asvishnyakov (changes requested):
That's definitely not that I mean when I asked for reused code. I simply mean you should create shared component for that VcActionDropDownMenu with it's items and then use it twice in code. I think refactoring of the VcActionDropDownMenu should be part of UI Kit work and it should be done in the other way, like https://vuetifyjs.com/en/components/menus/ so you can be able to use component-based approach to build the menu, instead of passing classes, event handlers and other stuff as props. So it should be a separate task with mitigation about UI kit first, especially because we have the same actions on other pages and also other component which using VcActionDropDownMenu

NaMax66 (approved):

client-app/ui-kit/components/atoms/action-dropdown-menu/vc-action-dropdown-menu.vue
```diff
@@ -1,3 +1,4 @@
+<!-- @deprecated the component will be removed from ui-kit -->
```
asvishnyakov, line 1: https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/c31ea77061d7407bf6b1cb385c37c7a8c190061c/client-app/ui-kit/components/atoms/types.d.ts#L7
Put it here instead please

asvishnyakov (changes requested):

asvishnyakov (approved):

PR #723, opened 2023-09-05
fix: suppress v-html warnings

Andrew-Orlov:
Description:

https://virtocommerce.atlassian.net/browse/ST-4906

--

QA-test:

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.39.0-pr-723-a067-a067a2f5.zip


client-app/shared/static-content/components/text-block.vue
NaMax66, line 1: It goes from the page builder. What will happen if we add some script there?

NaMax66 (approved):

asvishnyakov (approved):

PR #731, opened 2023-09-06
Partially added support for French and German

NaMax66:
Description:
https://virtocommerce.atlassian.net/browse/ST-4913

--
Also fix edit language from the platform
Fix performance issue with separate language load

Before:
before

After:
after

QA-test:

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.39.0-pr-731-fc17-fc172326.zip

### DIFF
--- a/client-app/app-runner.ts
+++ b/client-app/app-runner.ts
@@ -26,19 +26,27 @@ export default async (getPlugins: (options: { router: Router }) => { plugin: Plu
 
   const { fetchUser } = useUser();
   const { themeContext, fetchThemeContext } = useThemeContext();
-  const { currentLocale, currentLanguage, supportedLocales, setLocale } = useLanguages();
+  const { currentLocale, currentLanguage, supportedLocales, setLocale, fetchLocaleMessages } = useLanguages();
   const { currentCurrency } = useCurrency();
+
+  const fallback = {
+    locale: "en",
+    message: {},
+    async setMessage() {
+      this.message = await fetchLocaleMessages(this.locale);
+    },
+  };
 
   /**
    * Fetching required app data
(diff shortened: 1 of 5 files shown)


client-app/app-runner.ts
```diff
@@ -26,19 +26,27 @@ export default async (getPlugins: (options: { router: Router }) => { plugin: Plu
...
+    async setMessage() {
+      this.message = await fetchLocaleMessages(this.locale);
+    },
+  };
+
   /**
    * Fetching required app data
    */
-  await Promise.all([fetchThemeContext(), fetchUser()]);
+  await Promise.all([fetchThemeContext(), fetchUser(), fallback.setMessage()]);
```
asvishnyakov, line 43: I propose to expose `fetchFallbackLocale and fallbackLocale` from `useLanguages`, call it here, and pass to i18n.

client-app/i18n.ts
```diff
@@ -35,6 +43,12 @@ export function createI18n(locale: string, currency: string) {
         },
       },
     },
+    fallbackLocale: fallback?.locale,
+    messages: {
+      ...fallbackMessage,
+    },
+    fallbackWarn: false,
+    missingWarn: false,
```
asvishnyakov, line 51: While `fallbackWarn` is discussable, `missingWarn` shouldn't be set to `false`, because in this case you will not receive warning about missed locale until you will see (if you will see) ugly localization key in UI

asvishnyakov (changes requested):

asvishnyakov (approved):

PR #736, opened 2023-09-11
fix: navigation semantic structure

NaMax66:
Description:
https://virtocommerce.atlassian.net/browse/ST-4604

--

QA-test:

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.39.0-pr-736-25b0-25b08322.zip

### DIFF
--- a/client-app/shared/layout/components/header/_internal/bottom-header.vue
+++ b/client-app/shared/layout/components/header/_internal/bottom-header.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="relative">
-    <div
+    <nav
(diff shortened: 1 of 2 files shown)


asvishnyakov (changes requested):
@goldenmaya Look at this please, I'm feeling something like BottomHeaderLink or MobileMenuLink probably should be <li> itself and then use something like container-component for <ul>, but I'm not sure and this solution is probably ok as is

asvishnyakov (approved):

PR #755, opened 2023-09-19
fix: error notification in cart

NaMax66:
Description:
https://virtocommerce.atlassian.net/browse/ST-4989

--

QA-test:

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.40.0-pr-755-cef2-cef278ba.zip

### DIFF
--- a/client-app/shared/cart/components/add-to-cart.vue
+++ b/client-app/shared/cart/components/add-to-cart.vue
@@ -92,10 +92,7 @@ const inputElement = shallowRef<HTMLInputElement>();
 
 const isDigital = computed<boolean>(() => props.product.productType === ProductType.Digital);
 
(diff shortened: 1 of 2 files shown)


asvishnyakov (changes requested):
You don't need to return values from composable, approach is to have some properties and methid which mutate them, so if you need cart, then just use `cart.value` after calling `addToCart` or `changeItemQuantity`

NaMax66:
> You don't need to return values from composable, approach is to have some properties and methid which mutate them, so if you need cart, then just use `cart.value` after calling `addToCart` or `changeItemQuantity`

Other method mutates `cart` when we have the result of async `addToCart`

asvishnyakov (approved):

PR #760, opened 2023-09-22
PageBuilder. Add "alt" attribute to the "Call to action with image" section

NaMax66:
Description:
https://virtocommerce.atlassian.net/browse/ST-5033

--

QA-test:

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.40.0-pr-760-fa96-fa963117.zip

### DIFF
--- a/client-app/shared/static-content/components/call-to-action-with-image.vue
+++ b/client-app/shared/static-content/components/call-to-action-with-image.vue
@@ -31,7 +31,7 @@
 import type { RouteLocationRaw } from "vue-router";
 
 type ImagePositionType = "left" | "right";
-type ButtonsType = {
+type ButtonType = {
   label: string;
   link: string;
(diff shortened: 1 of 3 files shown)


client-app/shared/static-content/components/call-to-action-with-image.vue
```diff
@@ -30,17 +30,24 @@
...
-  model: {
-    type: Object,
-    required: true,
-  },
+type ImagePositionType = "left" | "right";
+type ButtonType = {
+  label: string;
+  link: string;
+};
+
```
asvishnyakov, line 38: Don't use separate types please if they are not used anywhere else, inline if enough

asvishnyakov (approved):

PR #763, opened 2023-09-25
fix: make errors in payment card component upper cased

NaMax66:
Description:
https://virtocommerce.atlassian.net/browse/ST-4969

--

QA-test:

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.40.0-pr-763-e573-e573778e.zip


asvishnyakov (changes requested):
You just need to provide correct label in `useField` options

asvishnyakov (approved):

PR #764, opened 2023-09-26
feat: reduce main page image size

NaMax66:
Description:
https://virtocommerce.atlassian.net/browse/ST-4973

--

QA-test:

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.40.0-pr-764-5e7f-5e7f3141.zip


asvishnyakov:
Why do you changed size of hexa images

NaMax66:
> Why do you changed size of hexa images?

LightHouse suggestion. Actual size of hexas is larger than needed for elements containers on the page.
Look at the screenshot in the task

asvishnyakov:
I mean `width` and `height`

NaMax66:
> I mean `width` and `height`

Because it is reduce size and it is enough for the section context

PR #766, opened 2023-09-27
fix: absolute link in CTA

NaMax66:
Description:
https://virtocommerce.atlassian.net/browse/ST-5110

--

QA-test:

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.41.0-pr-766-392f-392f02e5.zip

### DIFF
--- a/client-app/shared/static-content/components/call-to-action.vue
+++ b/client-app/shared/static-content/components/call-to-action.vue
@@ -1,10 +1,15 @@
 <template>
-  <div class="bg-gray-800 py-10 lg:py-24">
+  <div class="py-10 lg:py-24" :class="background">
     <div class="mx-auto w-full max-w-screen-xl px-5 md:px-12">
       <div class="flex flex-col items-center justify-between space-y-10 lg:block">
-        <div class="text-center text-2xl font-bold text-white lg:text-3xl">{{ model.title }}</div>
+        <div class="text-center text-2xl font-bold text-white lg:text-3xl">{{ title }}</div>
         <div class="flex w-full items-center justify-center space-x-6">
-          <VcButton v-for="(item, index) in model.buttons" :key="index" :to="item.link" class="flex-1 lg:flex-none">
+          <VcButton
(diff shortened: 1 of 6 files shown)


client-app/shared/static-content/components/call-to-action.vue
```diff
@@ -14,15 +19,32 @@
...
+const getLinkAttr = (link?: string): LinkAttrType => {
+  if (link) {
+    if (link.startsWith("/")) {
+      return { to: link };
+    } else {
+      return { externalLink: link };
+    }
+  }
+  return {};
+};
```
asvishnyakov, line 49: We already have exactly the same code in `call-to-action-with-image`, move shared code to composable please

asvishnyakov (changes requested):

content/pages/test_demo_landing.en-US.page
```diff
@@ -66,7 +66,7 @@
       "type": "call-to-action",
       "id": "calltoactionsNDR",
       "title": "Book a free ecommerce consulting session",
-      "background": null,
+      "background": "bg-gray-800",
```
goldenmaya, line 69: bad idea to use tailwind classes in json, it should be HEX code, like "\#DDDDDD"

goldenmaya (changes requested):

content/pages/test_demo_landing.en-US.page
NaMax66: We could select only one of the predefined  in _sections.json

The context is the Page Builder
As a user we can see variants in the settings tab in the preview mode
image

asvishnyakov (approved):

content/pages/test_demo_landing.en-US.page
goldenmaya: ok, use new color palette variables then.

goldenmaya (approved):

PR #772, opened 2023-10-02
feat: related products

goldenmaya:
Description:
https://virtocommerce.atlassian.net/browse/ST-4931

--

QA-test:

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.41.0-pr-772-f8c9-f8c95600.zip


client-app/pages/product.vue
```diff
@@ -173,14 +128,12 @@ watchEffect(() => {
...
-}
-</style>
+watch(
+  _relatedProducts,
+  (newVal) => {
+    relatedProducts.value = [...newVal];
+  },
+  { deep: true },
+);
+</script>
```
asvishnyakov, line 139: Why we need this? It's exactly the same as just use `relatedProducts` directly

asvishnyakov: That's because we use `shallowRef`, so changes will trigger update only if array was changed, not its elements. Need to find why do we need this, what changes product properties

NaMax66: Let's fix bug first

client-app/ui-kit/components/organisms/product-image/vc-product-image.stories.ts
```diff
@@ -0,0 +1,44 @@
...
+};
+
+export const NoImage = Template.bind({});
+NoImage.args = {
+  alt: "Product",
+};
+
+export const Carousel = Template.bind({});
+Carousel.args = {
+  imgSrc: "https://vcst-dev.paas.govirto.com/cms-content/assets/catalog/VIM70C3/1426013445000_1121979.jpg",
```
NaMax66, line 34: image should be from this repo. This links can became invalid

NaMax66, line 34: All images should be from this repo. Links from the platform can become invalid.
UI Kit should not know anything about any other storages

NaMax66 (changes requested):

client-app/pages/product.vue
goldenmaya: Not `shallowRef `. Fixed.

NaMax66 (approved):

asvishnyakov (changes requested):
Use `webp` instead of `jpeg` or `png` please. Me or @NaMax66 can convert them, if you need help

asvishnyakov (approved):

PR #783, opened 2023-10-05
fix: remove quantity header in wishlist

NaMax66:
Description:
https://virtocommerce.atlassian.net/browse/ST-4929

--

QA-test:

Demo-test:

Download artifact URL:

Artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.41.0-pr-783-04be-04bef9f7.zip

### DIFF
--- a/client-app/shared/wishlists/components/wishlist-line-items.vue
+++ b/client-app/shared/wishlists/components/wishlist-line-items.vue
@@ -1,26 +1,5 @@
 <template>
   <VcLineItems :items="items" removable disable-subtotal @remove:items="$emit('remove:items', $event)">
-    <template #header-product>
-      <div class="text-center">
-        {{ $t("common.labels.product") }}
-      </div>
-    </template>
-    <template #header-properties>
-      <div class="text-center">
-        {{ $t("common.labels.properties") }}
-      </div>
-    </template>
-    <template #header-price>
-      <div class="text-center">
-        {{ $t("common.labels.price_per_item") }}
-      </div>
-    </template>
-    <template #titles>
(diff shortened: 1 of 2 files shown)


client-app/shared/wishlists/components/wishlist-line-items.vue
```diff
@@ -1,7 +1,22 @@
...
+    <template #header-properties>
+      <div class="text-center">
+        {{ $t("common.labels.properties") }}
+      </div>
+    </template>
+    <template #header-price>
+      <div class="text-center">
+        {{ $t("common.labels.price_per_item") }}
+      </div>
+    </template>
```
asvishnyakov, line 17: I appreciate your wish to refactor this, but I will propose to do it in other task.
You make this header columns slots, but there is no same slots for columns in line item itself. Also, I think it's better to use another approach and put everything in default slot, because now you can't remove or add additional columns between these.

asvishnyakov: Resolved as "leave as is" after discussion

goldenmaya, line 17: we have design, roll it back, please

goldenmaya (changes requested):

client-app/shared/wishlists/components/wishlist-line-items.vue
goldenmaya: align quantity only

goldenmaya (approved):

PR #822, opened 2023-10-19
feat: item quantity fields validation for cart and wishlists pages

Andrew-Orlov:
## Description

Add item quantity fields validation for cart and wishlists pages

### Jira-link:

https://virtocommerce.atlassian.net/browse/ST-5061
https://virtocommerce.atlassian.net/browse/ST-5065

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.42.0-pr-822-09f3-09f383f3.zip


client-app/core/composables/useFieldValidationSchema.ts
```diff
@@ -0,0 +1,30 @@
+import { useI18n } from "vue-i18n";
+import type { NumberSchema } from "yup";
+
+function mutateQuantityFieldSchema(schema: NumberSchema, minQuantity?: number, maxQuantity?: number): NumberSchema {
```
asvishnyakov, line 4: 1. File name and composable name should match
2. Move it to cart comporables
3. Currently it's utility and not composable, because [composable should be stateful](https://vuejs.org/guide/reusability/composables.html#what-is-a-composable), but we probably want to move quantity validation schema here fully, because quantity type (number), label, error message and so on are equal in all places.

asvishnyakov (changes requested):

asvishnyakov (changes requested):

asvishnyakov (approved):

PR #832, opened 2023-10-24
refactor: validation errors in cart

NaMax66:
## Description
Change deprecated useCartValidationErrorTranslator -> useErrorsTranslator in cart-line-items.vue and add-to-cart.vue

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.43.0-pr-832-496b-496b5764.zip


client-app/shared/cart/components/add-to-cart.vue
```diff
@@ -90,6 +90,7 @@ const maxQuantity = computed(() => props.product.maxQuantity);
 const { cart, addToCart, changeItemQuantity } = useCart();
 const { t } = useI18n();
 const ga = useGoogleAnalytics();
+const { getTranslation } = useErrorsTranslator("validation_error.");
```
Andrew-Orlov, line 93: Use only locale key prefix `validation_error` without dot. Dot symbol should be a part of composable logic.

NaMax66: What if we have several dots? "shared.account.sign_in_form.errors."

asvishnyakov (approved):

PR #857, opened 2023-11-07
feat: relocate "change password" to account menu

NaMax66:
## Description
Separate change password form from profile page.

### Jira-link:
https://virtocommerce.atlassian.net/browse/ST-4665


client-app/public/static/images/dashboard/icons/pass.svg
asvishnyakov, line 1: Name it `key` or `password` please. I know we already have some abbreviations, but it's better to avoid them at least in new code.

asvishnyakov (approved):

PR #868, opened 2023-11-22
feat: cart loading without blocking user input & proper concurrent requests

asvishnyakov:
# Description

### Typed document node
[Recommended way](https://the-guild.dev/graphql/codegen/docs/guides/react-vue) to use GraphQL Codegen is [client preset](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client) now. We use `maybeValue` configuration, which isn't supported by this preset. Omitting this will cause typing change for generated types from `field?: Type` to `field?: Type | null`, so we can't use it now. But this PR bring one of important features of it: `typed-document-mode` plugin.

By default GraphQL uses `DocumentNode` object for generated documents, which doesn't provide information about which queries/mutations was requested and with which variables. [TypedDocumentNode](https://github.com/dotansimha/graphql-typed-document-node) add such information to generated code.

#### Why it's important?

That's how we used generated mutations earlier:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/4b11db8acc90b4423615ddc73eee0395ecd72e5a/client-app/core/api/graphql/cart/mutations/addOrUpdateCartPayment/addOrUpdateCartPaymentMutation.graphql#L3-L7
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/4b11db8acc90b4423615ddc73eee0395ecd72e5a/client-app/core/api/graphql/cart/mutations/addOrUpdateCartPayment/index.ts#L3
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/4b11db8acc90b4423615ddc73eee0395ecd72e5a/client-app/core/api/graphql/cart/mutations/addOrUpdateCartPayment/index.ts#L11-L19
* `Required<Pick<Mutations, "addOrUpdateCartPayment">>` is a workaround to get as non-optional `cart` field of `addOrUpdateCartPayment` (not our `AddOrUpdateCartPayment`!) mutation of XAPI and
* `MutationsAddOrUpdateCartPaymentArgs` is arguments of `addOrUpdateCartPayment` (not our `AddOrUpdateCartPayment`!) mutation of XAPI

GraphQL Codegen already generate all required code: it's `AddOrUpdateCartPaymentMutation` and `AddOrUpdateCartPaymentVariables`, so this code can be rewritten as
```typescript
import mutationDocument from "./addOrUpdateCartPaymentMutation.graphql";

await graphqlClient.mutate<
    AddOrUpdateCartPaymentMutation,
    AddOrUpdateCartPaymentVariables
  >({
    mutation: mutationDocument,
    variables: ...
    })
```
but this can be simplified even more: if you will use typed document node as `mutationDocument`, typescript will automatically infer all types:
```typescript
import { AddOrUpdateCartPaymentDocument } from "@/core/api/graphql/types";

await graphqlClient.mutate({
  mutation: AddOrUpdateCartPaymentDocument,
  variables: ...
})
```

### Operation names
`named-operations-object` plugin allows to generate special object (`OperationNames`) with names of all our (not XAPI!) queries, mutations & fragments. It's useful for strong typing in Apollo, for example:
instead of
```typescript
refetchQueries: ["GetFullCart"]
```
we can write
```typescript
refetchQueries: [OperationNames.Query.GetFullCart]
```
### @vue/apollo-composable
`mutate` and other function of Apollo GraphQL client is low-level API. [Recommended way](https://www.apollographql.com/docs/react/data/mutations#executing-a-mutation) to use it is to call `useQuery, useMutation, etc.`, which is React hooks originally (as Apollo provide native client for React), but has almost identical alternatives for Vue.js: integration package is `@vue/apollo-composable`: while package is still in beta status, it has long life (more than a few years), regularly receive updates and [supported officially](https://www.apollographql.com/docs/react/integrations/integrations/#vue).

#### Why it's important?
It's properly handle loading status, watching (i.e. which query or mutation used on current page) and other built-in features. For example:
without (Apollo doesn't see any active query):
image
with:
image

Query watching is important for query refetching to sustain correct application state in case of concurrent requests.

### Dependency injection
Vue.js [support dependency injection](https://vuejs.org/api/composition-api-dependency-injection.html) and `@vue/apollo-composable` uses it.
The following code used to provide Apollo client for `@vue/apollo-composable`:

https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/d4e03ae92f4fa54bcfb2707497135c7bc453fe7f/client-app/app-runner.ts#L76-L82

After that, you can simply use all composables:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/d4e03ae92f4fa54bcfb2707497135c7bc453fe7f/client-app/core/api/graphql/cart/mutations/addOrUpdateCartPayment/index.ts#L8
without direct client call.
In case we will need to use them outside of `setup` (DI works only here), we can [inject it manually](https://v4.apollo.vuejs.org/guide-composable/setup.html):
```typescript
provideApolloClient(apolloClient)(() => useQuery(...))
```
If you need to use Apollo client for some additional actions, you can use `resolveClient` function from `useApolloClient` composable:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/d4e03ae92f4fa54bcfb2707497135c7bc453fe7f/client-app/core/api/graphql/cart/mutations/clearCart/index.ts#L9-L11

### State management
[As described](https://www.apollographql.com/docs/react/local-state/local-state-management) in Apollo documentation

> At its core, Apollo Client is a state management library that happens to use GraphQL to interact with a remote server. 

While we don't use some of it's features, like local state, because Vue.js has built-in alternative (composables), Apollo's state management is very useful for concurrent requests handling, updates in other tabs and etc.

#### Cache
Apollo Client has built-in [in-memory cache](https://www.apollographql.com/docs/react/caching/overview), which is also represent GraphQL objects state.

Instead of manual storing `cart` and assigning the result of each mutation to it:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/4b11db8acc90b4423615ddc73eee0395ecd72e5a/client-app/shared/cart/composables/useCart.ts#L57
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/4b11db8acc90b4423615ddc73eee0395ecd72e5a/client-app/shared/cart/composables/useCart.ts#L183-L184
simply expose the result of `query` reference from `useQuery`:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/d1b4546f7e1b47c8db47cc4f69d2c456e801ead4/client-app/shared/cart/composables/useCart.ts#L52-L53
Apollo will update it property based on combination of type name & ID after each query request or mutation.
```typescript
  watch(cart, (result) => {
    console.log(result);
  });
```
cache

#### Loading
Apollo client & it's integration with Vue.js has built-in [loading tracking](https://www.apollographql.com/docs/react/data/queries#inspecting-loading-states).

Instead of manual tracking with `loading` reference
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/4b11db8acc90b4423615ddc73eee0395ecd72e5a/client-app/shared/cart/composables/useCart.ts#L56
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/4b11db8acc90b4423615ddc73eee0395ecd72e5a/client-app/shared/cart/composables/useCart.ts#L179-L180
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/4b11db8acc90b4423615ddc73eee0395ecd72e5a/client-app/shared/cart/composables/useCart.ts#L190-L192
simple expose `loading` reference from `useQuery` or `useMutation`:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/d1b4546f7e1b47c8db47cc4f69d2c456e801ead4/client-app/shared/cart/composables/useCart.ts#L52
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/d1b4546f7e1b47c8db47cc4f69d2c456e801ead4/client-app/shared/cart/composables/useCart.ts#L117-L123
As you can see, you can combine all loading states from all queries and mutations or separate them (as on cart page):
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/d1b4546f7e1b47c8db47cc4f69d2c456e801ead4/client-app/shared/cart/composables/useCart.ts#L354-L373

#### Refetching
Let's closer look at this scenario:
1. User add Product 1 to cart
2. User add Product 2 to cart
3. User go to cart page

In some cases, these requests may be executed in the following order:
1. Product 1 added to cart (short cart with 1 item loaded)
2. Cart page opens (full cart with 1 item loaded)
3. Product 2 added to cart (short cart with 2 items loaded)

In such cases, we need to refetch full cart. To do it, add
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/d4e03ae92f4fa54bcfb2707497135c7bc453fe7f/client-app/core/api/graphql/cart/mutations/addItemToCart/index.ts#L11

Also, we may want to force query refetching manually. For example, if user opens cart page, it should always be requested again, even if nothing changed, just in case something changed in backend (other user bought product and available quantity changed, product was deleted, etc.). To do it, just call `refetch` function, if `load` return `false`:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/d1b4546f7e1b47c8db47cc4f69d2c456e801ead4/client-app/shared/cart/composables/useCart.ts#L134

Apollo separate first loading and refetching states; to track refetching and update `loading` correspondingly, the following flag should be used for queries:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/d4e03ae92f4fa54bcfb2707497135c7bc453fe7f/client-app/core/api/graphql/cart/queries/getShortCart/index.ts#L7

#### Batching
Because now we don't block UI on cart, user can do multiple changes quickly. It's better to aggregate them and send as a single request to server. There are two cases:
1. It's the same action and we have mutation which accept array of elements, such as `selectItem` for select for checkout functionality or it's the same single element such as comment (`changeComment`)
2. It's different actions or it's not the same element and we don't have mutation which accept array of such elements (like `changeCartItemQuantity`)
In first case we will need for manual aggregation any way, as there is no way to tell Apollo what we want replace multiple calls of one mutation with single call of another.
In second case, Apollo will help us.

Apollo provide middleware in graphql operation processing chain. It's calling [links](https://www.apollographql.com/docs/react/api/link/introduction). One of them is HTTP Batch link, which will aggregate (on sending) and then separate (after response) requests into one:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/babb59d816e187e3657d05101f16b2433f6dc92c/client-app/core/api/graphql/cart/links/index.ts#L61-L64
batching
To apply this link only for selected operation, we `split` chain (create conditional branch):
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/753591bb80877b01294b2fed0fea63f4453728fb/client-app/core/api/graphql/cart/links/index.ts#L58-L59

##### Cart query skipping
Because every cart mutation return cart object and mutations in graphql executed in sequence, we need cart only from the last one.
Apollo has no good point where you can detect what operation is the last one and change variables so we can skip all except the last one. But thee is workaround:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/babb59d816e187e3657d05101f16b2433f6dc92c/client-app/core/api/graphql/cart/mutations/addOrUpdateCartShipment/addOrUpdateCartShipmentMutation.graphql#L3-L6
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/babb59d816e187e3657d05101f16b2433f6dc92c/client-app/core/api/graphql/cart/links/index.ts#L74-L85
skip-query
It's also working when you try to load both short & full cart:
skip-query
But in case of mutations, they will be executed and only returning fields (query) will be skipped, while short cart query will be skipped entirely and will not be executed on server. Compare:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/babb59d816e187e3657d05101f16b2433f6dc92c/client-app/core/api/graphql/cart/queries/getShortCart/getShortCartQuery.graphql#L10-L13

#### Broadcasting
When we change shopping cart in one tab (add product to cart, change quantity, change selection or any other cart related action), we need to reload it in other tabs. Previously it was done manually: 
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/7259fbd827596060f22d44e4a3fa22a40f34f93c/client-app/shared/cart/composables/useCart.ts#L248
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/7259fbd827596060f22d44e4a3fa22a40f34f93c/client-app/shared/cart/composables/useCart.ts#L258
Now there is custom link, which will broadcast `cartReloadEvent` after getting response from any of listed mutations:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/753591bb80877b01294b2fed0fea63f4453728fb/client-app/core/api/graphql/cart/links/index.ts#L43-L52

#### Variables
Apollo [allows](https://www.apollographql.com/docs/react/data/mutations#providing-options) you to set global variables for your queries and mutations. For example:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/babb59d816e187e3657d05101f16b2433f6dc92c/client-app/core/api/graphql/cart/mutations/addOrUpdateCartShipment/index.ts#L8
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/babb59d816e187e3657d05101f16b2433f6dc92c/client-app/shared/cart/composables/useCart.ts#L260

Globally used variables is placed here (and may be extended by adding other combinations, such as only `storeId, cultureName`, if needed):
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/babb59d816e187e3657d05101f16b2433f6dc92c/client-app/core/api/graphql/composables/useVariables.ts#L11-L25

And locally (like in cart) variation is placed in corresponding folder. It also allow to add new global variables in single place and don't miss optional, but available variables, such as `cartId` in every mutation:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/babb59d816e187e3657d05101f16b2433f6dc92c/client-app/core/api/graphql/cart/composables/useCartVariables.ts#L24
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/babb59d816e187e3657d05101f16b2433f6dc92c/client-app/core/api/graphql/cart/composables/useCartVariables.ts#L31
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/babb59d816e187e3657d05101f16b2433f6dc92c/client-app/core/api/graphql/cart/composables/useCartVariables.ts#L37-L40

The problem is what Apollo will merge such variables shallowly, i.e. like a spread operator, without deep merge:
```typescript
useLazyQuery(..., { storeId })
load(); // works

useMutation(..., { command: { storeId } })
mutate({ command: { id }}); // don't work

// because it's like
const variables = {
  command: { ... }
};
const test = {
  ...variables
  command: { ... }
}  
```
The following custom composable fix it:
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/babb59d816e187e3657d05101f16b2433f6dc92c/client-app/core/api/graphql/composables/useMutation.ts#L10
https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/babb59d816e187e3657d05101f16b2433f6dc92c/client-app/core/api/graphql/composables/useMutation.ts#L26

I hope they will fix this issue in feature, because Apollo already has deep merge helpers, the problem is typing (`DeepOmitByType` helper)

### Jira-link:
https://virtocommerce.atlassian.net/browse/ST-5056
https://virtocommerce.atlassian.net/browse/ST-5123
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.51.0-pr-868-9c66-9c6672ca.zip
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.47.0-pr-868-4ead-4ead8d78.zip


client-app/core/api/graphql/cart/links/index.ts
```diff
@@ -0,0 +1,94 @@
...
+import { API_URL } from "@/core/api/graphql/consts";
+import { OperationNames } from "@/core/api/graphql/types";
+import { cartReloadEvent, useBroadcast } from "@/shared/broadcast";
+import { DEFAULT_DEBOUNCE_IN_MS } from "@/shared/cart/constants";
+import type { Operation } from "@apollo/client/core";
+
+const queryNames = [
+  OperationNames.Query.GetFullCart,
+  OperationNames.Query.GetShortCart,
+  // validateCoupon doesn't modify anything and should be a query
```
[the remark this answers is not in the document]
asvishnyakov: https://virtocommerce.atlassian.net/browse/PT-14789

client-app/core/api/graphql/cart/mutations/addCoupon/index.ts
```diff
@@ -1,23 +1,16 @@
...
+export function useAddCouponMutation(cart?: MaybeRef<CartIdFragment | undefined>) {
+  return useMutation(AddCouponDocument, useCartMutationVariables(cart));
+}
 
-  return data!.addCoupon;
+/** @deprecated Use {@link useAddCouponMutation} instead. */
+export async function addCoupon(couponCode: string): Promise<CartType> {
+  const { mutate } = useAddCouponMutation();
+  const result = await mutate({ command: { couponCode } });
+  return result!.data!.addCoupon as CartType;
```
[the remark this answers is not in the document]
asvishnyakov: It's a backward compatible code
image
There is no possibility to return CartType without union null/undefined, if we will add condition. Code is already marked deprecated, new version return nullable result and doesn't make assertions.

client-app/core/api/graphql/cart/links/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    return forward(operation).map((data) => {
+      broadcast.emit(cartReloadEvent);
+      return data;
+    });
+  }),
+);
+
+const batchOperationNames = queryNames.concat(fullCartMutationNames);
+
+// BatchHttpLink is used to batch multiple operations into one request
```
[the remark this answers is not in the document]
asvishnyakov: Done

NaMax66 (approved):

client-app/core/api/graphql/cart/composables/useCartVariables.ts
```diff
@@ -0,0 +1,45 @@
...
+): Ref<UseMutationOptions<TResult, ICartMutationVariables>> {
+  return computed(() => ({
+    variables: {
+      command: {
+        ...unref(useAllGlobalVariables()),
+        cartId: unref(cart)?.id,
+      },
+      skipQuery: false,
+    },
+    ...toValue(options),
```
goldenmaya, line 43: I don’t like the unclear behavior of toValue in this case. Instead of using toValue, we can simply check whether options is a function or a reactive object. Something like
`const _options = typeof options === 'function' ? options() : unref(options) || {};`

client-app/core/types/utility.ts
goldenmaya, line 1: Too complicated

client-app/core/api/graphql/cart/composables/useCartVariables.ts
asvishnyakov: toValue is built-in vue.js helper what do exactly the same as you wrote

client-app/core/types/utility.ts
asvishnyakov: Yep, and better should be covered by tests, I'll try to do it. But the main reason of this is to allow deep merge of variables i.e. allow to specify `storeId`, `userId` in one place and use in mutation only mutation-specific variables like `shipmentId` or `lineItemIds`. `DeepOmitByType` will omit properties of one type (mutation variables) by properties of another type (global variables) and left all other properties as is, which is exactly behavior we need to. It's important because some other projects may need additional global variables like `contractId` and would like to add it in one place.

goldenmaya (approved):

Andrew-Orlov (approved):

PR #884, opened 2023-12-06
fix: prevent checkout/place order only for selected cart items with validation errors

Andrew-Orlov:
## Description

Prevent checkout/place order only for selected items with validation errors

### Jira-link:

https://virtocommerce.atlassian.net/browse/ST-5453

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.45.0-pr-884-e91d-e91de3b4.zip

### DIFF
--- a/client-app/core/enums/validation-error-object-type.enum.ts
+++ b/client-app/core/enums/validation-error-object-type.enum.ts
@@ -1,0 +1,7 @@
+export enum ValidationErrorObjectType {
+  CartProduct = "CartProduct",
+  CatalogProduct = "CatalogProduct",
+  LineItem = "LineItem",
+  Payment = "Payment",
+  Shipment = "Shipment",
+}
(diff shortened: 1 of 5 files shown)


client-app/core/enums/validation-error-object-type.enum.ts
```diff
@@ -0,0 +1,7 @@
+export enum ValidationErrorObjectType {
+  CartProduct = "CartProduct",
+  CatalogProduct = "CatalogProduct",
```
asvishnyakov, line 3: https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/7259fbd827596060f22d44e4a3fa22a40f34f93c/client-app/core/constants/products.ts#L11 replace and remove please, it's used for the same reason. Thank you for the enum

asvishnyakov (changes requested):

client-app/core/enums/validation-error-object-type.enum.ts
NaMax66: I would name it ValidationErrorType

Andrew-Orlov: > I would name it ValidationErrorType

I keep original class name for naming in this case.

asvishnyakov (approved):

goldenmaya (approved):

NaMax66 (approved):

PR #885, opened 2023-12-06
fix: empty list styles

goldenmaya:
### Jira-link:

https://virtocommerce.atlassian.net/browse/ST-5404

### DIFF
--- a/client-app/pages/account/list-details.vue
+++ b/client-app/pages/account/list-details.vue
@@ -82,7 +82,7 @@
         </template>
 
         <!-- Empty list -->
-        <VcEmptyView v-else :text="$t('shared.wishlists.list_details.empty_list')">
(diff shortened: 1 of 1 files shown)


asvishnyakov (approved):

client-app/pages/account/list-details.vue
```diff
@@ -82,7 +82,7 @@
         </template>
 
         <!-- Empty list -->
-        <VcEmptyView v-else :text="$t('shared.wishlists.list_details.empty_list')">
+        <VcEmptyView v-else :text="$t('shared.wishlists.list_details.empty_list')" class="lg:mt-32">
```
NaMax66, line 85: Why use such a large indent?  If you want to align an element in the middle of the screen, then it's better to do it more explicitly.

goldenmaya: there's no middle of the screen

PR #886, opened 2023-12-06
fix: remove trim from user input

NaMax66:
## Description
https://virtocommerce.atlassian.net/browse/ST-5463

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.45.0-pr-886-15f1-15f1a2e6.zip

### DIFF
--- a/client-app/shared/layout/components/search-bar/search-bar.vue
+++ b/client-app/shared/layout/components/search-bar/search-bar.vue
@@ -180,8 +180,11 @@ const searchPhraseInUrl = useRouteQueryParam<string>(QueryParamName.SearchPhrase
 const categoriesRoutes = useCategoriesRoutes(categories);
 
 const searchPhrase = ref("");
+const trimmedSearchPhrase = computed(() => {
+  return searchPhrase.value.trim();
+});
 
-const isApplied = computed<boolean>(() => searchPhraseInUrl.value === searchPhrase.value.trim());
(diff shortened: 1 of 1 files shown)


client-app/shared/layout/components/search-bar/search-bar.vue
```diff
@@ -213,7 +213,7 @@ async function searchAndShowDropdownResults(): Promise<void> {
   hideSearchDropdown();
 
   if (
-    searchPhrase.value === "" ||
+    searchPhrase.value.trim() === "" ||
```
goldenmaya, line 216: May be we should optimize the code by storing the trimmed string in a variable trimmedSearchPhrase or smth, to avoid multiple calls of trim().

asvishnyakov: I agree it's better to have single computed with trim instead

asvishnyakov (changes requested):

client-app/shared/layout/components/search-bar/search-bar.vue
asvishnyakov: I agree it's better to have single computed with single `trim` instead

asvishnyakov (approved):

goldenmaya (approved):

PR #939, opened 2024-01-26
feat(vc-select): implement multiselect

goldenmaya:
### Jira-link:

https://virtocommerce.atlassian.net/browse/VCST-178
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.51.0-pr-939-8a07-8a07b901.zip


client-app/pages/account/checkout-defaults.vue
```diff
@@ -34,26 +34,28 @@
           size="auto"
         >
           <template #placeholder>
-            <VcSelectItem>
-              <VcSelectItemImage src="/static/icons/placeholders/select-payment.svg" class="bg-gray-100/80" />
-              <VcSelectItemText>
-                {{ $t("common.placeholders.select_payment_method") }}
-              </VcSelectItemText>
-            </VcSelectItem>
+            <div class="flex items-center gap-3 p-3 text-sm">
```
asvishnyakov, line 37: Why do we moved from component to duplicated `div`'s with same classes? Looks like regression for me.
I think it's better to have component here, may be customizable. If we don't want to have specific for `VcSelect`, then reuse `MenuItem` or something like that.

client-app/ui-kit/components/molecules/select/vc-select.vue
```diff
@@ -1,143 +1,125 @@
 <template>
```
asvishnyakov, line 1: This component is too much complex, I think we need to find a way to separate it intomultiple components (single/multiple, container/filter/single/multi, etc.)

asvishnyakov (changes requested):

client-app/ui-kit/components/molecules/input/vc-input.vue
```diff
@@ -144,6 +157,18 @@ function togglePasswordVisibility() {
   inputType.value = isPasswordVisible.value ? "text" : "password";
 }
 
+function handleContainerClick() {
+  if (inputElement.value) {
+    inputElement.value.focus();
+    inputElement.value.click();
+  }
+}
```
asvishnyakov, line 165: Why do we need this?

client-app/pages/account/checkout-defaults.vue
goldenmaya: We have new design for selects https://virtocommerce.atlassian.net/browse/VCST-468

client-app/ui-kit/components/molecules/select/vc-select.vue
goldenmaya: Yes, we can, we need to talk about it and create a new ticket.

client-app/ui-kit/components/molecules/input/vc-input.vue
asvishnyakov: Needed to bypass events from inner elements like show password eye

client-app/ui-kit/components/molecules/select/vc-select.vue
asvishnyakov: https://virtocommerce.atlassian.net/browse/VCST-471

asvishnyakov (approved):

NaMax66 (approved):

asvishnyakov (approved):

PR #970, opened 2024-02-20
refactor: useFetch from vueuse, interceptors and shared error handling

asvishnyakov:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-341
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.52.0-pr-970-9f2e-9f2e6a68.zip

### DIFF
--- a/client-app/broadcast.ts
+++ b/client-app/broadcast.ts
@@ -82,7 +82,7 @@ export function setupBroadcastGlobalListeners() {
     });
   });
   on(openReturnUrl, () => {
-    location.href = getReturnUrlValue() ?? location.href;
+    location.href = getReturnUrlValue() ?? "/";
   });
 
   on(forbiddenEvent, () => {
--- a/client-app/broadcast.ts
+++ b/client-app/broadcast.ts
@@ -82,7 +82,7 @@ export function setupBroadcastGlobalListeners() {
     });
   });
   on(openReturnUrl, () => {
-    location.href = getReturnUrlValue() || location.href;
+    location.href = getReturnUrlValue() ?? location.href;
   });
 
   on(forbiddenEvent, () => {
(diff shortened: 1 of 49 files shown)


client-app/broadcast.ts
```diff
@@ -80,7 +82,7 @@ export function setupBroadcastGlobalListeners() {
     });
   });
   on(openReturnUrl, () => {
-    location.href = getReturnUrlValue() || "/";
+    location.href = getReturnUrlValue() ?? location.href;
```
NaMax66, line 85: why?

client-app/core/api/graphql/utils.ts
```diff
@@ -1,5 +1,30 @@
...
+import { GraphQLErrorCode } from "@/core/api/graphql/enums";
 import type { ApolloClient } from "@apollo/client/core";
+import type { GraphQLErrors, NetworkError } from "@apollo/client/errors";
+import type { GraphQLError } from "graphql";
+
+export function hasErrorCode(graphQLErrors: ReadonlyArray<GraphQLError> | undefined, errorCode: GraphQLErrorCode) {
+  return graphQLErrors?.some((graphQLError) => graphQLError.extensions.code === errorCode);
+}
+
+export function toServerError(
```
NaMax66, line 12: it is a function with the same name in client-app/core/api/common/utils/to-server-error.ts
you sure it should be two?

asvishnyakov: Yes. They has different parameters. We probably may do overload, but they are weird style in typescript if they have different types for arguments and I'll prefer to keep everything for graphql in graphql folder.

NaMax66 (approved):

Andrew-Orlov (approved):

PR #976, opened 2024-02-23
fix: shipments or payments may accidentally duplicate on cart

asvishnyakov:
### What is [optimistic response](https://www.apollographql.com/docs/react/performance/optimistic-ui/)?
Optimistic response is assumption of how server will respond to the request. It allows to immediately update UI and will be replaced with backend response as soon as client gets it.

### Why not manually?
1. Need to write a lot of supporting code (see removed code)
2. Need to manually track negative cases (like errors or unexpected response) (even wasn't implemented and will add another ton fo code)
3. Need to manually properly track side effects (such as physical products add or remove if digital exists) (wasn't implemented too and caused part of bug)

### Why generate id and use typename?
For temporary (between call start and response from server) cache only (see docs).

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-526
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.52.0-pr-976-7e77-7e779caa.zip


client-app/core/api/graphql/utils.ts
```diff
@@ -12,6 +14,19 @@ export function filterActiveQueryNames<TCacheShape = any>(
...
 }
 
+export function toOptimisticResponse<T extends { id?: string } | undefined>(
+  data: T,
+  __typename: string,
+): T | undefined {
+  if (isDefined(data)) {
+    return {
+      ...data,
+      id: data.id ?? uuidv4(),
```
NaMax66, line 24: do you really need generate id here?

asvishnyakov: There are 3 stages:
1. Already loaded
2. Run mutation, but don't get server response
3. Get server response
optimisticResponse is overwriting state only at stage 2 and revert changes (on error) or apply response (if success) at stage 3. This ID generation is only for stage 2.

asvishnyakov: Renamed, removed unnecessary

NaMax66 (approved):

PR #1033, opened 2024-03-29
chore: generate graphql types in separate files

asvishnyakov:
### What's changed
`yarn generate:graphql:types` now generates separate files (one per each `*.graphql`) instead of generating single `core/api/graphql/types.ts`.

`core/api/graphql/types/index.ts` is now generated by `yarn generate:graphql:bundle` and simply include all `*.generated.ts` files from `client-app/core/api/graphql` for backward compatibility.

### Reasons
1. **In next steps we may reduce main bundle size** by direct import from separate generated files (it will work only we will replace all of them), so only generated documents necessary to page will be included into page bundle and will be excluded from main bundle, i.e. `getFullCart` only on cart page, `getMe` in still main bundle.
2. It's a **first step for modularity**: now you can place `graphql` files anywhere in project (i.e. in separate module) and they will _not_ be included into `core/api/graphql/types/index.ts`. In the next steps we may separate schema and module-related types, so we will no longer require module to be present, if related theme module isn't used.

### Required actions after merge
Just run `yarn generate:graphql:types` once.  No other changes.

To simplify review of this PR, the result of `yarn generate:graphql:types` is in separate PR #1034 which needs to be merged into this one after review.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-626


asvishnyakov:
Transformed into #1253

PR #1037, opened 2024-04-01
fix: reset styles before global

asvishnyakov:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-431
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.57.0-pr-1037-5206-5206e044.zip


goldenmaya (changes requested):
will be updated

goldenmaya (approved):

PR #1068, opened 2024-05-01
fix: hide properties without values

goldenmaya:
### Jira-link:

https://virtocommerce.atlassian.net/browse/VCST-1146
https://virtocommerce.atlassian.net/browse/VCST-1162
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.58.0-pr-1068-1316-13165be7.zip


client-app/shared/catalog/components/product/variations.vue
```diff
@@ -52,6 +52,10 @@ interface IProps {
 }
 
 defineProps<IProps>();
+
+function filteredProperties(properties: Property[]) {
+  return properties.filter((prop) => !!prop.value).slice(0, 3);
```
ivan-kalachikov, line 57: Good job! The only think we might want to consider - can our prop value be explicitly set to `0`?
If so, maybe it's better to use explicit check on `undefuned`, `null` and empty string, otherwise we never see `0` value in product variations.

goldenmaya: fixed with another bug

NaMax66 (approved):

PR #1115, opened 2024-05-20
fix: comment caching

NaMax66:
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.59.0-pr-1115-f557-f557c548.zip

### DIFF
--- a/client-app/shared/checkout/composables/useCheckout.ts
+++ b/client-app/shared/checkout/composables/useCheckout.ts
@@ -37,6 +37,9 @@ const useGlobalCheckout = createGlobalState(() => {
   const purchaseOrderNumberChanging = ref(false);
   const _purchaseOrderNumber = ref<string>();
 
+  function clearState() {
+    _comment.value = undefined;
+  }
   return {
     loading,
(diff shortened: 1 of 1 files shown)


client-app/shared/checkout/composables/useCheckout.ts
```diff
@@ -459,6 +466,10 @@ export function _useCheckout() {
     return placedOrder.value;
   }
 
+  function clearState() {
```
ivan-kalachikov, line 469: The point of creating this function is not clear - can we just call `clearGlobalCheckoutState` on the line **451**?

NaMax66: The useGlobalCheckout is a separate entity and has there own state.
The useCheckout also is a separate entity with their own state.
The idea is to have the same signature in every entity.

ivan-kalachikov: Ok, I got it, thanks for explaining :)

ivan-kalachikov (approved):

PR #1136, opened 2024-06-04
feat: switch to storefront less

NaMax66:
## Description
new storefront less architecture

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.0.0-pr-1136-1be8-1be89467.zip


package.json
```diff
@@ -6,13 +6,13 @@
...
   "license": "SEE LICENSE IN LICENSE.txt",
-  "version": "1.63.0",
+  "version": "2.0.0",
   "private": true,
   "type": "module",
   "scripts": {
     "dev": "vite",
     "dev:expose": "yarn dev --host",
-    "build": "yarn validate && yarn generate:routes && yarn build-only",
+    "build": "yarn validate && yarn build-only",
```
ivan-kalachikov, line 15: do we still need script `generate:routes` below?

.github/workflows/theme-ci.yml
```diff
@@ -116,9 +115,13 @@ jobs:
         if: success()
         run: echo "BUILD_STATE=successful" >> $GITHUB_ENV
 
-      - name: Packaging
+      - name: Add files to zip
         run: |
-          yarn compress
+          mkdir -p ./tmp/default
```
ivan-kalachikov, line 120: since we don't more use `yarn compress`, is that used somewhere else, or script can be deleted?

package.json
NaMax66: no. I keep it to not forget to remove the whole service latter

.github/workflows/theme-ci.yml
NaMax66: will be removed latter

PR #1141, opened 2024-06-10
chore: refactor routing

NaMax66:
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.61.0-pr-1141-d61d-d61dc0db.zip


Andrew-Orlov (approved):

client-app/pages/external/matcher.vue
```diff
@@ -0,0 +1,84 @@
...
+  builderIo: {
+    id: "builderIo",
+    priority: 1,
+    state: "initial",
+    isActive: isBuilderIOEnabled.value,
+  },
+});
+
+const visibleComponent = computed(() =>
+  getVisiblePreviewer(Object.keys(previewers.value).map((key) => previewers.value[key])),
```
ivan-kalachikov, line 76: `getVisiblePreviewer(Object.values(previewers.value),` does the same thing a bit more clearer

client-app/pages/external/slug-content.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <div v-if="!loading && (hasContent || objectType)" class="slug-content">
+    <component :is="CategoryComponent" v-if="objectType === 'Category'" :category-id="slugInfo?.entityInfo?.objectId" />
```
ivan-kalachikov, line 3: What is an advantage of using dynamic components here? `CategoryComponent`, `Product` and `StaticPage` are always the same components.

client-app/pages/external/slug-content.vue
```diff
@@ -0,0 +1,86 @@
...
+    case ObjectType.Category:
+      matchingRouteName = "Category";
+      break;
+  }
+
+  setMatchingRouteName(matchingRouteName);
+});
+</script>
+
+<style scoped lang="scss"></style>
```
ivan-kalachikov, line 86: can be removed as it's empty

ivan-kalachikov (changes requested):
Nice job! Just a few small things to do it even better

client-app/pages/external/slug-content.vue
NaMax66: it is for chunks https://vuejs.org/guide/components/async

ivan-kalachikov: yes, that's fine about async components, perfect solution. I meant using `<component :is="...`

it aims to dynamically render a component depends on the variable passed in `is`. But here we can use `<CategoryComponent ...` right away, since after loading it will never be changed.

ivan-kalachikov (approved):

PR #1162, opened 2024-06-28
feat: implement customer multi-organization

Andrew-Orlov:
## Description

Implement multiple organizations for customer

### Jira-link:

https://virtocommerce.atlassian.net/browse/VCST-1355

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.62.0-pr-1162-3cd4-3cd47086.zip


client-app/core/composables/useNavigations.ts
```diff
@@ -93,6 +95,27 @@ export function useNavigations() {
...
+
+    if (isMultiOrganization.value) {
+      mobileContactOrganizationsMenu.value = {
+        id: "contact-organizations",
+        title: "common.labels.my_organizations",
+        icon: "/static/images/dashboard/icons/company.svg#main",
+        children: organizationsMenuItems,
+      };
+
+      mobileAccountMenuItem.value?.children?.splice(1, 0, mobileContactOrganizationsMenu.value);
```
ivan-kalachikov, line 115: I'm not sure if it's okay to mutate a computed property. Maybe there is another way to achieve what you want here?

client-app/shared/account/composables/useUser.ts
```diff
@@ -322,9 +334,51 @@ export function useUser() {
...
+    } finally {
+      loading.value = false;
+    }
+  }
+
   return {
     isAuthenticated,
     isCorporateMember,
+    isMultiOrganization: computed(
+      () => user.value?.contact?.organizations?.items && user.value?.contact?.organizations?.items?.length > 1,
```
ivan-kalachikov, line 380: could be just `user.value?.contact?.organizations?.items?.length > 1` , since you use optional chaining

client-app/shared/account/composables/useUser.ts
```diff
@@ -322,9 +334,51 @@ export function useUser() {
...
 
+  async function switchOrganization(organizationId: string): Promise<void> {
+    loading.value = true;
+
+    try {
+      await refresh(organizationId);
+
+      broadcast.emit(reloadAndOpenMainPage, null, TabsType.ALL);
+
+      /*const { error, data } = await useFetch("/connect/token")
```
ivan-kalachikov, line 345: Not clear enough - will we need this part of code or it can be deleted completely?

ivan-kalachikov (commented):
Great job, there are a few minor things to make it even better.

client-app/shared/account/composables/useUser.ts
Andrew-Orlov: No, `items?.length` is possibly undefined.

client-app/core/composables/useNavigations.ts
ivan-kalachikov: why don't we do it something like that to avoid mutation of computed value?:

```js
const mobileContactOrganizationsMenu = ref([])
...
const mobileAccountMenuItem = computed<ExtendedMenuLinkType | null>(() =>
  menuSchema.value 
    ? getTranslatedMenuLink(menuSchema.value.header.mobile.account).splice(1, 0,  ...mobileContactOrganizationsMenu,value)
    : null,
);

function getMobileContactOrganizationsMenu() {
...
mobileAccountMenuItem.value = mobileContactOrganizationsMenu.value;
...
}
```

client-app/core/composables/useNavigations.ts
```diff
@@ -158,6 +191,11 @@ export function useNavigations() {
     matchingRouteName.value = value;
   }
 
+  watch(
+    () => menuSchema,
+    (newValue) => (newValue.value ? getTranslatedMenuLink(newValue.value.header.mobile.account) : null),
```
ivan-kalachikov, line 196: There is no any side effect in the watcher. Looks like the code should work without that.

ivan-kalachikov (approved):

PR #1216, opened 2024-07-31
chore: replace deprecated client

NaMax66:
## Description
Remove deprecated mark
Remove batch link hack

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.2.0-pr-1216-4ccd-4ccd4ea0.zip


client-app/core/api/graphql/cart/links/index.ts
```diff
@@ -1,5 +1,4 @@
 import { from } from "@apollo/client/core";
-import { batchLink } from "@/core/api/graphql/cart/links/batch";
 import { broadcastLink } from "@/core/api/graphql/cart/links/broadcast";
 
-export const cartLink = from([broadcastLink, batchLink]);
+export const cartLink = from([broadcastLink]);
```
ivan-kalachikov, line 4: batchLink supposed to decrease the amount of requests - are we sure that we want to remove that?

NaMax66: Yes, because implementation marks as Dirty hack.

PR #1221, opened 2024-08-01
fix: theme settings from Page Builder

NaMax66:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-1581
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.2.0-pr-1221-c62a-c62aad72.zip

### DIFF
--- a/client-app/public/config/settings_schema.json
+++ b/client-app/public/config/settings_schema.json
@@ -1,0 +1,69 @@
+[
+  {
+    "name": "Logo",
+    "icon": "general",
+    "settings": [
+      {
+        "type": "header",
+        "label": "Identica"
+      },
+      {
+        "type": "string",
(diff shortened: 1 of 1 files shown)


client-app/public/config/settings_schema.json
```diff
@@ -0,0 +1,69 @@
...
+      }
+    ]
+  },
+  {
+    "name": "Colors",
+    "icon": "colors",
+    "settings": [
+      {
+        "type": "color",
+        "id": "color_primary",
```
goldenmaya, line 33: Weird color variables. We don't have such colors.

NaMax66:
Transform to store level white labeling settings

PR #1253, opened 2024-08-16
chore: separate gql types for PushMessages service

NaMax66:
## Description
Separate types for PushMessages module

Add console details
image

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-626
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.4.0-pr-1253-7606-76067526.zip


scripts/graphql-codegen/generator.ts
```diff
@@ -0,0 +1,141 @@
...
+            plugins: GENERAL_PLUGINS,
+            config: GENERAL_CONFIG,
+          },
+        },
+      },
+      true,
+    );
+    // eslint-disable-next-line no-console
+    console.log(`Types for "${module.name}" module have been generated in "${moduleTypesPath}"`);
+  }
```
asvishnyakov, line 118: This will cause duplicate types. See https://github.com/VirtoCommerce/vc-theme-b2b-vue/blob/feat/VCST-626-graphql-multiple-files/scripts/generate-graphql-bundle.ts approach:
1. Generate separate files for all
2. Create files which imports them for backward-compatibility or to make separate module bundle

NaMax66: duplicates if ok

scripts/graphql-codegen/generator.ts
```diff
@@ -0,0 +1,141 @@
...
+  const JSONString = await readJsonAndReturnString(SCHEMA_PATH);
+
+  const installedIndependentModules = independentModules.filter((el) => JSONString?.includes(el.searchKey));
+
+  if (installedIndependentModules.length) {
+    // eslint-disable-next-line no-console
+    console.log("\nGenerate types for independent modules:");
+  }
+
+  for (let i = 0; i < installedIndependentModules.length; i++) {
```
ivan-kalachikov, line 100: It would be better to use Promise.allSettled here to make modules' types be generated in parallel.
Something like that:
```ts
 await Promise.allSettled(installedIndependentModules.map((module) => {
	const moduleTypesPath = `${module.apiPath}/types.ts`;
	return generate( {
	        schema: SCHEMA_PATH,
	        documents: addExtension(module.apiPath),
	        generates: {
	          [moduleTypesPath]: {
	            plugins: GENERAL_PLUGINS,
	            config: GENERAL_CONFIG,
	          },
	        },
	      },
	      true,
	    );
})

ivan-kalachikov (changes requested):

scripts/graphql-codegen/generator.ts
asvishnyakov: It will not require you to cast all types which are duplicated, but also increase bundle size

PR #1267, opened 2024-08-26
chore: consolidate apollo clients and links

ivan-kalachikov:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-1706
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.4.0-pr-1267-75d9-75d92e9f.zip

### DIFF
--- a/client-app/core/api/graphql/client.ts
+++ b/client-app/core/api/graphql/client.ts
@@ -1,41 +1,13 @@
-import { ApolloClient, InMemoryCache } from "@apollo/client/core";
-import { cache, deprecatedLink, link, options } from "@/core/api/graphql/config";
-
-const fetchPolicy = "no-cache";
+import { ApolloClient } from "@apollo/client/core";
+import { cache, link, options } from "@/core/api/graphql/config";
 
 /**
- * Non-cached version of Apollo Client
- *
- * @deprecated
- *
- * Use {@link apolloClient} instead.
- *
- * **IMPORTANT!** Because of cache usage, query deduplication and `@vue/apollo-composable` usage,
- * you can't just replace {@link graphqlClient} with {@link apolloClient} in your code.
- * Look at https://github.com/VirtoCommerce/vc-theme-b2b-vue/pull/868 for more information.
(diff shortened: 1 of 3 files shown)


client-app/core/api/graphql/client.ts
```diff
@@ -1,41 +1,13 @@
...
- *
- * Use {@link apolloClient} instead.
- *
- * **IMPORTANT!** Because of cache usage, query deduplication and `@vue/apollo-composable` usage,
- * you can't just replace {@link graphqlClient} with {@link apolloClient} in your code.
- * Look at https://github.com/VirtoCommerce/vc-theme-b2b-vue/pull/868 for more information.
- */
-export const graphqlClient = new ApolloClient({
-  ...options,
-  link: deprecatedLink,
```
ivan-kalachikov, line 19: We can safely use `link` instead of `deprecatedLink`, because the first one differs by having `cartLink` and condition use  `wsLink` for subscriptions and `httpLink` for others.
Cart link has a defined list of operations it will be applied for. So it doesn't influence anything else.
The second also works perfectly well, because all subscriptions should use `wsLink` (actually we have only one which already use our "new" client.

client-app/core/api/graphql/client.ts
```diff
@@ -1,41 +1,13 @@
...
-
-  cache: new InMemoryCache({
-    addTypename: false,
```
ivan-kalachikov, line 22: `addTypename`  will slightly influence the volume of network traffic, because requests and responses include __typename field. But this increasing can be considered negligible.

client-app/core/api/graphql/client.ts
```diff
@@ -1,41 +1,13 @@
...
-  }),
-
-  queryDeduplication: false,
-
-  defaultOptions: {
-    watchQuery: { fetchPolicy },
```
ivan-kalachikov, line 28: cache is disabled by types for operations used old client, in the cache settings `client-app/core/api/graphql/config/cache.ts`

PR #1282, opened 2024-09-05
feat: quotes as feature flag

NaMax66:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-751
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.5.0-pr-1282-40d2-40d28b46.zip


client-app/shared/account/components/account-navigation.vue
```diff
@@ -100,6 +93,13 @@
         :text="$t('shared.account.navigation.links.notifications')"
         class="notifications-icon"
       />
+
+      <AccountNavigationLink
```
ivan-kalachikov, line 97: If I'm not mistaken, you wanted the main app to know as less as possible about modules (ideally nothing)
maybe account-navigation can be extended by Teleport here?

client-app/pages/cart.vue
```diff
@@ -113,7 +113,7 @@
 
         <!-- Create quote widget -->
         <VcWidget
```
ivan-kalachikov, line 115: Can we remove that widget and add it from the module by some kind of "extension point" with vue teleport?

client-app/shared/account/components/account-navigation.vue
NaMax66: Why not. But it will not be always in the DOM
I dont see how to do it. Let me know

client-app/pages/cart.vue
NaMax66: Show me how :)

client-app/shared/account/components/account-navigation.vue
ivan-kalachikov: I'll try to figure it out a bit later, will let you know if I'll get it :)

ivan-kalachikov (approved):

PR #1324, opened 2024-09-20
feat: create cart extension point

NaMax66:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-1776
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.6.0-pr-1324-e254-e2546179.zip


client-app/shared/cart/composables/useCartExtensionPoints.ts
```diff
@@ -0,0 +1,26 @@
...
+  id: string;
+  /* eslint-disable @typescript-eslint/no-explicit-any */
+  element: DefineComponent<Record<string, any>, Record<string, any>, any>;
+};
+
+function _useCartExtensionPoints() {
+  const sidebarWidgets = ref<ElementType[]>([]);
+
+  function registerSidebarWidget(element: ElementType) {
+    if (!sidebarWidgets.value.find((el) => el.id === element.id)) {
```
ivan-kalachikov, line 15: better use `some` instead of `find` as we don't need an element here, just checking if the array contains that

PR #1364, opened 2024-10-09
fix: pagination reset after filters apply in category

NaMax66:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-1926

### DIFF
--- a/client-app/shared/catalog/components/category.vue
+++ b/client-app/shared/catalog/components/category.vue
@@ -124,6 +124,7 @@
               :items="PRODUCT_SORTING_LIST"
               class="w-0 grow lg:w-48"
               size="sm"
+              @change="resetCurrentPage"
             />
           </div>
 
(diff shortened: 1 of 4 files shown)


client-app/shared/catalog/components/category/category-controls.vue
```diff
@@ -60,6 +65,7 @@ defineProps<IProps>();
 
 interface IEmits {
   (event: "openBranchesModal", value: boolean): void;
+  (event: "applySort"): void;
```
ivan-kalachikov, line 68: why `applySort`?

client-app/shared/catalog/components/category.vue
```diff
@@ -141,6 +142,7 @@
             :loading="fetchingProducts"
             :saved-branches="localStorageBranches"
             @open-branches-modal="openBranchesModal"
+            @apply-sort="resetCurrentPage"
```
ivan-kalachikov, line 145: maybe it's better to move page resetting from event handlers to existing watcher? there is one which looks like it covers all these cases of sorting, filtering etc.

```ts
watchDebounced(
  computed(() => JSON.stringify(searchParams.value)),
	...
  ```

NaMax66: Not all. Show in stock don't go into searchParams

ivan-kalachikov (approved):

PR #1373, opened 2024-10-10
chore: change eslint warn to error for some ts rules

ivan-kalachikov:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-1888
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.7.0-pr-1373-5411-54116ece.zip


client-app/pages/product.vue
```diff
@@ -369,11 +368,11 @@ watchEffect(async () => {
     await fetchRelatedProducts({ productId: productId.value, itemsPerPage: 30 });
   }
 
-  const recommendedProductsBlocks = (recommendedProductsSection.value?.blocks ?? []) as { model: string }[];
-  if (!recommendedProductsSection.value?.hidden && recommendedProductsSection.value?.blocks.length) {
+  const recommendedProductsBlocks = recommendedProductsSection.value?.blocks?.filter((block) => !block.model) ?? [];
```
NaMax66, line 371: you sure new logic should not be tested by qa?

ivan-kalachikov: Great that you paid attention to this line, because I made a mistake in condition. I fixed the condition and check recommendation blocks on myself - they work fine. Block with no model won't be showed anyway, because return no products, so I do filtration to avoid extra network request.

goldenmaya (approved):
I really hope that the pr will be checked by QA

asvishnyakov (commented):
Most of rules set to "warn" or "off" are actually rules included into larger rulesets (i.e. "plugin:vue/vue3-recommmeded") and it's better to check it and just delete them, rather than explicitly set to "error", as they may experience changes, such as name change or parameters may be added

ivan-kalachikov:
> Most of rules set to "warn" or "off" are actually rules included into larger rulesets (i.e. "plugin:vue/vue3-recommmeded") and it's better to check it and just delete them, rather than explicitly set to "error", as they may experience changes, such as name change or parameters may be added

That's a good note, thank you. I'll create a separate task.

PR #1412, opened 2024-10-28
fix: localization support in storybook

asvishnyakov:
### Before
Screenshot of storybook localization before

### After
Screenshot of storybook localization after

### What else changed
Fixed storybook support for typescript and eslint

## Possible regression
Language loading & switching on site

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,8 +1,9 @@
 .deployment
 .github
 .husky
+!.storybook
 .vscode
-.idea
+.yarn
 artifacts
 assets
 node_modules
--- a/.storybook/mocks.ts
+++ b/.storybook/mocks.ts
@@ -1,0 +1,69 @@
+import type { CurrencyType, LanguageType, StoreResponseType } from "@/core/api/graphql/types";
+
+export const defaultLanguage: LanguageType = {
+  isInvariant: false,
+  cultureName: "en-US",
+  nativeName: "English (United States)",
+  threeLetterLanguageName: "eng",
+  twoLetterLanguageName: "en",
+  twoLetterRegionName: "US",
+  threeLetterRegionName: "USA",
+};
+
+export const defaultCurrency: CurrencyType = {
+  code: "USD",
+  symbol: "$",
+  exchangeRate: 1.0,
+  englishName: "US Dollar",
+  cultureName: "en-US",
+};
+
+export const store: StoreResponseType = {
+  availableCurrencies: [defaultCurrency],
+  availableLanguages: [
+    defaultLanguage,
+    // Add other supported languages here
+    {
+      isInvariant: false,
+      cultureName: "fr-FR",
+      nativeName: "français (France)",
+      threeLetterLanguageName: "fra",
+      twoLetterLanguageName: "fr",
+      twoLetterRegionName: "FR",
+      threeLetterRegionName: "FRA",
+    },
+    {
+      isInvariant: false,
+      cultureName: "de-DE",
+      nativeName: "Deutsch (Deutschland)",
+      threeLetterLanguageName: "deu",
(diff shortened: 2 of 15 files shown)


goldenmaya (dismissed):

client-app/core/composables/useLanguages.ts
```diff
@@ -10,134 +10,143 @@ import type { I18n } from "@/i18n";
...
     }
-
-    Logger.debug(`The "${moduleName}" module locale: "${locale}" was loaded successfully.`);
-  } catch (error) {
-    Logger.error(`Error loading the ${moduleName} module locale: "${locale}"`);
   }
-}
 
-export function useLanguages() {
+  const fallback = {
```
NaMax66, line 140: refactoring in the separate pull request

client-app/core/composables/useLanguages.ts
```diff
@@ -10,134 +10,143 @@ import type { I18n } from "@/i18n";
 import type { LocaleMessage } from "@intlify/core-base";
 import type { Composer } from "vue-i18n";
 
-const { themeContext } = useThemeContext();
+function _useLanguages() {
+  const { themeContext } = useThemeContext();
 
-const pinedLocale = useLocalStorage<string | null>("pinedLocale", null);
+  const pinedLocale = useLocalStorage<string | null>("pinedLocale", null);
```
NaMax66, line 16: what the problem with this file? Only because of createGlobalState? Move it to refactoring with language fallback pull request

.storybook/mocks.ts
```diff
@@ -0,0 +1,69 @@
+import type { CurrencyType, LanguageType, StoreResponseType } from "@/core/api/graphql/types";
```
NaMax66, line 1: separate pull request

.prettierignore
```diff
@@ -1,8 +1,9 @@
 .deployment
 .github
 .husky
+!.storybook
 .vscode
-.idea
+.yarn
```
NaMax66, line 6: how is it related to task?

NaMax66 (changes requested):

.storybook/mocks.ts
asvishnyakov: These mocks are required for
https://github.com/VirtoCommerce/vc-frontend/blob/f7fce67749006daffb78352e7545f77ec7dc21d5/.storybook/preview.ts#L17
which is part of bug fix (otherwise it will throw exception "No theme context was found")

.prettierignore
asvishnyakov: > What else changed
Fixed storybook support for typescript and eslint

I found and fixed an issue when linting didn't work for files in `.storybook` folder because of issue in eslint (there is a comment about that). This change just make file exceptions specified in `prettier` exactly the same as for `eslint`

client-app/core/composables/useLanguages.ts
asvishnyakov: Fallback is used at
https://github.com/VirtoCommerce/vc-frontend/blob/f7fce67749006daffb78352e7545f77ec7dc21d5/.storybook/preview.ts#L17
so alternative is to copy-paste it in this PR, which I would like to avoid

client-app/core/composables/useLanguages.ts
```diff
@@ -152,6 +161,7 @@ export function useLanguages() {
         throw new Error("currentLanguage is read only.");
       },
     }),
+
     initLocale,
     fetchLocaleMessages,
```
NaMax66, line 166: remove

asvishnyakov: Done

NaMax66:
Will be implemented separately

PR #1454, opened 2024-11-15
feat: ability to disallow user to select cart items

asvishnyakov:
### ⚠️ **Breaking change**
`CartLineItems` now accept `selectable`, so line items can be reused, but without selection feature

Non selectable line items

### DIFF
--- a/client-app/shared/cart/components/cart-line-items.vue
+++ b/client-app/shared/cart/components/cart-line-items.vue
@@ -11,7 +11,7 @@
     with-total
     with-subtotal
     removable
-    selectable
+    :selectable="selectable"
     @select:items="$emit('select:items', $event)"
     @remove:items="$emit('remove:items', $event)"
   >
@@ -71,6 +71,7 @@ interface IProps {
   readonly?: boolean;
   items?: LineItemType[];
   validationErrors?: ValidationErrorType[];
+  selectable?: boolean;
   sharedSelectedItemIds?: string[];
 }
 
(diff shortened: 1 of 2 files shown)


goldenmaya (approved):

client-app/shared/cart/components/cart-line-items.vue
```diff
@@ -71,6 +71,7 @@ interface IProps {
   readonly?: boolean;
   items?: LineItemType[];
   validationErrors?: ValidationErrorType[];
+  selectable?: boolean;
```
NaMax66, line 74: set default true
then in your code selectable="false"

asvishnyakov: That's keep backward compatibility for sure, but will against our own and vue's conventions where boolean props are always false by default to allow their use with other components as just
```html
<input property>
```
instead of
```html
<input property="value">
```

asvishnyakov: And you already accepted https://github.com/VirtoCommerce/vc-frontend/pull/1446 with the same change

ivan-kalachikov: @asvishnyakov 
Couldn't find the explicit mention that default value shouldn't be `true`. I mean yes, it makes sense, especially when creating new component. But don't see any problems in doing that when modifying existing component to avoid breaking changes. How could it harm?

asvishnyakov: @ivan-kalachikov It couldn't, it cost me nothing, I just can't understand then why you accepted #1446 as is, while asked to change this one.

NaMax66 (approved):

PR #1561, opened 2025-01-31
feat: upload images for product review

Andrew-Orlov:
## Description

Upload images for product review

### Jira-link:

https://virtocommerce.atlassian.net/browse/VCST-2462


client-app/modules/customer-reviews/components/product-reviews.vue
```diff
@@ -45,6 +45,22 @@
           <div>
             {{ review.review }}
           </div>
+
+          <div v-if="review.images?.length" class="image-gallery__thumbs-container">
+            <Swiper
+              :slides-per-view="MAX_REVIEW_IMAGES_COUNT"
```
NaMax66, line 51: what if hey 10? You want to show all of them in a row?

Andrew-Orlov: We have 5 images limit on the backend right now. BTW I'll add a Prev/Next navigation. 5 images are too much for mobile version.

client-app/modules/customer-reviews/components/product-reviews.vue
```diff
@@ -45,6 +45,22 @@
           <div>
             {{ review.review }}
           </div>
+
+          <div v-if="review.images?.length" class="image-gallery__thumbs-container">
```
goldenmaya, line 49: According to BEM methodology, an element cannot be used separately from its block.  
Since we are not using the corresponding BEM block here, we should apply TailwindCSS classes instead.  
Please replace the BEM-specific styles with Tailwind utilities.

goldenmaya (changes requested):

client-app/modules/customer-reviews/components/product-reviews.vue
```diff
@@ -172,6 +232,12 @@ const productReviewsPayload = ref({
   page: 1,
   sort: "createddate:desc",
 });
+const imagesSwiper = ref<SwiperCore | null>(null);
+const swiperModules = [Navigation, Thumbs];
+
+const isMobile = breakpoints.smaller("lg");
+
+const displayedReviewImagesCount = computed(() => (isMobile.value ? 2 : (imageOptions.value?.maxFileCount ?? 5)));
```
ivan-kalachikov, line 240: what if `imageOptions.value?.maxFileCount` is `20`?
maybe it's better `Math.min(imageOptions.value?.maxFileCount ?? 5, 5)`

ivan-kalachikov: move 5 to the constant

client-app/modules/customer-reviews/components/product-reviews.vue
```diff
@@ -45,6 +45,40 @@
...
+          <div v-if="review.images?.length" class="flex items-center justify-between">
+            <VcNavButton
+              v-if="review.images.length > displayedReviewImagesCount"
+              :id="`images_${review.id}_prev`"
+              :label="$t('common.buttons.previous')"
+              size="xs"
+              direction="left"
+            />
+
+            <Swiper
```
ivan-kalachikov, line 58: What if not using Swiper for cases when review.images.length <= displayedReviewImagesCount ?
just show simple images for the sake of performance?  (and make swiper components lazy imported so no import when no needs?)

ivan-kalachikov (dismissed):

goldenmaya (dismissed):

ivan-kalachikov (dismissed):

IrinaSer (approved):

PR #1597, opened 2025-02-20
fix: close vc-popover on blur

ivan-kalachikov:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-2456


client-app/shared/catalog/components/category.vue
```diff
@@ -101,10 +101,28 @@
           @click="showFiltersSidebar"
         />
 
+        <!-- View options -->
```
ivan-kalachikov, line 104: since flex order property doesn't influence on the tabindex order, I changed it to be in a natural order it positioned on the screen for desktop, and changed theirs' order programmably for mobile, since there is no sense in navigating by tab key

Andrew-Orlov (approved):

NaMax66 (approved):

Andrew-Orlov (approved):

PR #1619, opened 2025-03-06
feat: file section for configurable products

ivan-kalachikov:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-2796
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.19.0-pr-1619-61d7-61d7b277.zip


client-app/shared/catalog/components/configuration/option-file.vue
```diff
@@ -0,0 +1,95 @@
...
+}
+
+function onFileDownload(file: FileType) {
+  if (file && file.url) {
+    void downloadFile(file.url, file.name);
+  }
+}
+
+watchEffect(async () => {
+  await fetchFileOptions();
```
NaMax66, line 85: why async? 
have you tried watchEffect(fetchFileOptions)?

client-app/shared/catalog/components/configuration/product-configuration.vue
```diff
@@ -84,9 +97,42 @@
...
+                  files: $event,
+                  type: section.type,
+                })
+              "
+            />
+
+            <OptionNone
+              v-if="!section.isRequired"
+              :name="section.id"
+              :selected="selectedConfiguration[section.id]?.selectedOptionTextValue === undefined"
```
NaMax66, line 132: could it be null or other falsy value?

client-app/shared/catalog/composables/useConfigurableProduct.ts
```diff
@@ -73,70 +75,75 @@ function _useConfigurableProduct(configurableProductId: string) {
...
+  function getSelectedOptionTextValue(section: SectionValueType, sectionId: string) {
     const rawSection = configuration.value.find(({ id }) => id === sectionId);
     switch (rawSection?.type) {
       case CONFIGURABLE_SECTION_TYPES.text:
         return section.customText;
       case CONFIGURABLE_SECTION_TYPES.product:
-        return rawSection.options?.find(({ product }) => product?.id === section.option?.productId)?.product?.name;
+        return rawSection.options?.find(({ product }) => product?.id === section.productId)?.product?.name;
+      case CONFIGURABLE_SECTION_TYPES.file:
+        return section.files?.map((file) => file.name).join(", ");
```
NaMax66, line 132: what about toCSV helper?

client-app/shared/catalog/components/configuration/option-file.vue
ivan-kalachikov: Just copy-pasted it from similar places. But you are right async is not needed here as well as a watchEffect, removed that.

client-app/shared/catalog/components/configuration/product-configuration.vue
ivan-kalachikov: it's the required field, there will be always boolean value. Even if there is any other falsy value it's okay - we would consider the section optional unless it's explicitly has required: true

client-app/shared/catalog/composables/useConfigurableProduct.ts
ivan-kalachikov: It was added in the PR which is not merged yet, but yeah, copied it to this PR and used

IrinaSer (dismissed):

NaMax66 (dismissed):

NaMax66 (approved):

PR #1764, opened 2025-06-03
fix: redirect from notifications route if not authenticated

NaMax66:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-3290
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.25.0-pr-1764-5e7b-5e7bf43e.zip


client-app/modules/push-messages/index.ts
```diff
@@ -96,6 +97,11 @@ export async function init(router: Router, i18n: I18n) {
   const isFCMModuleEnabled = isEnabled(PUSH_MESSAGES_MODULE_FCM_ENABLED_KEY);
 
   if (!themeContext.value?.settings?.push_messages_enabled || !isAuthenticated.value) {
+    router.addRoute("Account", {
+      path: "notifications",
+      redirect: { name: ROUTES.SIGN_IN.NAME },
```
ivan-kalachikov, line 102: just a thought - will it work if we just add `meta: { requiresAuth: true },` ?

like that:
```ts
const route: RouteRecordRaw = {
      path: "notifications",
      name: "Notifications",
      component: Notifications,
    	meta: { requiresAuth: true },
    };
    ```

NaMax66: Then we should import the Component. But if we cant use it why import? :)

ivan-kalachikov (approved):

PR #1847, opened 2025-07-28
fix: pickup point popup map layout

ivan-kalachikov:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-3640

### DIFF
--- a/client-app/shared/common/components/google-maps/google-map.vue
+++ b/client-app/shared/common/components/google-maps/google-map.vue
@@ -1,5 +1,5 @@
 <template>
-  <div :id="mapElementId" ref="mapContainer" class="size-full"></div>
+  <div :id="mapElementId" ref="mapContainer" class="google-map"></div>
   <slot />
 </template>
 
@@ -67,3 +67,9 @@ onBeforeUnmount(() => {
   listener?.remove();
 });
 </script>
+
+<style lang="scss">
+.google-map {
+  @apply size-full;
+}
+</style>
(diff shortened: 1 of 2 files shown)


client-app/shared/common/components/google-maps/google-map.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div :id="mapElementId" ref="mapContainer" class="size-full"></div>
+  <div :id="mapElementId" ref="mapContainer" class="google-map-container"></div>
```
goldenmaya, line 2: please add BEM block === filename

client-app/shared/common/components/google-maps/google-map.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div :id="mapElementId" ref="mapContainer" class="size-full"></div>
+  <div :id="mapElementId" ref="mapContainer" class="google-map"></div>
   <slot />
```
goldenmaya, line 3: Could you please clarify the purpose of the slot in this component? 
If the slot isn’t necessary, would it make sense to move the outer wrapper inside the component to keep the markup simpler?
Screenshot 2025-07-31 at 14 46 59

ivan-kalachikov: We need the slot in the google-map component, so as to:
- ensure mount order (child mounting only after parent was mounted)
- make it looks semantically - one component belongs another

.select-address-map-modal__map wrapper has it's own styles specific to that place, don't see a problem with that - is there any?

NaMax66 (dismissed):

NaMax66 (approved):

PR #1932, opened 2025-09-15
fix: delete lineItemId param if there is no such line item in cart

ivan-kalachikov:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-3536
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.31.0-pr-1932-950b-950b9151.zip

### DIFF
--- a/client-app/pages/product.vue
+++ b/client-app/pages/product.vue
@@ -466,10 +466,14 @@ watch(
 );
 
 watch(
-  () => lineItemId,
+  () => [lineItemId, cart],
   () => {
+    if (!cart.value || !lineItemId) {
+      return;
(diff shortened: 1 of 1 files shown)


client-app/pages/product.vue
```diff
@@ -451,6 +464,21 @@ watch(
   },
   { immediate: true },
 );
+
+watch(
+  () => [lineItemId, cart],
+  () => {
+    if (!cart.value || !lineItemId) {
```
goldenmaya, line 471: i think we should check cart loading.value here as well

PR #2245, opened 2026-04-08
fix(VCST-4669): hide reset button when filter is not applied

vas11yev1work:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-4669
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.45.0-pr-2245-706c-706c3fe8.zip


client-app/shared/checkout/components/select-address-map/select-address-map-list.vue
```diff
@@ -66,6 +71,7 @@ interface IProps {
   addresses: PickupLocationType[];
   selectedAddressId?: string;
   selectable?: boolean;
+  filterIsApplied?: boolean;
```
goldenmaya: How about isFiltered or filtered?

goldenmaya (approved):

PR #2278, opened 2026-05-04
feat(VCST-4715): add support for default selection in product configu…

ToxaKZ:
…ration options

## Description
Implemented support for default options in configurable products across Catalog, XCart GraphQL, and Storefront. The feature allows a product configuration option to be marked as default in admin, exposes that flag through XCart GraphQL, and makes the storefront automatically preselect default values, including defaults in dependent configuration sections.

Implementation Details

Updated configurable product default-selection logic.

Changes include:

Refactored updateWithDefaultValues.

Added a pass-based default application flow:

apply defaults for currently visible sections
repeat while new defaults make dependent sections visible
stop when no new defaults are applied
Added helper logic to avoid overriding existing selections.

Default values are now applied only when:

the section is visible
the section has no selected value yet
the section has an option marked as default

This fixes the dependent-section case where selecting a default in one section makes another section visible, and that newly visible section also has its own default option.

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.50.0-pr-2278-c88f-c88fae3a.zip


client-app/shared/catalog/composables/useConfigurableProduct.ts
```diff
@@ -408,25 +408,74 @@ function _useConfigurableProduct(configurableProductId: MaybeRef<string>) {
...
 
   function updateWithDefaultValues() {
-    configuration.value.forEach((section) => {
-      if (!section.isRequired || !isSectionVisible(section.id)) {
-        return;
+    let hasChanges = true;
+    let iterations = 0;
+    const maxIterations = configuration.value.length;
+
+    while (hasChanges && iterations < maxIterations) {
```
OlegoO: Explain me this code )

OlegoO (changes requested):

client-app/shared/catalog/composables/useConfigurableProduct.ts
ToxaKZ: Refactored this part to make the intent clearer.
The loop is still needed because applying a default value in one section can make dependent sections visible, and those newly visible sections may also have defaults. So we repeat the default-applying pass until no new defaults are added, with the number of passes bounded by the section count.

To make it easier to read, I split the logic into small functions:
- updateWithDefaultValues controls the bounded repeat
- applyDefaultsForVisibleSections performs one pass
- canApplyDefaultValue / hasSelectedSectionValue make the conditions explicit

Behavior should stay the same, but the reason for the repeated pass is now more obvious.

OlegoO (approved):

PR #2293, opened 2026-05-18
feat(VCST-4642): frontend initialize application via XAPI

vas11yev1work:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-4642
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.50.0-pr-2293-55fc-55fc80c2.zip


client-app/core/api/graphql/store/queries/initializeApplication/index.ts
```diff
@@ -0,0 +1,60 @@
...
+const CACHE_TTL_MS = 60 * 60 * 1000;
+
+type CachedEntryType = {
+  timestamp: number;
+  data: InitialStoreType;
+};
+
+function readCache(domain: string): InitialStoreType | undefined {
+  try {
+    const raw = localStorage.getItem(CACHE_KEY_PREFIX + domain);
```
goldenmaya: Could we use useLocalStorage from @vueuse/core here?

client-app/core/api/graphql/store/queries/initializeApplication/index.ts
```diff
@@ -0,0 +1,60 @@
+import { InitializeApplicationDocument } from "@/core/api/graphql/types";
+import { graphqlClient } from "../../../client";
+import type { InitializeApplicationQuery } from "@/core/api/graphql/types";
+
+type InitialStoreType = InitializeApplicationQuery["store"];
+
+const CACHE_KEY_PREFIX = "vc:initialStore:v1:";
```
goldenmaya: would you mind moving CACHE_KEY_PREFIX to core/constants/local-storage-keys.ts? All other storage keys live there

client-app/core/api/graphql/store/queries/initializeApplication/index.ts
vas11yev1work: It doesn't make sense. `useLocalStorage` is a composable and it meant to be used in setup. This is a cache helper called from an async function - no reactivity needed. Instanting a `useLocalStorage` per call would add an unnecessary storage listener and watcher on every call

goldenmaya (dismissed):

goldenmaya (approved):

PR #2294, opened 2026-05-18
fix(VCST-4205): fixed save for later saving

DmitryGrishinVirtoworks:
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.49.0-pr-2294-a28b-a28b6fce.zip


client-app/pages/account/list-details.vue
```diff
@@ -154,6 +154,8 @@ interface IProps {
   listId: string;
   hideSettings?: boolean;
   listName?: string;
+  addToCartHandler?: (item: LineItemType, quantity: number) => Promise<void>;
+  addAllToCartHandler?: (items: LineItemType[]) => Promise<void>;
```
vas11yev1work: It seems to me that we shouldn’t pass functions as props. The prop types start describing behavior rather than what should be rendered. Let’s use emits instead, along with a prop that tells us which handler to use. Later, we’ll stop using `ListDetails` in "Saved for later" and remove this branching altogether

DmitryGrishinVirtoworks: Done

vas11yev1work (approved):

PR #2337, opened 2026-06-18
perf: shrink eager bundle (lazy-load route-only SDKs, lodash-es, vendor split)

ivan-kalachikov:
## What
Reduce the JavaScript shipped on first load on every page by deferring SDKs that are only needed by specific routes/interactions, and by letting `lodash` tree-shake.

## Why
Profiling vc-frontend (2.52.0) against a QA backend showed the eager entry carried ~997 KB gz of JS before any route renders. Several heavy libraries were loaded eagerly even though they are route- or interaction-specific:
- `skyflow-js` (payment SDK, ~82 KB gz) shipped on the home page — only needed at checkout/payment. It was pulled in eagerly because the `@/shared/payment` barrel re-exported the Skyflow component, so importing any payment composable/type dragged the SDK along, and the component was shared across several lazy routes (hoisted to the entry).
- `barcode-detector` ponyfill loaded on every page for a scan button whose modal only opens on click.
- `marked` + `nouislider` loaded eagerly though they render only for markdown content / the price-range slider.
- Full `lodash` (CommonJS) was bundled because of `import { x } from "lodash"` / `import _ from "lodash"` usages, which do not tree-shake.

## How
- **Skyflow:** `defineAsyncComponent(() => import(...))` in the three payment-method parents; removed `PaymentProcessingSkyflow` from the `shared/payment/components` barrel so composable/type imports no longer pull the SDK.
- **Barcode:** scanner modal is now an async component loaded on first scan click.
- **marked + nouislider:** dynamic-imported at point of use (markdown render watcher; slider `onMounted`, incl. its CSS).
- **lodash to lodash-es** across the app **and the locale scripts**, so only used methods are bundled; the now-unused `lodash` and `@types/lodash` direct deps are removed (`@types/lodash` remains transitively via `@types/lodash-es`).
- **manualChunks:** keep a shared eager `vendor` chunk for libraries while leaving the dynamically-imported libs out of it so they stay lazy. (Folding libraries into the entry chunk instead produced one long evaluation task that inflated TBT for the same total work — so they're kept as a separate `vendor` chunk.)

## Results
**Bundle (deterministic build output):** eager entry JS **~997 to ~827 KB gz** — full `lodash` removed from the entry; skyflow / barcode / marked / nouislider moved to lazy chunks.

**Core Web Vitals** (Lighthouse, median of 5, same-session interleaved before(2.52.0)/after):

| Page (profile) | Score | FCP | LCP | TBT | SI | CLS |
|---|---|---|---|---|---|---|
| Home (mobile) | 26 → 27 | -19% | -4% | -4% | -11% | 0.29 → 0.29 |
| Catalog (mobile) | 38 → 40 | -21% | -5% | -10% | -9% | 0.06 → 0.06 |
| Product (mobile) | 23 → 27 | -12% | -7% | -16% | -5% | 0.29 → 0.29 |
| Home (desktop) | 52 → 54 | -8% | -2% | -7% | -7% | 0.53 → 0.53 |
| Catalog (desktop) | 51 → 63 | -11% | -20% | -39% | -2% | 0.08 → 0.08 |
| Product (desktop) | 67 → 70 | -7% | -2% | -27% | -6% | 0.13 → 0.13 |

_Legend: **FCP** = First Contentful Paint, **LCP** = Largest Contentful Paint, **TBT** = Total Blocking Time, **SI** = Speed Index, **CLS** = Cumulative Layout Shift. **Score** = Lighthouse performance score (0-100, higher is better). FCP/LCP/TBT/SI shown as % change (lower is better); Score and CLS as before → after._

**Takeaways**
- **Score** up on every page/profile.
- **FCP** down everywhere (**-7% to -21%**) and **TBT** down everywhere (**-4% to -39%**); SI -2% to -11%.
- **CLS unchanged** — this PR makes no CLS-affecting changes (loading-only); per-row differences are run-to-run noise.

### How it was measured
- **Tool:** Lighthouse 12.8 CLI, `--only-categories=performance`. Mobile = simulated **Slow 4G + 4x CPU**; desktop = `--preset=desktop` (no throttle). Pages served from a production `vite preview` build against a live QA backend.
- **Interleaved, same-session, one quiet machine:** both builds run on separate ports and Lighthouse alternates **pristine → optimized, one run at a time**, so both branches sample the same backend window with no CPU contention. **Median of 5** per metric (rejects outliers). Catalog is the noisiest page (each load round-trips the remote GraphQL backend); the median absorbs it.

## Test plan
- [x] `yarn validate:types` (vue-tsc; covers app + `scripts/`) passes.
- [x] Production build succeeds; eager entry ~997 to ~827 KB gz; deferred libs verified absent from the eager chunks (sourcemap), present only in their lazy chunks.
- [x] Skyflow lazy chunk loads at the payment step; barcode lib loads on scan click.
- [x] Same-session interleaved before/after Lighthouse matrix (above).
- [x] `yarn check-locales` runs clean after the lodash-es script migration.

> No logic changes — deferrals are loading-only. Markdown-dominant CMS pages and the Skyflow payment step were not in the measured matrix (home/catalog/product) and may paint the deferred content one chunk-fetch later on a cold cache. GTM deferral, responsive/next-gen images, and a GraphQL-document split were evaluated and intentionally left out of this PR.

> **On Skyflow lazy-loading and deploys:** moving Skyflow from eager to lazy does not introduce a new failure class. Route components are already lazy (`() => import()`) with no global chunk-load-error handler, so the stale-chunk-after-deploy case already applies app-wide on every navigation. The Skyflow chunk ships in the same build as the lazy checkout-route chunk that must load first, so they share the same purge fate (the route gates before Skyflow is ever fetched). Global stale-chunk resilience (e.g. a `vite:preloadError` / `router.onError` reload) is a pre-existing, app-wide concern, intentionally out of scope here.


ivan-kalachikov:
Superseded by #2345 - moved the branch onto the upstream repo (off the fork) so the CI/Sonar job runs (theme-ci.yml gates it to non-fork PRs).

PR #2349, opened 2026-06-23
fix(VCST-5202): Skyflow card form per-brand CVV length validation

Dan-BV:
## VCST-5202 — Skyflow per-brand CVV length validation

**DO NOT MERGE until human review.** Opened by the `/qa-fix` autonomous pipeline (Gates G2–G4 passed). Needs deploy/visual verification (G6).

### Problem
The Skyflow card form used a single brand-agnostic CVV rule `^[0-9]{3,4}$`, so a 3-digit CVV on American Express (requires 4) passed client validation and enabled **Place order** (and the mirror: a 4-digit CVV on Visa/Mastercard passed). Brand detection existed but was display-only.

### Fix
- New pure helper `client-app/shared/payment/utils/skyflow-cvv-validation.ts` — `getCvvValidation(cardScheme)`: Amex → `^[0-9]{4}$` / placeholder `1111`, all others/unknown → `^[0-9]{3}$` / `111` (case-insensitive).
- `payment-processing-skyflow.vue`: removed the static `CVV_REGEX`. New-card CVV element is updated live via `ComposableElement.update({ validations, placeholder })` on the CARD_NUMBER `CHANGE` event (`selectedCardScheme`), guarded to fire only when the rule changes. Saved-card CVV derives the rule from the card's `cardScheme ?? cardType`.
- `getSkyflowCardsQuery.graphql`: added `cardScheme`/`cardType` to the saved-cards selection set so the saved-card path is functional (previously always-undefined → silent no-op).
- `core/api/graphql/types.ts`: **codegen output only** (regenerated via `yarn generate:graphql-types`); the diff is exactly the two Skyflow fields — NOT hand-edited.

### Tests
- New `skyflow-cvv-validation.test.ts` (5 cases) — RED→GREEN for both new-card and saved-card brand→rule derivation. All pre-existing tests/stories unchanged. Local: typecheck, lint, build pass; vitest green except a pre-existing unrelated catalog/nav hook-timeout flake.

### Verification needed at deploy (G6)
The in-iframe live behavior can't be unit-asserted through jsdom. On the deployed QA env, confirm per-brand CVV for **both** a NEW Amex card and a SAVED Amex card (regression case PAY-SKY-015).

Linked: 
https://virtocommerce.atlassian.net/browse/VCST-5202
Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.53.0-pr-2349-1b81-1b8104ba.zip


goldenmaya (dismissed):

Dan-BV:
## Runtime fix pushed — Amex brand detection wiring (VCST-5202)

**G6 was found FAILING on the prior build** (`2.53.0-pr-2349`) against qa: a new Amex card `378282246310005` with a **3-digit** CVV `123` was still accepted (Place order stayed enabled), and a **4-digit** CVV `1234` was rejected — i.e. the CVV rule was stuck at exactly-3 for Amex. The helper `getCvvValidation` was correct in isolation (its unit tests passed), so the defect was purely in the runtime wiring.

### Real root cause (SDK-grounded)
The new-card handler read `selectedCardScheme` from the Skyflow `CARD_NUMBER` `CHANGE` event. Per the **installed `skyflow-js@2.7.9`** (`node_modules/skyflow-js/README.md`, line 1256):

> `selectedCardScheme` will exist for `CARD_NUMBER` element state and the value of `Skyflow.CardType` **will be only populated when cardbrand choice selection is triggered**, otherwise it will always be an empty string.

Card-brand-choice selection only fires for co-badged/multi-scheme cards. A plain, auto-detected Amex never triggers it, so `selectedCardScheme` stayed `""` → `getCvvValidation("")` → the 3-digit default. That is exactly why unit tests (which call `getCvvValidation("AMEX")` directly) passed while the live form failed.

### What changed
- **New-card path:** derive the brand from the card number's **IIN prefix** (`34`/`37` → Amex) instead of `selectedCardScheme`. The SDK returns the leading digits of the `CARD_NUMBER` `value` even in PROD (first 6 for Amex, per the same README note + the internal masking logic in `dist`), so `getCardSchemeFromNumber(value)` resolves Amex reliably. `selectedCardScheme` is kept as a higher-priority signal for the co-badged case: `selectedCardScheme || getCardSchemeFromNumber(value)`.
- **Robust matching:** `getCvvValidation` now normalizes the scheme (letters-only, uppercase) and accepts both `"AMEX"` (SDK enum) and `"AMERICAN EXPRESS"` (the Skyflow vault `card_scheme` the backend passes through verbatim for **saved** cards — confirmed in `vc-module-skyflow` `SkyflowClient.cs`, which reads `card_scheme`/`card_type` from the vault).

### Gates (local, green)
`vitest` 11/11 · `eslint` 0 errors · `vue-tsc` typecheck OK · production `build` OK.

### Note on verification
The true defect is integration-level and cannot be fully proven in vitest/jsdom (that is how this PR shipped green-but-broken). This change is the SDK-grounded code fix + green unit gate. **Please rebuild the theme and re-run the live Amex check on qa** (new Amex `378282246310005`: 3-digit CVV must be rejected / Place order disabled; 4-digit accepted).

Dan-BV:
## Fix: stale CVV validity across a brand change (VCST-5202)

**Confirmed correctness bug** (independent of the input-length/keystroke-block discussion): in the new-card form, typing a **3-digit CVV first**, then an **Amex** number (`378282246310005`), left **Place order clickable** — a 3-digit CVV is invalid for Amex.

### Root cause (skyflow-js@2.7.9)
On the CARD_NUMBER CHANGE → Amex, the handler swaps the CVV rule via `CVV.update({ validations: ^[0-9]{4}$ })`. But the composable element's `update()` takes only `CollectElementUpdateOptions` (no value — `types/core/external/collect/compose-collect-element.d.ts` + `types/utils/common/index.d.ts`) and does **not** re-emit a CHANGE for the value already in the field. So the CVV element keeps its **stale `isValid=true`** from the previous `^[0-9]{3}$` rule, `newCardFormElementsStatus[INPUT_FIELD].valid` stays true, `hasInvalid` stays false, and the button (and `usePayment` `setCardDataValid`) stays enabled.

### Fix
Inside the existing once-per-rule-transition guard (only when `regex !== currentNewCardCvvRegex`), drop the tracked CVV validity **before** calling `update()`:
```ts
updateValidationStatus({ elementType: Skyflow.ElementType.INPUT_FIELD, isValid: false });
```
The CVV's validity is restored only when the element emits a fresh valid CHANGE under the new rule, so the button reactively re-enables via the existing `hasInvalid` computed + `watch(isPaymentDataValid)`. The guard ensures this fires only on an actual length change, so the common path (Visa, or card-first-CVV-second) is never spuriously invalidated.

Covers: CVV(3)→Amex ⇒ disabled; Amex→CVV(4)→switch to Visa ⇒ disabled (rule `^4$`→`^3$` re-fires the reset); Amex→CVV(4) and Visa→CVV(3) still enable normally.

### Gates (local, green)
`vitest` **12/12** · `eslint` **0 errors** · `vue-tsc` typecheck OK · production `build` OK.

### Verification note
jsdom cannot exercise the Skyflow iframe integration, so the live re-verify of both CVV/card entry orders + the Amex→Visa mirror is done on the rebuilt theme. The added unit test asserts the util precondition the fix depends on (Amex regex is length-distinct from non-Amex/default; a non-Amex brand does not change the rule).

Head: `ea31a0c2`.

Dan-BV:
## ✅ QA Gate-6 verified — ready for review

The "needs deploy/visual verification (G6)" note in the PR description is now satisfied. Verified live on the PR theme build (`2.53.0-pr-2349-ea31a0c2`) served locally and proxied to the real **vcst-qa** backend/config.

**Verified behavior (Skyflow new-card form):**

| Scenario | Card | CVV | Place order | Result |
|---|---|---|---|---|
| Card-first | Amex | 3-digit | disabled + error | ✅ |
| Card-first | Amex | 4-digit | enabled | ✅ |
| Card-first | Visa | 3-digit | enabled | ✅ |
| Card-first | Visa | 4-digit | disabled + error | ✅ |
| **CVV-first → Amex** | Amex | 3-digit (stale) | **disabled** | ✅ |
| CVV-first → Amex → 4 | Amex | 4-digit | enabled | ✅ |
| **Amex → switch to Visa** | Visa | 4-digit (stale) | **disabled** | ✅ |

Placeholder flips `111` <-> `1111` with the detected brand. No order was submitted.

**Root cause (for reviewers):** the SDK's `selectedCardScheme` is only populated on explicit card-brand-choice (co-badged cards) and is empty for an auto-detected brand -> brand is now derived from the card-number IIN (`34/37` -> Amex). Additionally, `update({ validations })` does not re-validate an already-entered CVV, so the tracked CVV validity is reset when the required length changes (fixes the CVV-first / brand-switch permutations).

**CI:** `ci`, `auto-tests` (mysql/postgres/sqlserver), SonarCloud QG, CLA — all green. JIRA VCST-5202 -> In Review.

**DO NOT MERGE until human review** (no auto-merge).

goldenmaya (approved):

PR #2350, opened 2026-06-23
VCST-5177: Configurable products sorting

yuskithedeveloper:
## Description
The catalog "Sort by" dropdown is now built from the store's configured sort orderings (returned by the API) instead of a fixed, hardcoded list. It respects the configured order, hides options the store has hidden, applies the store's default, and stays in sync with the `?sort` URL parameter (shareable/deep-linkable). If the backend doesn't return `sortings` (older backend, or before the first fetch), the previous translated hardcoded list is used as a fallback.
Also includes a couple of small code-quality cleanups.

The "Sort by" dropdown follows the backend's `selected` flag, so it renders blank during initial load and when the active `?sort` maps to no visible option. Hidden sortings stay resolvable by code: a manually-edited URL (e.g. `?sort=name-ascending` for a hidden option) still sorts the results, but the dropdown shows no selection since the option isn't pickable in the UI — the same goes for unknown codes and raw expressions.

### Breaking changes⚠️
This PR adds `sortings` field to the `SearchProducts` operation in `searchProductsQuery.graphql`. That field only exists on backends carrying VCST-5177 (catalog + x-catalog).  
✅ Old frontend + new backend — works (new field is additive; old query doesn't request it).
❌ New frontend + old backend — breaks (query references a field the old schema lacks → validation error).

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5177
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.52.0-pr-2350-65f6-65f6dd13.zip


client-app/shared/catalog/composables/useProductSortings.ts
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Two-way binding for the sort control that tolerates an unknown or now-hidden `?sort` code: when the current value
+ * matches no visible option, the control presents the default ("") instead of rendering blank. Writes flow straight
+ * through to the route param (which drops `?sort` for the default).
+ */
+export function useSelectedSortOption(sortQueryParam: WritableComputedRef<string>) {
+  const { sortList } = useProductSortings();
+
+  return computed<string>({
+    get: () => (sortList.value.some((option) => option.id === sortQueryParam.value) ? sortQueryParam.value : ""),
```
goldenmaya: The dropdown uses selectedSort, which turns unknown codes into "" (useProductSortings.ts:54). But the request still uses the raw sortQueryParam.value (category.vue:513). So if ?sort has a code that works on the backend but is not in the store's list, the dropdown shows "default", but the products are really sorted by that code. Should searchParams.sort use selectedSort instead? Maybe I missed something.

client-app/shared/catalog/composables/useProductSortings.ts
```diff
@@ -0,0 +1,59 @@
...
+import type { ProductSortingType } from "@/core/api/graphql/types";
+import type { WritableComputedRef } from "vue";
+
+export interface IProductSortingOption {
+  id: string;
+  name: string;
+}
+
+// Module-level singleton, populated from the products search response (sortings field).
+const productSortings = ref<ProductSortingType[]>([]);
```
goldenmaya: productSortings is a module-level singleton. It is written on every withFacets: true fetch (useProducts.ts:416), also from pages/product.vue:259, which writes it but never reads it. So the product page can overwrite the category's sort list. Other state like facets is kept per useProducts instance — should sortings be the same?

client-app/shared/catalog/composables/useProductSortings.ts
yuskithedeveloper: Got it fixed and updated PR description (see "The "Sort by" dropdown follows...").

client-app/shared/catalog/composables/useProductSortings.ts
yuskithedeveloper: Refactoring implemented

goldenmaya (dismissed):

goldenmaya (approved):

PR #2362, opened 2026-07-02
feat(VCST-4984): extend vc-table component with default empty/error states & single/multiple row selection

goldenmaya:
### Known limitations / follow-ups
- **Row selection requires the default table header.** When `selectionMode` is set together with VcTableColumn column-slots, the body injects a leading selection cell, but the selection `<th>` (and the `multiple`-mode select-all) only lives in the default `thead`. Providing a `#header` slot or `hideDefaultHeader` therefore drops the header selection column while body rows keep it (off-by-one column misalignment for `#header`; unheaded column + no select-all for `hideDefaultHeader`). Not hit by any current caller (selection is new). Tracked as a follow-up: [VCST-5472](https://virtocommerce.atlassian.net/browse/VCST-5472).

### Jira-link:

https://virtocommerce.atlassian.net/browse/VCST-4984
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.53.0-pr-2362-c1c1-c1c1c519.zip

---

> [!NOTE]
> **Medium Risk**
> Touches a shared table primitive with new selection state, keyboard/a11y behavior, and error/retry flows; risk is mitigated by broad unit tests but adopters must wire `v-model:selection` and error handling correctly.
> 
> **Overview**
> **VcTable** gains built-in **empty** and **error** UX: default `VcEmptyView` (search / new error illustration), optional **`@retry`** when a listener is bound, a shared **`#error`** slot, and pagination/footer hidden while **`error`** is true. Empty still allows **`#desktop-empty`** / **`#mobile-empty`** overrides.
> 
> **Row selection** is new via **`selectionMode`** (`single` | `multiple`), **`v-model:selection`**, **`isRowSelectable`**, **`selectionChange`** metadata, a leading checkbox/radio column (with select-all for multiple), selected-row styling (including a dark-theme token), and **`#desktop-item`** / **`#mobile-item`** scope (`selected`, `toggle`, `selectable`). Selection is not injected when **`#desktop-body`** is used (DEV warning). Keyboard roving tabindex, Space/Enter behavior, and numeric key normalization are included.
> 
> Supporting changes: **`VcEmptyView`** **`error`** variant + **`load-error.svg`**, **`VcRadioButton`** **`tabindex`**, **`ui_kit.table.*`** i18n, Storybook stories, and a large **`vc-table.test.ts`** expansion.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1c1c5193d51d00169cbeb59956c8c9bd46b1dba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

[VCST-5472]: https://virtocommerce.atlassian.net/browse/VCST-5472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


client-app/ui-kit/components/organisms/table/vc-table.vue
```diff
@@ -136,13 +205,48 @@
...
+
+              <VcRadioButton
+                v-else
+                size="sm"
+                :value="getItemKey(item, rowIndex)"
+                :model-value="isRowSelected(item, rowIndex) ? getItemKey(item, rowIndex) : undefined"
+                :disabled="!isRowSelectable(item)"
+                :aria-label="$t('ui_kit.table.select_row')"
+                @change="toggleRow(item, rowIndex)"
+              />
```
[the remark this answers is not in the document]
goldenmaya: This is intended behavior

client-app/ui-kit/components/organisms/table/vc-table.vue
```diff
@@ -136,13 +205,48 @@
...
+              <VcCheckbox
+                v-if="selectionMode === 'multiple'"
+                size="sm"
+                :model-value="isRowSelected(item, rowIndex)"
+                :disabled="!isRowSelectable(item)"
+                :aria-label="$t('ui_kit.table.select_row')"
+                @change="toggleRow(item, rowIndex)"
+              />
+
+              <VcRadioButton
```
ivan-kalachikov, line 249: **♿ Single-mode radios aren't grouped (missing `name`) — accessibility only, not a functional bug**

**Scope:** mouse selection works correctly (one row at a time, model-managed) — nothing selects wrongly. What's degraded is **accessibility**: the per-row radios all render with `name=""` (confirmed live in the DOM), so the browser treats each as its own one-item group. That means no arrow-key navigation between rows and no *"option X of N"* announcement for screen-reader users. For a shared UI-kit component with an a11y bar, that's worth closing; there's no upside to omitting `name`.

**Proposal:** pass a stable per-table-instance name shared across the rows, e.g. derive one with `useComponentId()` in setup and bind it:
```vue
<VcRadioButton ... :name="selectionRadioName" />
```

ivan-kalachikov (commented):
Deep-dive follow-up on the row-selection logic — 3 confirmed issues, all verified against the real components (single-select radio behavior, radio grouping/a11y, and a stuck-selection edge case). Details inline. Each is localized and fixable without an architectural change.

client-app/ui-kit/components/organisms/table/vc-table.vue
```diff
@@ -547,6 +746,130 @@ function getItemKey(item: T, index: number): string {
...
+  emit("selectionChange", keys, resolveRows(keys), meta);
+}
+
+function toggleRow(item: T, index: number): void {
+  // Slot scope exposes `toggle` even when selection is off; calling it must be a no-op.
+  if (!selectionEnabled.value) {
+    return;
+  }
+
+  if (!isRowSelectable(item)) {
```
ivan-kalachikov: **🔵 Design question, not clearly a bug: a selected row that becomes non-selectable can't be cleared from the table**

This is the most arguable of the three — it may be acceptable **by design**, so flagging it as a contract decision rather than a defect.

**Behavior:** if a row is in `selection` **and** `isRowSelectable` returns `false` for it, there's no in-table way to remove it — `toggleRow` returns early for non-selectable rows (this line) and its control is `:disabled`, while `toggleSelectAll`'s deselect filters against `selectableKeysOnPage` (~L844), which excludes disabled rows. So even "select-all → deselect-all" leaves the key behind. **Verified in the browser** (temporarily seeding an Inactive row as selected): the deselect-all gesture went `3 → 3,1,2,4,5 → 3`, leaving `3` stuck with a checked-but-disabled control.

**Why it's arguable:** the component is internally consistent — non-selectable rows are excluded from the selectable set everywhere — and reaching this state needs an unusual scenario (parent seeds an invalid selection, or the predicate flips *after* selection). It doesn't arise from normal use in this PR.

**The actual decision:** should `isRowSelectable` gate only *adding* to a selection, or both *adding and removing*? Today it gates both, which creates a UI dead-end (a visibly-selected row you can't clear). Most selection UIs let you always *remove*.

**Proposal:** either (a) allow the **deselect** direction even for non-selectable rows (and have deselect-all strip all current-page keys), so `isRowSelectable` only gates *adding*; or (b) keep as-is and **document it as a known limitation** (callers must not seed/keep non-selectable keys). No need to block the PR on this — just make it a conscious choice.

client-app/ui-kit/components/organisms/table/vc-table.test.ts
```diff
@@ -487,3 +494,717 @@ describe("getAriaSort", () => {
...
+  VcRadioButton: {
+    props: ["modelValue", "value", "disabled", "ariaLabel"],
+    emits: ["change"],
+    template: `
+      <button
+        class="radio-stub"
+        :data-checked="String(modelValue === value)"
+        :data-disabled="String(!!disabled)"
+        :aria-label="ariaLabel"
+        @click="$emit('change', value)"
```
ivan-kalachikov: **🟡 This stub masks a real gap: single-mode "click to deselect" doesn't work with the real `VcRadioButton`**

This `VcRadioButton` stub emits `change` on **every** click (`@click="$emit('change', value)"`), but the real component renders a native `<input type="radio">` that only fires `change` when it *becomes* checked. So the test *"clicking the already-selected row deselects it"* (L1047) passes here, yet in the app a re-click fires no `change`, `@change="toggleRow"` (`vc-table.vue:246`) never runs, and the single-mode deselect branch of `toggleRow` (~L819-823) is unreachable.

**Verified in the browser** (this branch's `SelectionSingle` story): select a row → `Selected key: 1`; re-click the same row → still `1`, **zero `change` events fired**; switching to another row works. So selecting/switching is fine — only the *documented* clear-by-reclick is dead.

Worth noting: native radios legitimately can't be un-clicked, so this missing behavior may be perfectly acceptable UX. The concrete, non-arguable problem is that **the story + this test assert it deselects when it doesn't** — a green test for behavior the component lacks.

**Proposal (pick one):**
- If deselect-by-reclick **is** intended → drive the single-mode toggle from `click` instead of `change` in `vc-table.vue` (watch for a double-fire if both handlers stay), and make this stub mirror native radio semantics (emit `change` only on transition *to* checked) so it can't hide regressions — or add one test that mounts the real `VcRadioButton`.
- If it's **not** intended → drop the deselect assertion (L1047) and the story's "clicking the already-selected row deselects it" claim, so the suite stops asserting something untrue.

client-app/ui-kit/components/organisms/table/vc-table.vue
goldenmaya: Good call. Instead of just adding name (incomplete group under pagination + arrows would select on navigation), we built a proper keyboard model: roving tabindex on the <tr>, arrows
  move focus only, Space/Enter select (Enter → rowClick if bound), no wrap. Unique name per radio kills native arrow-select. The real gap — the constant "Select row" label — is fixed via
  aria-selected + a per-row rowSelectionLabel. Verified in-browser + unit-tested.

client-app/ui-kit/components/organisms/table/vc-table.vue
goldenmaya: fixed

client-app/ui-kit/components/organisms/table/vc-table.test.ts
goldenmaya: fixed

goldenmaya:
bugbot run

ivan-kalachikov (dismissed):

ivan-kalachikov (approved):

PR #2371, opened 2026-07-07
fix(VCST-5324): clear stale facet filter on language switch

DmitryGrishinVirtoworks:
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.53.0-pr-2371-33db-33db25d0.zip


client-app/shared/account/composables/useSignMeIn.ts
```diff
@@ -30,6 +30,7 @@ export function useSignMeIn() {
 
       unpinLocale();
       removeLocaleFromUrl();
+      removeFacetsFromUrl();
```
ivan-kalachikov: **This call doesn't actually clean anything during sign-in — the stale facets live inside `returnUrl`, not in the sign-in page URL.**

At this point the browser is on `/sign-in?returnUrl=%2Fcatalog%3Ffacets%3D…`. `removeFacetsFromUrl()` only looks at top-level query params of the current URL, so it finds no `facets` and does nothing. Right after, `signIn` emits `openReturnUrl`, and its handler in `broadcast.ts` navigates with `location.href = getReturnUrlValue() ?? …` — and `getReturnUrlValue()` only validates the hostname, returning the value untouched.

Repro: an anonymous user on `/catalog?facets="COLOR":"Rot"` (German) gets redirected to sign-in, signs into an account whose contact locale is English, and lands right back on `/catalog?facets="COLOR":"Rot"` — the exact stale filter this PR wants to clear. The SSO path (`useIdentityProvider.ts`) forwards the same untouched `returnUrl`, so external sign-in has the same gap.

**Proposal:** strip the `facets` param from the `returnUrl` *value* instead — e.g. parse it inside `getReturnUrlValue()` (or right before the `location.href = …` in the `openReturnUrl` handler) and delete `facets` from its query string. Once that's in place, this call here can simply be removed.

client-app/pages/account/profile.vue
```diff
@@ -141,6 +141,7 @@ const [currencyCode] = defineField("currencyCode");
 function applyLanguage(): void {
   unpinLocale();
   removeLocaleFromUrl();
+  removeFacetsFromUrl();
```
ivan-kalachikov: **This call can never do anything here — the profile page URL never carries a `facets` param.**

`applyLanguage()` only runs inside the success-modal `onClose`, while the browser is on `/account/profile`. The `facets` param is only ever produced on catalog/search pages, so `url.searchParams.has("facets")` is always `false` here and the function exits without touching anything.

The stale facets after a profile language change actually live elsewhere — in other tabs sitting on catalog pages (see my comment on `removeFacetsFromUrl` in `useLanguages.ts`).

**Proposal:** drop this call — it suggests coverage the code doesn't provide. If you'd rather keep the symmetry with the other call sites, a short comment noting it's a no-op on this page would save the next reader some head-scratching.

client-app/core/composables/useLanguages.ts
```diff
@@ -133,11 +134,22 @@ function removeLocaleFromUrl() {
   }
 }
 
+// Facet term values are language-specific strings, so drop them whenever the effective locale
+// changes (language switch, sign-in/out, profile locale change) rather than keep a stale filter.
```
ivan-kalachikov: **Scope note: this cleans the URL only in the tab where the locale change happens — stale facets can survive in other open tabs and in back/forward history.**

After a locale change, other tabs receive either `pageReloadEvent` (a bare `location.reload()`, no URL cleanup) or `dataChangedEvent` (just a warning toast). Tabs on locale-prefixed URLs like `/de/catalog?facets=…` are safe — the URL prefix wins locale resolution on reload, so the facets stay consistent with the page language. But a tab sitting on an *unprefixed* (default-locale) URL like `/catalog?facets=…` keeps its old-language facets while its effective locale changes underneath it (e.g. contact locale applied after sign-in or a profile change). Same story for bfcache/history entries.

Not something this PR has to solve — the per-call-site approach structurally can't reach those URLs. **Proposal:** handle it as a follow-up with a single boot-time check ("this URL was built under a different culture → drop `facets`"), which would cover every current and future locale-change path in one place, including reloaded background tabs.

ivan-kalachikov (commented):
Reviewed the facet-clearing logic across the four locale-change flows. The language-selector and sign-out paths work as intended. Two of the call sites don't reach the facets they're meant to clear, and there's one scope note on multi-tab behavior — details in the inline comments.

client-app/shared/account/composables/useSignMeIn.ts
DmitryGrishinVirtoworks: Done

client-app/pages/account/profile.vue
DmitryGrishinVirtoworks: Done

client-app/core/composables/useLanguages.ts
DmitryGrishinVirtoworks: Done

client-app/core/composables/useLanguages.ts
```diff
@@ -17,6 +18,10 @@ const previousCultureSlug = useSessionStorage<{ cultureName: string; slug: strin
   cultureName: "",
   slug: "",
 });
+// Culture the URL's facets (if any) were built under. Compared against the resolved culture on every
+// initLocale() call (i.e. every full page load/reload) so stale facets are dropped regardless of which
+// tab or flow (language switch, sign-in/out, profile change) changed the effective culture.
+const facetsCultureName = useLocalStorage<string>("facetsCultureName", "");
```
ivan-kalachikov: **The boot-time approach is the right idea 👍 — but `localStorage` here causes a real bug: it's one shared key for all tabs, while tabs can legitimately run different languages. Valid filters get wiped.**

**Concrete cases:**

1. Tab A is on `/de/catalog` (German — the URL prefix wins locale resolution), tab B is on `/catalog?facets="COLOR":"Red"` — default English, filters perfectly valid. Reload tab A → the shared key becomes `de-DE`. Reload tab B → `en-US ≠ de-DE` → tab B's valid English filters are silently wiped. It repeats on every alternating reload between the two tabs.
2. Sign-out broadcasts `pageReloadEvent` to **all** tabs, so every tab races to compare-then-write the same key — whether a given tab keeps its filters depends on reload order.

**Suggested fix: `useSessionStorage`** (like `previousCultureSlug` right above). It's per-tab, which is exactly the semantics needed — "the culture *this tab's* URL was built under" — and it fixes both cases with a one-word change.

**To be upfront, sessionStorage still isn't ideal.** Any storage-based check guesses the facets' language from a side channel, and when that side channel is empty the guess silently passes:

- **Shared links / bookmarks / fresh sessions:** storage is empty on first load, so a link like `/catalog?facets="FARBE":"Rot"` shared by a German-locale user and opened by an English-locale user keeps the stale facets — the same symptom this PR targets, just arriving via a link instead of an in-app switch.
- **In-app navigation:** `initLocale()` only runs on full page loads. A `<router-link>` or CMS menu item pointing at a facet URL from another language bypasses the check entirely.
- The key records "culture at this tab's last page load" and *assumes* the facets were written under it — true today, but nothing enforces it for future code that builds facet URLs.

**Option A (recommended): make the URL self-describing — `&facetsLang=de-DE`.** Whenever facets are written, also write the culture they were built under, and drop both params when the stamp doesn't match the resolved culture. Key points:

- There's effectively one writer to touch: `useRouteQueryParam(QueryParamName.Facets)` in `useProducts` (plus `category-selector.vue` / `search-dropdown.vue`). Both params must go in the **same** `router.push` — two sequential `useRouteQueryParam` writes would lose the first one — so it wants a small dedicated composable wrapping the pair.
- The boot check in `initLocale()` becomes stateless and *simpler* than the current code: stamp missing or ≠ resolved culture → delete both params. All storage tracking is removed.
- For full coverage, the reader side (`useProducts`) should treat a mismatched/missing stamp as "no facets" — that closes the SPA-navigation path that no boot check can see.
- Failure modes are all fail-safe: a link with no stamp or a wrong stamp loses its filters (unfiltered list, one click to reapply) — it never applies stale ones. Shared links, bookmarks, new tabs, and multi-tab multi-locale all work by construction.

My take: **go with Option A in this PR** — it isn't much more work than the storage fix, and it's the only version whose guarantee survives shared links and in-app navigation. If it has to be split, sessionStorage is the acceptable one-word stopgap — but `localStorage` shouldn't ship as-is either way.

**Option B — longer-term thought, with a caveat.** For *dictionary* properties, a language-invariant token already exists in the index (the alias sits in the base, filterable field — see `CatalogDocumentBuilder.IndexCustomProperty`), and the GraphQL facet contract already separates `term` from `label` — so facet URLs *could* be made language-stable (term = alias, label = localized), making this whole cleanup unnecessary for those properties. But it doesn't work for non-dictionary multilanguage ShortText (the VCST-5324 repro — no alias exists there), it needs an aggregation change + reindex in `vc-module-x-catalog`/`vc-module-catalog`, and it pulls in the opposite direction from the culture-aware filter fix already routed for this WI. Not a proposal for this PR — just noting that if language-stable filter URLs ever become a product requirement, the dictionary/alias path is where they'd come from.

DmitryGrishinVirtoworks: Fixed

ivan-kalachikov (approved):

PR #2382, opened 2026-07-15
feat(VCST-4400): implement outline icons

goldenmaya:
## Description
Screenshot 2026-07-17 at 11 25 46
Screenshot 2026-07-17 at 12 30 30

### Jira-link:

https://virtocommerce.atlassian.net/browse/VCST-4400
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.54.0-pr-2382-5916-59165778.zip

---

> [!NOTE]
> **Medium Risk**
> Broad visual change across navigation, checkout, and account surfaces; wrong alias or missing SVG could break icons, but behavior is mostly presentational with tests on the icon resolver.
> 
> **Overview**
> **`VcIcon`** now supports **solid** and **outline** variants (Lucide-style SVGs under `icons/solid` and `icons/outline`), with outline stroke scaling, optional `strokeWidth` / `label` a11y, and theme-driven default variant via new **`icon_variant`** (`solid` | `outline`, default **outline** in settings and boot in `app-runner`).
> 
> Icon coloring shifts from **`fill-*`** to **`text-*`** / `--vc-icon-color` across header, checkout, catalog, and account UI; some glyphs are renamed or aliased (e.g. **`edit` → `pencil`**, compare remove **`trash-2`**, payment selection **`circle-check`**). Chips and status badges that need filled glyphs use **`variant="solid"`** explicitly.
> 
> The icon loader adds **`resolveIcon`**, **`setDefaultIconVariant`**, and alias handling; **`VcIcon`** tests and Storybook stories cover variants and stroke grids. Minor unrelated tweaks: product properties import `PropertyValueTypes` from core GraphQL, map marker uses solid **`cube`**, mobile header icon sizes normalized to 24px.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5916577869332a70f7c455aabbe13c358be5a872. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


client-app/shared/layout/components/header/_internal/bottom-header-link.vue
```diff
@@ -19,14 +19,13 @@
...
         <slot name="icon">
-          <VcIcon v-if="link.icon" :name="link.icon" :size="24" class="mb-0.5 fill-primary" />
+          <VcIcon v-if="link.icon" :name="link.icon" :size="24" color="primary" class="mb-0.5" />
         </slot>
 
         <transition mode="out-in" name="scale">
           <VcBadge
             v-if="count"
             class="absolute -right-3 -top-2 transition-transform"
-            variant="outline"
```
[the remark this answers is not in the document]
goldenmaya: Yes, it was changed

ivan-kalachikov:
## Bundle-size concern + proposed fix (with perf numbers)

### The issue
The two glob declarations in **`client-app/ui-kit/utilities/icons.ts` (lines ~7–15)** — `import.meta.glob(..., { query: "?raw" })`, non-eager — ship **all 2,689 icons** two costly ways:

1. **Eager registry bloat** — the glob still emits a registry of all 2,689 dynamic-import wrappers into the module that imports `icons.ts`. Since `VcIcon` is used app-wide, that lands in the **main `index.js`**. Confirmed: all 2,689 icon paths are baked into the eager bundle.
2. **Chunk explosion** — each SVG becomes its own lazy chunk: **3,016 JS chunks** (2,812 under 2 KB) vs 655 on `dev`.

Names resolve at runtime, so nothing tree-shakes — the full set ships even though the app statically references ~130.

### The fix — 👉 draft PR #2385
Load the icons as **hashed static assets fetched on demand** instead of lazy JS chunks + an eager registry of import wrappers. The registry becomes tiny name→URL strings, SVGs stay static hashed files (fetched + memoized on demand), and a Vite `assetsInlineLimit` guard keeps them from being inlined back into the bundle. Public API (`resolveIcon`/`loadIconRaw`) and `VcIcon` are unchanged.

**Implemented as [draft PR #2385](https://github.com/VirtoCommerce/vc-frontend/pull/2385)** — targets `feat/VCST-4400`, so it's *just* the icon-loading diff. Cherry-pick / fold in as you like.

### Bundle size
| | dev | this PR | PR + fix |
|---|---|---|---|
| main `index.js` (gzip) | 418 KB | 605 KB | **317 KB** |
| eager critical JS (gzip) | 846 KB | 1032 KB | **744 KB** |
| JS chunks | 655 | 3,016 | **327** |

### Performance testing
5 runs/variant, cold cache, Slow 4G + 4× CPU, real backend, medians:

| Metric | dev | this PR | PR + fix |
|---|---|---|---|
| **FCP** | 9.56 s | 10.72 s | **9.06 s** |
| **LCP** | 12.39 s | 13.52 s | **11.92 s** |
| **TBT** | 732 ms | 802 ms | 828 ms |
| **CLS** | 0.263 | 0.263 | 0.262 |

- **FCP tracks eager-JS bytes almost 1:1.** As-is, the PR regresses FCP **+1.16 s** and LCP **+1.1 s** vs `dev`. The fix reverses both — FCP even lands ahead of `dev`.
- **TBT** is a wash (~730–830 ms, overlapping ranges) — main-thread time is dominated by app bootstrap, not icons; on Slow 4G the bottleneck is download, not parse.
- **CLS** unchanged — icons render into fixed-size spans.

### Notes worth knowing (details in #2385)
1. **Memoize per icon.** `import()` is deduped by the module cache; `fetch()` is not — without memoization every `<VcIcon>` instance re-requests the same icon (observed `chevron-right` fetched 20× on the homepage).
2. **`assetsInlineLimit` must exclude icons**, else Vite inlines the <4 KB SVGs as base64 `data:` URIs straight back into the eager registry (`?no-inline` in the query does **not** work).
3. **Tradeoff:** `fetch()` is lower network priority than `import()`. Didn't surface as a regression (icons aren't the LCP element); same-origin, so no CSP impact.

### Verified
- **Types:** `vue-tsc --build` clean · **Unit tests:** 31/31 · **Runtime** (`yarn dev` + `yarn preview`): icons render in outline/solid, correct stroke styling, requests dedupe to 1 per icon, zero console errors.

**Takeaway:** the icon set itself is fine; shipping the whole catalog as an eager JS registry is what regresses first-paint. The `?url` change (#2385) removes that regression with no behavior change.

client-app/shared/account/components/order-status.vue
```diff
@@ -4,11 +4,12 @@
       <VcChip
         :variant="orderStatus?.variant"
         :color="orderStatus?.color || defaultColor"
-        :icon="icon ?? orderStatus?.icon"
         :truncate="truncate"
         size="sm"
         rounded
       >
+        <VcIcon variant="solid" :name="icon ?? orderStatus?.icon" />
```
ivan-kalachikov: **Regression: unmapped statuses now render a ghost "document" icon.**

Previously the icon went through `VcChip`'s `:icon` prop, which is guarded — `vc-chip.vue` renders `<VcIcon v-if="icon" …>`, so nothing showed when there was no icon. Here the `<VcIcon>` is rendered unconditionally, and none of the three callers (`order-card`, `orders-table`, `order-details`) pass `:icon`, so for any status not in `orders_statuses`, `name` is `undefined` → `withDefaults` falls back to `name: "document-text"` → `solid/document-text.svg` renders where before there was nothing.

Unmapped statuses are an explicitly-supported case (hence `orderStatus?.` + `defaultColor`). Not visible in the default theme (its 7 configured statuses all carry an icon), but any platform/custom status outside that set will show the stray glyph.

```suggestion
        <VcIcon v-if="icon ?? orderStatus?.icon" variant="solid" :name="icon ?? orderStatus?.icon" />
```

scripts/check-icon-parity.mjs
```diff
@@ -0,0 +1,18 @@
...
+const base = "client-app/ui-kit/icons";
+const svgs = (dir) =>
+  readdirSync(`${base}/${dir}`)
+    .filter((f) => f.endsWith(".svg"))
+    .map((f) => f.replace(".svg", ""));
+
+const solid = new Set(svgs("solid"));
+const outline = new Set(svgs("outline"));
+
+const orphanOutline = [...outline].filter((n) => !solid.has(n));
```
ivan-kalachikov: **Minor / optional: this warning prints ~2,250 false "orphans" on every run** — and nothing invokes the script (manual-only helper `yarn check:icon-parity`, not in CI or husky), so it's low-stakes.

`outline/` uses Lucide names, `solid/` uses legacy names, bridged by `icon-aliases.ts` — so comparing raw filenames (`!solid.has(n)`) flags nearly the whole outline set as orphans, burying any real signal. Comparing **through the alias map** (`resolveIconName`) instead reports ~149 genuine gaps:

```js
const solidWithoutOutline = [...solid].filter((name) => !outline.has(resolveIconName(name)));
```

client-app/ui-kit/utilities/icons.ts
```diff
@@ -0,0 +1,94 @@
...
+): { loader?: () => Promise<string>; isOutline: boolean } {
+  if (!name) {
+    return { isOutline: false };
+  }
+
+  const effectiveVariant = variant ?? defaultIconVariant;
+  const canonicalName = resolveIconName(name);
+
+  if (effectiveVariant === "solid") {
+    if (solidMap.has(name)) {
```
ivan-kalachikov, line 84: The solid branch matches on the raw `name`, but the alias map (`canonicalName`) is only applied to the outline fallback just below. So `variant="solid"` with a canonical/Lucide name whose solid file is stored under a legacy name misses `solidMap` and silently renders the **outline** glyph instead — e.g. `variant="solid" name="shopping-cart"` returns outline even though a solid `cart` exists.

No current usage hits this (today's solid usages all pass legacy filenames), but the fallback is silent. Consider checking `solidMap.has(canonicalName)` too before falling back to outline.

client-app/ui-kit/components/atoms/icon/vc-icon.vue
```diff
@@ -1,45 +1,75 @@
...
 const _color = computed(() => getColorValue(props.color));
 
-async function loadIcon(name?: string) {
-  icon.value = await loadIconRaw(name);
+const ariaHidden = computed(() => (props.label ? undefined : "true"));
+const role = computed(() => (props.label ? "img" : undefined));
+
+async function loadIcon(name?: string, variant?: IconVariantType) {
+  const { raw, isOutline: outline } = await loadIconRaw(name, variant);
+  icon.value = raw;
```
ivan-kalachikov, line 71: `loadIcon` assigns its result without checking it's still the current icon. If `name`/`variant` change quickly, two loads race — if the earlier one resolves last it overwrites the newer icon, leaving a stale glyph on screen. A request-token guard avoids it:

```js
let latest = 0;
async function loadIcon(name, variant) {
  const token = ++latest;
  const { raw, isOutline: outline } = await loadIconRaw(name, variant);
  if (token !== latest) return; // a newer load started — drop this stale result
  icon.value = raw;
  isOutline.value = outline;
}
```

Low severity — only shows up under rapid programmatic name changes.

client-app/ui-kit/utilities/icon-aliases.ts
```diff
@@ -0,0 +1,78 @@
...
+  cash: "banknote",
+  chat: "message-circle",
+  "check-bold": "check",
+  "chevron-double-left": "chevrons-left",
+  "chevron-double-right": "chevrons-right",
+  "chevron-up-thin": "chevron-up",
+  "circle-outline": "circle",
+  clear: "eraser",
+  "clipboard-arrow": "clipboard-copy",
+  "clipboard-list": "file-text",
```
ivan-kalachikov: `clipboard-list` (and `grid`, L71) alias to `file-text` / `layout-grid`, but `outline/clipboard-list.svg` and `outline/grid.svg` both exist — so these names can never resolve to their actual Lucide icon. If intentional, a short comment would help; otherwise the alias can be dropped.

Related: `outline-*` coverage is uneven — aliases exist for `outline-cart` / `outline-case` etc., but not `outline-payments` / `outline-products` / `outline-security`, which then fall back to their solid glyph. A row mixing them (e.g. the home feature cards) ends up showing some outline and some solid icons.

client-app/shared/common/components/address-selection.vue
```diff
@@ -15,7 +15,7 @@
         v-if="!readonly"
         :disabled="disabled"
         variant="outline"
-        icon="edit"
+        icon="pencil"
```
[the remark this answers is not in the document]
goldenmaya: icon was changed from edit to pencil

ivan-kalachikov (dismissed):

ivan-kalachikov (dismissed):

client-app/shared/account/components/credit-card.vue
```diff
@@ -58,14 +58,16 @@ defineProps<IProps>();
...
+
+    @apply size-6;
 
     @media (width > theme("screens.xs")) {
       @apply size-8;
     }
 
     #{$disabled} & {
-      @apply fill-inherit;
+      @apply text-inherit;
```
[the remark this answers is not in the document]
goldenmaya: fixed

vas11yev1work (approved):

PR #2388, opened 2026-07-17
feat(VCST-5485): Sales Rep hub Dashboard

ivan-kalachikov:
## Description

Builds the **Sales Rep hub → Dashboard** (VCST-5485). Stacked on #2383.

- **Navigation + route** — adds *Dashboard* as the first item of the "Sales Rep hub" left-rail section (`view-grid` icon); `/company/dashboard`, gated to reps via the shared `guardSalesRep`.
- **KPI widget row** — shared presentational `StatWidget` card + `StatWidgets` grid wrapper; four cards per the design. Figures are **mock** (no sales-rep endpoint backs stats/deltas yet).
- **Orders** — a shared, reusable Orders widget used by both the dashboard (cross-customer, Customer column) and the customer profile (single org, Items column). Configurable `title` and `limit`; the status column shows the backend `statusDisplayValue`, localized via `cultureName`.
  - Extends `salesRepOrders` (optional `organizationId`, `cultureName`, `organizationName`, `statusDisplayValue`); types regenerated via codegen.
  - Generalizes `useSalesRepCustomerOrders` → `useSalesRepOrders`; replaces `customer-profile-orders` with the shared `orders` component.

Data policy: all dashboard data comes from the sales-rep BE (`salesRep*`); sections with no such endpoint (Top Sellers, Documents, Activity, etc.) are intentionally omitted. Titles localized across all 13 module locales.

> **Order status filter tabs** were split out into a follow-up PR (#2389, stacked on this one) so this PR delivers the Orders widget without the filter feature.

Verified: ESLint, `vue-tsc --build --force`, `check-locales`, and unit tests all pass.

> **Stacked PR** — base is `feat/VCST-5308_customer-profile-view` (#2383). Merge #2383 first; retarget once it lands.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5485
Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.54.0-pr-2388-3f18-3f18b01b.zip

PR #2392, opened 2026-07-20
feat(VCST-5310): [Sales Rep] push messages & emails to customers

vkrashenko:
## VCST-5310

A Sales Rep can broadcast a message to **all members of one customer organization** over **Email and/or Push**, from two entry points:

- **My customers** list — a per-row **Actions** column with an envelope icon.
- **Customer profile** (VCST-5308) — a **Quick actions** widget (this PR introduces the widget).

Member-level targeting is out of scope by design — the backend resolves recipients (all org members).

## What's included
- **Module's first mutation** — `sendCustomerCommunication` (`.graphql` + codegen into the module's own `api/graphql/types.ts`; core `types.ts` untouched).
- **`useSalesRepCommunication`** composable — merges `storeId`/`cultureName` from `globals`; returns a plain `boolean` so the UI can show success/failure.
- **`customer-communication-modal.vue`** — reuses `VcModal` + vee-validate/yup (like the Invite-members dialog): optional **Title**, required **Message** (max 1000, URLs preserved), **Email**/**Push** checkboxes with "at least one channel" validation.
- **Quick actions widget** (`customer-profile-actions.vue`) — 2×2 tile grid matching the storefront design. **Send email** is wired; **Order on behalf / Create quote / Schedule call** are disabled placeholders for their own tickets.
- **i18n** under `sales_rep.communication.*` + `my_customers.table.actions` — en + all 12 locales.

## Backend dependency
Requires the `sendCustomerCommunication` mutation on the `/graphql/sales-rep` schema, shipped in **vc-module-sales-rep #3**. Already live on the dev backend.
Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.54.0-pr-2392-d4b9-d4b9fa49.zip


client-app/modules/sales-rep/components/customer-communication-modal.vue
```diff
@@ -0,0 +1,147 @@
+<template>
+  <VcModal ref="modalComponent" :title="t('sales_rep.communication.title')" is-mobile-fullscreen dividers>
+    <p class="customer-communication-modal__subtitle">
+      {{ t("sales_rep.communication.subtitle", { organization: organizationName }) }}
+    </p>
+
+    <form class="space-y-4" @submit.prevent>
```
ivan-kalachikov: let's move it to <styles> block as everywhere else in the component we use BEM approach

client-app/modules/sales-rep/components/customer-communication-modal.vue
```diff
@@ -0,0 +1,147 @@
+<template>
+  <VcModal ref="modalComponent" :title="t('sales_rep.communication.title')" is-mobile-fullscreen dividers>
```
ivan-kalachikov: by BEM we need to add a root class customer-communication-modal to the parent element

client-app/modules/sales-rep/components/customer-profile-actions.vue
```diff
@@ -0,0 +1,133 @@
...
+  });
+}
+</script>
+
+<style lang="scss">
+// `@apply` keeps the module self-contained as an MF remote (no global utility layer). See PORT_TO_MF.md.
+.customer-profile-actions {
+  // Header divider (size=lg drops the built-in one).
+  .vc-widget__header-container {
+    @apply border-b border-neutral-200;
```
ivan-kalachikov: let's use vc-widget with size="md" to avoid necessity of adding border manually - moreover I've already changed all other widgets through the module to "md" size in my PRs

client-app/modules/sales-rep/pages/my-customers.vue
```diff
@@ -107,6 +117,23 @@
 
               <template v-else>—</template>
             </VcTableColumn>
+
+            <VcTableColumn
+              id="actions"
+              v-slot="{ item }"
+              :title="t('sales_rep.my_customers.table.actions')"
+              class="w-16 text-center"
```
ivan-kalachikov: let's continue using BEM approach here

client-app/modules/sales-rep/components/customer-communication-modal.vue
```diff
@@ -0,0 +1,147 @@
...
+const sendEmail = ref(false);
+const sendPush = ref(false);
+const channelSelected = computed(() => sendEmail.value || sendPush.value);
+
+const { errors, meta, handleSubmit } = useForm({
+  initialValues: { title: "", message: "" },
+});
+
+const { value: title } = useField<string>("title", toTypedSchema(string().max(128)));
+const { value: message } = useField<string>("message", toTypedSchema(string().required().max(1000)));
```
ivan-kalachikov: `string().required()` only rejects an empty string, so a message of just spaces passes validation. Submit then does `data.message.trim()` → `""`, i.e. an empty broadcast to every member of the org.

Suggest `string().trim().required().max(1000)` so validation and the sent payload use the same trimmed value.

client-app/modules/sales-rep/components/customer-communication-modal.vue
```diff
@@ -0,0 +1,147 @@
...
+      {{ t("sales_rep.communication.subtitle", { organization: organizationName }) }}
+    </p>
+
+    <form class="space-y-4" @submit.prevent>
+      <VcInput
+        v-model="title"
+        :label="t('sales_rep.communication.title_label')"
+        :placeholder="t('common.placeholders.enter_value')"
+        :disabled="loading"
+        :max-length="128"
```
ivan-kalachikov: `VcInput`'s prop is `maxlength` (lowercase), not `max-length` — so this cap is silently ignored (it just lands in `$attrs` as an invalid HTML attribute; `my-customers.vue` uses `maxlength`).

A long title then only trips the yup `max(128)` rule, but the Title field has no `:error`/`:message` binding, so Send goes disabled with nothing on screen explaining why. Use `:maxlength="128"` (and optionally bind the error like the textarea does).

client-app/modules/sales-rep/components/customer-profile-actions.vue
```diff
@@ -0,0 +1,133 @@
+<template>
+  <VcWidget :title="t('sales_rep.communication.quick_actions.title')" size="lg" class="customer-profile-actions">
```
ivan-kalachikov: This is the same UI-Kit-internals mutation that was rejected on #2388 — the `.vc-widget__header-container` / `.vc-widget__slot` overrides in the style block below. It was fixed there in 4ae236dcb by using `size="md"` (md keeps the header divider natively) and owning the inset on the grid's own wrapper. Worth doing the same here rather than reaching into VcWidget's internals.

client-app/modules/sales-rep/components/customer-communication-modal.vue
```diff
@@ -0,0 +1,147 @@
...
+  const succeeded = await sendCommunication({
+    organizationId: props.organizationId,
+    sendEmail: sendEmail.value,
+    sendPush: sendPush.value,
+    title: data.title.trim() || undefined,
+    message: data.message.trim(),
+  });
+
+  if (succeeded) {
+    emit("result", true);
```
ivan-kalachikov: Nit: `emit("result", ...)` is never consumed — neither call site (`my-customers.vue`, `customer-profile-actions.vue`) passes an `onResult` handler, so this is dead code. Either wire it (like invite-member-modal does) or drop the emit.

client-app/modules/sales-rep/composables/useSalesRepCommunication.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  const { ref } = await import("vue");
+  const mutate = vi.fn();
+  const loading = ref(false);
+  const useMutation = vi.fn(() => ({ mutate, loading }));
+  return { mutate, loading, useMutation };
+});
+
+vi.mock("@vue/apollo-composable", () => ({ useMutation: mutationMock.useMutation }));
+// The operation is imported as a raw document; useMutation is mocked, so a stub is enough.
+vi.mock("../api/graphql/mutations/sendCustomerCommunication/sendCustomerCommunicationMutation.graphql", () => ({
```
ivan-kalachikov: Nit: this mock is dead code — the composable imports `SendCustomerCommunicationDocument` from `../api/graphql/types`, not this `.graphql` file, so it never enters the test's module graph (and the comment is inaccurate). Safe to remove.

client-app/modules/sales-rep/components/customer-profile-actions.vue
```diff
@@ -0,0 +1,133 @@
...
+
+  &__grid {
+    // auto-rows-fr keeps both rows the same height even when a label wraps (e.g. "Order on behalf").
+    @apply grid grid-cols-2 gap-3 px-5 pb-5 pt-4;
+
+    grid-auto-rows: 1fr;
+  }
+
+  &__tile {
+    @apply flex h-24 flex-col items-center justify-center gap-2 rounded-lg border-2 px-2 text-center text-sm font-medium leading-tight transition-shadow;
```
ivan-kalachikov: Nit (styling, per #2388): `rounded-lg` → `rounded-[--vc-radius]` to respect radius customization, and the hover `box-shadow` a few lines down uses a raw value instead of a token. `font-medium` is Lato 500 which the font doesn't ship — minor, since existing code uses it too.

ivan-kalachikov (commented):
A few inline findings from a review pass — two real defects (whitespace message, title max-length), one repeat of feedback from #2388, plus small nits.

ivan-kalachikov (dismissed):

ivan-kalachikov (dismissed):

yuskithedeveloper (dismissed):

ivan-kalachikov (approved):

PR #2399, opened 2026-07-24
feat(VCST-5281): added organization invite and status for multi organization customers

DmitryGrishinVirtoworks:
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.55.0-pr-2399-f69e-f69e3377.zip


yuskithedeveloper (changes requested):
Reviewed `dev...feat/VCST-5281` (3 commits, 41 files) alongside the paired changes in `vc-module-customer`
and `vc-module-profile-experience-api`.

The pending-invites widget and the members-page invite actions are clean and read well. Main concerns are one
upgrade regression, an abandoned codegen path, and a query shape that amplifies an N+1 on the server.

---

### 🔴 Blocker

#### 1. `statuses: ["Approved"]` hides every organization for existing users **[cross-repo]**
```ts
// client-app/shared/account/composables/useUserOrganizations.ts:50
statuses: [ContactStatus.Approved],
```
The backing column `CustomerOrganizationMembership.Status` is added by this ticket as **nullable with no
backfill**, and the X-API resolver computes `effectiveStatus = membership.Status ?? contact.Status`, dropping
anything that isn't literally `"Approved"`.

On a sample DB: **1031 of 1048 contacts have `Status = NULL`**, plus 5 `New`. After upgrade those users get an
empty organization switcher, and `getDisplayStatus()` in `client-app/pages/company/members.vue:418` renders a
blank status cell.

Blocked on a backfill migration in vc-module-customer (raised there). If that lands, no change needed here; if
it doesn't, this filter needs to tolerate a missing status.

---

### 🟠 Performance

#### 2. `myStatusInOrganization` on a `first: 50` connection triggers ~150 backend queries
```graphql
# client-app/core/api/graphql/account/queries/getPendingOrganizationInvites/getPendingOrganizationInvites.graphql
organizations(after: $after, first: $first, statuses: $statuses) {
  items { id name myStatusInOrganization }
}
```
`OrganizationType.myStatusInOrganization` is an unbatched `ResolveAsync` that does
`FindByIdAsync` + `GetByIdAsync` + a membership search **per organization**. The fix belongs on the server
(raised in the profile-experience-api PR), but worth knowing this query is what surfaces it — the dashboard
widget pays it on every page load.

---

### 🟡 Findings

#### 3. Codegen abandoned for two operations
`client-app/core/api/graphql/account/queries/getOrganizations/index.ts` was *converted away* from
`GetOrganizationsDocument` to a raw `.graphql` import plus a hand-written `IGetOrganizationsQuery` interface,
and `getPendingOrganizationInvites/index.ts` follows the same pattern.

`types.ts` **was** regenerated (commit `7a625452a`), so the generated documents should exist — this looks like
a local workaround that got committed. It breaks the repo's typed-operation convention and the hand-rolled
shapes will silently drift from the schema.

#### 4. Redundant and fragile client-side status filter
`useUserPendingInvites` asks the server for `statuses: [Invited]` and then filters again:
```ts
items.value.filter((item) => item.myStatusInOrganization === ContactStatus.Invited)
```
The FE `ContactStatus` enum duplicates the backend `ModuleConstants.MembershipStatuses` string constants with
no contract between them — a rename on either side breaks this silently. Either drop the client filter, or add
a shared constant source.

#### 5. `PAGE_SIZE = 50` with no pagination
`client-app/shared/account/composables/useUserPendingInvites.ts` — invite someone to 51 organizations and the
51st is silently invisible. `hasNextPage` / `endCursor` are fetched and then discarded.

#### 6. `createGlobalState` + `onMounted` + an `initialized` flag that never resets
Pending invites survive a logout / login-as-another-user within the same SPA session.
(`useUserOrganizations` has the same shape, so it's a pre-existing pattern — but the invite list is more
sensitive, so it's worth resetting on auth change.)

#### 7. Drive-by CSS change on an unrelated menu item
`client-app/shared/company/components/members-dropdown-menu.vue` changed the existing **delete** item from
`class="text-danger"` to `class="fill-danger"`. Different CSS property, unrelated to this ticket — please
confirm it's intentional.

client-app/pages/company/members.vue
```diff
@@ -586,6 +602,50 @@ function openDeleteModal(contact: ExtendedContactType): void {
...
+      text: t("shared.company.revoke_invite_modal.text", {
+        name: contact.extended.emails[0] ?? contact.fullName,
+      }),
+      async onConfirm() {
+        try {
+          await revokeInvite(contact.id);
+        } catch {
+          notifications.error({ duration: 5000, single: true, text: t("common.messages.invite_revoke_failed") });
+        }
+        closeRevokeInviteModal();
```
vas11yev1work, line 641: The modal closes even if the revoke was unsuccessful. Is this correct behavior?

client-app/shared/company/composables/useOrganizationContacts.ts
```diff
@@ -127,6 +132,41 @@ export function useOrganizationContacts(organizationId: MaybeRef<string>) {
...
+    }
+
+    page.value = 1;
+    await fetchContacts();
+
+    notifications.success({
+      text: t("pages.company.members.notifications.invite_revoked"),
+      duration: 10000,
+      single: true,
+    });
```
vas11yev1work: `revokeInvite` and `resendInvite` have inconsistent notification ownership: `revokeInvite` shows the success toast itself but leaves error toasts to the caller, while `resendInvite` shows neither — both success and failure toasts are pushed entirely onto the caller. Worth unifying so a future caller can't accidentally drop a toast

client-app/shared/account/composables/useUserOrganizations.ts
```diff
@@ -46,6 +47,7 @@ function _useUserOrganizations() {
         first: ORGANIZATIONS_PER_PAGE,
         sort: "name:asc",
         searchPhrase: formattedSearchPhrase.value,
+        statuses: [ContactStatus.Approved],
```
ivan-kalachikov, line 50: **This drops every organization for users whose global contact status is not literally `Approved`.**

The `statuses` argument is resolved by `ContactType.ResolveMyOrganizationIdsByStatusAsync`, which filters in memory on `OrganizationMembership.ResolveEffectiveStatus(overrideStatus, globalStatus)`. With no per-membership override (all pre-upgrade rows), the effective status **is the contact's global `Contact.Status`**. A null/empty global status falls through to `"Approved"` and is fine — but a non-empty, non-`Approved` one fails `lifecycleStatuses.Contains(effectiveStatus)` and the organization is dropped.

That population is not hypothetical. In vc-module-profile-experience-api:

- `RegisterRequestCommandHandler.ToContact`: `result.Status = requireEmailVerification ? ModuleConstants.ContactStatuses.Locked : DefaultContactStatus;`
- `ConfirmEmailCommandHandler` only calls `userManager.SetLockoutEndDateAsync(user, null)` — **nothing ever resets `Contact.Status`**.
- `ModuleConstants.ContactStatuses` = `{ Locked, Invited, Approved }`.

So **every self-registered user on a store with email verification required carries global status `"Locked"` permanently**, logs in normally, and after this change loses every organization from the switcher list. The same happens on any store where `Customer.ContactDefaultStatus` is set to something other than `Approved` (the setting has no default value, so unset is null and therefore safe).

It also fails visibly rather than silently: the switcher's visibility gate is `isMultiOrganization` (`useUser.ts:382`), computed from the **unfiltered** `me.contact.organizations.totalCount`, so these users get a rendered switcher showing its "no results" empty state.

Options: drop the filter here and rely on the effective status the resolver already returns; or include the other lifecycle values the user should still see; or have the backend coalesce a non-`Approved` *global* status differently from a genuine per-membership override. Worth deciding with the backend PR rather than patching one side.

client-app/pages/company/members.vue
```diff
@@ -408,6 +415,10 @@ function canShowDropdownFor(contact: ExtendedContactType): boolean {
   return contact.id !== user.value.memberId && (userCanEditOrganization.value || canLoginOnBehalfOf(contact));
 }
 
+function getDisplayStatus(contact: ExtendedContactType): string | undefined {
+  return contact.isLockedInOrganization ? ContactStatus.Locked : (contact.statusInOrganization ?? contact.status);
```
ivan-kalachikov, line 423: Same root cause as the `useUserOrganizations.ts` comment, surfacing on the members grid.

For a contact with global `Contact.Status = "Locked"` and no per-membership override, `statusInOrganization` resolves to `"Locked"`, so this returns `ContactStatus.Locked` and the row renders the red **Blocked** badge. But the **Blocked** facet is matched server-side against `IsCurrentlyLocked` membership rows only (`ResolveMyOrganizationIdsByStatusAsync` / `ContactMatchesStatusFilter` strip `"Locked"` out of `lifecycleStatuses` and test the lock flag instead), and these contacts have no locked membership row — so the facet will not return them. "Active" won't either, since their effective status is not `Approved`.

Net: rows badged Blocked that **no facet can find**. Before this PR the badge read raw `contact.status` and the facet filtered on the same global value, so the two agreed; the divergence is new here.

To be explicit, since it was in the previous review: the lock channel itself is fine — `ResolveIsLockedInOrganization` queries with `OnlyLocked`, which `OrganizationMembershipSearchService.BuildQuery` implements as `x.IsLocked && (x.LockoutEnd == null || x.LockoutEnd > now)`, identical to `IsCurrentlyLocked`. For genuine membership locks the badge and the facet agree exactly. Only the global-status-derived case diverges, so fixing the comment above resolves this too.

client-app/pages/company/members.vue
```diff
@@ -471,8 +482,12 @@ async function resetKeyword() {
 function syncFilterFromFacets() {
   const roleFacet = appliedFacets.value.find((f) => f.paramName === "roleId");
   roleIds.value = roleFacet?.values.filter((v) => v.selected).map((v) => v.value) ?? [];
-  const nonRoleFacets = appliedFacets.value.filter((f) => f.paramName !== "roleId");
-  filter.value = getFilterExpressionFromFacets(nonRoleFacets);
+
+  const statusFacet = appliedFacets.value.find((f) => f.paramName === "status");
+  statuses.value = statusFacet?.values.filter((v) => v.selected).map((v) => v.value) ?? [];
```
ivan-kalachikov, line 491: Diverting the status facet out of `filter` into `statuses` leaves the empty-state logic behind, because it still only tests `keyword`, `filter` and `roleIds`:

- L206 — `keyword || filter || roleIds.length ? no_results_message : no_members_message`
- L211 — `:variant="!!keyword || !!filter || !!roleIds.length ? 'search' : 'empty'"`
- L214 — the reset button's `v-if`

Repro: on an organization with no invited members, select only the **Invited** status filter. `contacts` comes back empty, and `keyword`/`filter`/`roleIds` are all falsy, so the page renders *"You don't have members yet"* with the continue-shopping button instead of *"no results"* with **Reset search** — while the applied filter chip is still visible directly above it.

This worked before the change: at the merge base the status facet was included in `getFilterExpressionFromFacets(nonRoleFacets)`, which returns a non-empty `"status":"Invited"`, so `filter` was truthy. The applied-facets chip row mitigates it on desktop only (`hidden ... lg:flex`); on mobile there's just the filter-button badge.

`resetFilters()` already clears `statuses`, so the reset path itself is correct — the button just isn't rendered. Adding `statuses.length` to the three conditions above is the whole fix.

client-app/shared/account/components/pending-invites-widget.vue
```diff
@@ -0,0 +1,91 @@
...
+          class="flex flex-col items-start gap-3 py-3 first:pt-0 last:pb-0 sm:flex-row sm:items-center sm:justify-between"
+        >
+          <span class="font-bold">{{ invite.name }}</span>
+
+          <div class="flex gap-2">
+            <VcButton
+              size="sm"
+              variant="outline"
+              color="secondary"
+              :disabled="loading"
```
ivan-kalachikov: Gating only on `loading` leaves a window where a second invite can be started, and `processingOrgId` cannot represent two actions.

`acceptInvite` releases the shared `loading` before it has finished: the mutation's `finally` sets it false (`useUserPendingInvites.ts:58`), `fetchPendingInvites()` sets it true then false again, and then lines 65-67 run `resetOrganizations(); await search()` with `loading` **false for the whole duration of that `getOrganizations` request**. Vue re-renders, so every remaining row is clickable again. (`rejectInvite` has no such tail, so it isn't affected.)

Two consequences, with two or more pending invites:

1. Accepting a second invite in that window sets `processingOrgId` to the new org; the *first* handler's `finally` (L66) then nulls it while the second mutation is still in flight, so that row shows no spinner at all. Last-write-wins on set, first-completion-wins on clear.
2. Worse, the org list can end up duplicated. `useUserOrganizations.reset()` sets `loading.value = false` (L88), which defeats the `if (loading.value || !hasNextPage.value) return` guard at L38. Interleaving: accept A -> `search()` -> `loadOrganizations` takes the guard, request R1 in flight; accept B during R1 -> B's `reset()` forces `loading` false -> B's `search()` clears `organizations` and issues R2 -> R1 and R2 both `organizations.value.push(...)` (L53), so every organization appears twice in the switcher, with a corrupted `endCursor`.

Smallest fix is to make `processingOrgId` authoritative for the gate too, since it already spans the whole handler including `search()`:

```suggestion
              :disabled="!!processingOrgId || loading"
```

(and the same on the Accept button below). Alternatively hold `loading` across the entire `acceptInvite` including the switcher refresh, or stop `reset()` from touching `loading`.

ivan-kalachikov (commented):
### Code review

Reviewed `dev...feat/VCST-5281` together with the paired backend PRs (vc-module-profile-experience-api#141, vc-module-customer#312), since most of the risk here is in how the frontend reads the new status contract.

The invite mutations, the pending-invites widget and the members-page actions are wired correctly, and the codegen/pagination/input-type items from the previous review are genuinely fixed. Eight comments inline: one I'd treat as blocking, three worth fixing, the rest minor.

**One correction to the previous review worth flagging:** finding #1 there attributed the problem to the nullable `Status` column having no backfill (`1031/1048 contacts have Status = NULL`). Those contacts are actually **safe** — `OrganizationMembership.ResolveEffectiveStatus` coalesces `null -> contact.Status -> "Approved"`, and the `statuses` argument is filtered in memory against that effective value, not against the raw column. So a `NULL -> Approved` backfill in #312 would **not** fix the real issue. The affected population is instead contacts whose global `Contact.Status` is non-empty and not `"Approved"` — see the comment on `useUserOrganizations.ts`.

Also verified and clean, to save reviewer time: `"Locked"` is a legitimate first-class value of the new `statuses` argument (`LockedFilterValue`, special-cased in both `ContactType` and `OrganizationType`), so the Blocked facet value itself is correct; `organizationId: ""` from `confirm-invitation.vue` is harmless because `ApproveInvitedMembershipsAsync` treats empty as absent; `resendOrganizationInvite`/`revokeOrganizationInvite` taking only `memberId` is unambiguous because `ProfileSchema` injects `command.OrganizationId = context.GetCurrentOrganizationId()`; the `organizationId` vs `memberId` asymmetry between accept/reject and revoke/resend is intentional and correctly wired on both sides; rejecting removes the org from `Contact.Organizations`, so counts are not left inflated; the `while (hasNextPage)` loop terminates because these connection cursors are offset-based; and `convertToExtendedContact` keying the "Invite sent" name fallback off global `contact.status` is correct on purpose (it substitutes for a missing *name*, not a missing per-org status).

client-app/shared/account/composables/useUserPendingInvites.ts
```diff
@@ -0,0 +1,101 @@
...
+      loading.value = false;
+    }
+
+    await fetchPendingInvites();
+
+    // useUserOrganizations only fetches once on mount, so the switcher needs an explicit refresh
+    // to show the newly-accepted organization without a full page reload.
+    const { reset: resetOrganizations, search } = useUserOrganizations();
+    resetOrganizations();
+    await search();
```
ivan-kalachikov: This refreshes the organization **list**, but not the flag that decides whether the switcher is rendered at all — so the comment above ("the switcher needs an explicit refresh ... without a full page reload") doesn't hold in the case that matters most.

Both switchers are gated by `isMultiOrganization` (`top-header.vue:169`, `mobile-menu.vue:145`), which is `(user.value?.contact?.organizations?.totalCount ?? 0) > 1` (`useUser.ts:382`) from the `me` query. Neither this composable nor the widget ever calls `fetchUser()`.

Reachable symptom: a user who is in exactly one organization and is invited **mid-session** (after their `getMe` ran) opens the account page, sees the widget, accepts, gets the success toast and a refreshed list — but `totalCount` is still the stale `1`, so the switcher stays hidden until a full reload or re-login.

If the invite already existed when the session started there's no hiding bug, since invited organizations are included in `Contact.Organizations`; instead the gate is inflated, so a one-organization user with a pending invite sees a switcher listing a single organization. Rejecting does remove the organization server-side, but the stale count means the switcher doesn't disappear either.

Calling `fetchUser()` after accept (and reject) fixes all three.

client-app/shared/account/components/pending-invites-widget.vue
```diff
@@ -0,0 +1,91 @@
...
+
+  try {
+    await acceptInvite(organizationId);
+
+    notifications.success({
+      text: t("shared.account.pending_invites_widget.invite_accepted"),
+      duration: 10000,
+      single: true,
+    });
+  } catch {
```
ivan-kalachikov, line 66: Minor: this `catch` also covers the post-mutation refresh, so a successful accept can be reported as a failure.

`acceptInvite` awaits `fetchPendingInvites()` and `useUserOrganizations.search()` **after** its own `try/finally`, and both rethrow (`useUserPendingInvites.ts:43`, `useUserOrganizations.ts:66`). If either fails, the invite has already been accepted server-side but the user sees *"Something went wrong. Please try again"* and no success toast — then plausibly retries an invite that no longer exists, and the stale row is still there precisely because the refetch that removes it is what failed.

Low probability, and showing *something* is defensible since the UI is in an unknown state, but the message is misleading. Separating the mutation's error handling from the refresh's would let you keep the success toast and surface a "couldn't refresh" state instead.

client-app/core/api/graphql/account/queries/getOrganizations/index.ts
```diff
@@ -1,9 +1,11 @@
 import { GetOrganizationsDocument } from "@/core/api/graphql/types";
 import { graphqlClient } from "../../../client";
-import type { GetOrganizationsQueryVariables, OrganizationFieldsFragment } from "@/core/api/graphql/types";
+import type { GetOrganizationsQueryVariables, Organization } from "@/core/api/graphql/types";
+
+export type OrganizationFieldsType = Pick<Organization, "id" | "name">;
```
ivan-kalachikov: Nit: the generated fragment type is still available and still the one the query actually selects.

`getOrganizations.graphql` spreads `...organizationFields`, and codegen emits `OrganizationFieldsFragment = { id: string, name?: string }`, which this file used before. Hand-rolling the `Pick` leaves that generated type orphaned, and it will silently stay narrow if anyone widens `fragments/organizationFields.graphql` later. It also types `name` as `Maybe<string>` (so `null` is included) where the fragment type has `string | undefined`.

No behavioural difference today — the fragment is exactly `id` + `name` and every consumer reads only those — so purely a typing-consistency point. Same applies to `PendingOrganizationInviteType` in `getPendingOrganizationInvites/index.ts`, though there's no fragment there, so `GetPendingOrganizationInvitesQuery`'s item type would be the equivalent.

client-app/shared/account/components/pending-invites-widget.vue
```diff
@@ -0,0 +1,91 @@
+<template>
+  <VcWidget v-if="pendingInvites.length" :title="$t('shared.account.pending_invites_widget.title')">
+    <template #default-container>
+      <ul class="divide-y px-[--p-x] pb-5 pt-4">
+        <li
+          v-for="invite in pendingInvites"
+          :key="invite.id"
+          class="flex flex-col items-start gap-3 py-3 first:pt-0 last:pb-0 sm:flex-row sm:items-center sm:justify-between"
+        >
+          <span class="font-bold">{{ invite.name }}</span>
```
ivan-kalachikov, line 10: Nit (a11y): the organization name isn't associated with the row's buttons, so with several pending invites a screen reader announces identical "Accept" / "Decline" pairs with nothing distinguishing them.

Not a WCAG failure — no criterion requires unique accessible names, and the name is adjacent visible context — but an easy improvement, e.g. give this span an `:id` and point the buttons at it with `aria-describedby`, or include the name in each button's `aria-label`.

client-app/pages/company/members.vue
DmitryGrishinVirtoworks: Fixed

client-app/shared/company/composables/useOrganizationContacts.ts
DmitryGrishinVirtoworks: Fixed

client-app/shared/account/composables/useUserOrganizations.ts
DmitryGrishinVirtoworks: Done

client-app/pages/company/members.vue
DmitryGrishinVirtoworks: Done

client-app/pages/company/members.vue
DmitryGrishinVirtoworks: Done

client-app/shared/account/components/pending-invites-widget.vue
DmitryGrishinVirtoworks: Done

client-app/shared/account/composables/useUserPendingInvites.ts
DmitryGrishinVirtoworks: Done

client-app/shared/account/components/pending-invites-widget.vue
DmitryGrishinVirtoworks: Done

client-app/core/api/graphql/account/queries/getOrganizations/index.ts
DmitryGrishinVirtoworks: Done

client-app/shared/account/components/pending-invites-widget.vue
DmitryGrishinVirtoworks: Done

vas11yev1work (approved):

ivan-kalachikov (approved):

yuskithedeveloper (approved):

PR #2400, opened 2026-07-27
feat(VCST-5367): [Sales Rep] drag-and-drop saved layout for hub and customer profile

ivan-kalachikov:
## Description

A Sales Rep can reorder and hide the blocks on their **Dashboard** and **Customer profile**, and the arrangement survives reload, a new session and another device. One layout per rep per surface — the customer-profile layout applies to every customer.

Backed by `salesRepLayout` / `saveSalesRepLayout` on `/graphql/sales-rep` (live on `vcptcore-dev`). **v1 persists order and visibility only**; per-block `settings` goes out empty.

Design spec: `client-app/modules/sales-rep/specs/VCST-5367-srh-layout-drag-and-drop/2026-07-27-srh-saved-layout-design.md` (kept inside the module, which is meant to travel as an MF remote)

### How it works

- **`layout/document.ts`** — pure reconcile + serialize. A block&#39;s **region always comes from the registry, never from the saved document**, so a widget that moves rail-to-main in a later release takes existing layouts with it instead of pinning a stale position. Unknown block types are dropped; newly shipped blocks are appended without disturbing an arrangement the rep already chose.
- **`layout/registry.ts`** — the block registry for both surfaces, plus `registerBlock()` so widgets from later stories participate without touching the engine. Its `statistics` blocks are derived from `layout/stat-cards.ts`, the single table that also feeds the KPI cards — so a stat id or caption exists in exactly one place, and `buildStatCards` makes a card shipped without its data a compile error rather than a blank tile.
- **`layout-surface.vue`** — one component for everything both pages had in common, which was every line from the load alert to the `aria-live` announcer. The pages keep only their own identity (title / customer hero), the customer they name, and where the edit button sits. It renders the blocks itself rather than exposing a slot per region: which component a block is comes from the registry, so a slot would hand the page a job it could only do by re-deriving that.
- **`layout-block.vue` + `layout-widget.vue`** — the block renders no chrome of its own. It `provide`s the drag controls and `layout-widget.vue` — the wrapper every sales-rep widget renders instead of `VcWidget` — injects them into the kit&#39;s own `#prepend` / `#append` slots, composed with whatever header content the widget already had. **So the controls are placed by the widget&#39;s real padding and header height, and no rule outside the kit touches a `.vc-widget__*` class.** A `LayoutWidget` with no block above it is a plain widget; injection is consumed once, so a nested widget renders no second handle.
- **`useSalesRepLayout`** — explicit edit mode over a draft; Save writes the whole document in one mutation (the backend replaces rather than merges). A failed save keeps the draft and stays in edit mode; a failed **read** disables editing, because a full-replace save would otherwise overwrite a layout we could not fetch.
- **`useKeyboardSort` + `aria-live`** — space grabs, arrows move, space drops, escape restores; blur cancels. The stat row is horizontal (←/→ reorder, ↑/↓ park and restore); widget columns are vertical.
- **Drag** — `sortablejs` used directly. `mainLeft` and `mainRight` get distinct Sortable groups, so a rail widget can never be dropped into the wide column — enforced by construction rather than by a runtime guard. A widget drags by its whole header, with a `filter` so a mousedown on the hide button inside that header does not start a drag instead.

### Behaviour worth reviewing

- Nothing block-shaped renders until the layout query resolves, so the rep never sees the default arrangement shuffle under them. Page title, breadcrumbs and the customer hero still render immediately — they don&#39;t reorder, so gating them would add a flash rather than remove one.
- **Hidden blocks are never rendered, so they never fetch.** Hiding Recent Orders means no `salesRepOrders` request. Pinned by a test.
- The right rail is mounted only while something is visible in it — an empty `aside` would otherwise hold its desktop width open. Hidden widgets come back through the tray, which is pinned by a test.
- Stat cards drag by the whole card, carry an accent-coloured handle and have no ✕ — they are parked in the unused zone instead. Widgets drag by their whole header, handle at the start, hide at the end.
- The drag handle is a real `<button>` because it is the only keyboard route into reordering (space/enter grab, arrows move). A plain mouse click on it does nothing by design — the pointer gesture is the drag.
- Dark mode needs no override: every token used inverts correctly in the presets (verified by reading `assets/presets/default{,.dark}.json`, not by inspecting pixels — worth one visual pass in QA).

### Refactor included

`dashboard-widgets.vue`, `customer-profile-widgets.vue` and `stat-widgets.vue` are removed. The KPI cards are now data looked up by id and the responsive grid lives in `layout-region.vue`. `stat-widget.vue` itself is unchanged apart from a new `leading` slot. Only the two pages referenced the deleted components.

`layout-skeleton.vue` draws its widget boxes with the kit&#39;s `VcWidgetSkeleton` and its KPI boxes with the real `.stat-widget` classes, so it copies none of VcWidget&#39;s metrics — the kit&#39;s skeleton comes out at the same 3.125rem header height at `size="md"`.

`api/graphql/types.ts` was regenerated against `vcptcore-dev`; unrelated core/quotes drift from that run was pruned back out.

### Verification

`eslint` exit 0 · `vue-tsc --build --force` exit 0 · `vitest client-app/modules/sales-rep/` 189 tests passing · `yarn build` exit 0.

Manual browser passes against `vcptcore-dev` caught several defects already fixed in this branch (cross-zone drops disabled, duplicated cards, drag preview not reflowing). **Further manual QA is still needed** — automated checks cannot judge drag feel or the screen-reader announcements. Two things to look at specifically after the review round: clicking ✕ on a widget header (the drag `filter`), and the widget-header grab, whose `cursor: grab` hint moved to the handle when the kit-internal style rule was removed.

https://github.com/user-attachments/assets/bcd981be-4c5c-45b5-84ca-4bff0a57d77c

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5367
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.55.0-pr-2400-d55f-d55f339c.zip


client-app/modules/sales-rep/pages/dashboard.vue
```diff
@@ -4,21 +4,138 @@
...
 
-    <SalesRepOrders :title="t('sales_rep.orders.title')" filterable />
+    <!-- Nothing block-shaped renders until the saved layout is known; see layout-skeleton.vue. -->
+    <LayoutSkeleton v-if="layoutLoading" :scope="SCOPE" />
 
-    <TopSellers :title="t('sales_rep.top_sellers.title')" />
+    <!-- Nothing may change while a save is in flight, keyboard included, so inert rather than the
+         overlay alone. Coerced to undefined because inert is presence-based: Vue renders a false value
+         as the attribute itself, which would inert the layout permanently. -->
+    <div v-else class="sales-rep-dashboard__layout" :inert="saving || undefined">
```
vkrashenko: This whole chrome block is repeated verbatim in `customer-profile.vue`: alert → skeleton → inert wrapper + loader overlay → `LayoutEditBar` → `LayoutStats` (identical down to all nine bindings) → column row → hidden tray → edit button → announcer, plus the same 19 destructured keys and the same SCSS.

Could this be a `layout-surface.vue` with slots only for what actually differs (`#main-block`, `#aside-block`, edit-button placement)? Minimum version: keep the composable as one object instead of re-listing 19 keys in both pages, and move the `aria-live` paragraph into a small component — right now it's copy-pasted with two different BEM names.

client-app/modules/sales-rep/components/layout-region.vue
```diff
@@ -0,0 +1,233 @@
...
+ *
+ * Remove before reading the child index, or a backwards move is off by one. `oldIndex` addresses
+ * `children`; the handlers below address `props.entries`.
+ */
+function restore(event: Sortable.SortableEvent): void {
+  event.item.remove();
+  event.from.insertBefore(event.item, event.from.children[event.oldIndex ?? 0] ?? null);
+}
+
+const titleOf = (id: string) => {
```
vkrashenko: Same three lines live in `layout-hidden-tray.vue:48` and `useLayoutAnnouncer.ts:28`. A `useBlockTitle(scope)` returning `titleOf` would cover all three call sites.

client-app/modules/sales-rep/layout/registry.ts
```diff
@@ -0,0 +1,111 @@
...
+const TopSellers = markRaw(defineAsyncComponent(() => import("../components/top-sellers.vue")));
+const CustomerProfileActions = markRaw(
+  defineAsyncComponent(() => import("../components/customer-profile-actions.vue")),
+);
+const CustomerProfileInfo = markRaw(defineAsyncComponent(() => import("../components/customer-profile-info.vue")));
+
+// Statistics-block ids are `StatWidgetCardType.key` values from useSalesRepDashboardWidgets /
+// useSalesRepCustomerWidgets — the stat row renders them by key rather than by component, so they
+// carry no `component` of their own. A key that drifts from those composables renders nothing.
+const dashboardBlocks: SalesRepBlockType[] = [
```
vkrashenko: These 11 statistics entries duplicate `key` + `labelKey` from `useSalesRepDashboardWidgets` / `useSalesRepCustomerWidgets` — the `titleKey` values are byte-identical to the composables' `labelKey`. `stat-card-keys.test.ts` pins the ids but not the label keys, so renaming a locale key in the composable leaves one caption on the card and another in the tray and the announcements.

A shared `STAT_CARD_DEFS` table per surface, which the composable turns into cards and the registry into blocks, would remove the second copy. (The description mentions `layout/stat-cards.ts` — it isn't in the branch, looks like this was the original plan.) Cheap alternative: assert `titleKey === labelKey` in the existing test.

client-app/modules/sales-rep/components/layout-skeleton.vue
```diff
@@ -0,0 +1,149 @@
...
+
+  &__block {
+    @apply flex flex-col rounded-[--vc-radius] border border-neutral-200 bg-additional-50 shadow-sm;
+  }
+
+  // VcWidget's `--header-min-h` and `--p-x` at size md, which every registered widget renders at.
+  &__head {
+    @apply flex items-center px-4 sm:px-6;
+
+    height: 3.125rem;
```
vkrashenko: This hand-builds a widget box (header height, `px-4 sm:px-6`, border/shadow), which makes it the third place mirroring `VcWidget`'s internal metrics. ui-kit already ships `VcWidgetSkeleton` (`head`, `size`, slots), used in `plugins/builder-preview/components/preview-page.vue` — `<VcWidgetSkeleton head size="md">` with N empty divs per block would drop most of this SCSS and one source of drift. The KPI card has no kit equivalent, so keep it local, or just render `<StatWidget loading>`. One caveat: `VcWidgetSkeleton` at `md` uses `--p-x: padding.6` where `VcWidget` uses `px-4 sm:px-6`, so expect a small visual delta.

vkrashenko (commented):
Solid work overall — the pure reconcile layer, the echo validation and the keyboard story are all in good shape. Four things about reuse/duplication, left inline.

client-app/modules/sales-rep/components/layout-block.vue
```diff
@@ -0,0 +1,200 @@
...
+
+    // The bar itself must not swallow drags aimed at the header underneath it.
+    > * {
+      @apply pointer-events-auto;
+    }
+  }
+
+  // The whole header is the drag surface, so the title has to step aside for the two controls.
+  &--editing:not(#{&}--whole) {
+    .vc-widget__header-container {
```
goldenmaya: Such component mutations should not exist out of the box. Such styles complicate client-side project customization. Such mutations are only allowed within the UI kit.

client-app/modules/sales-rep/components/layout-block.vue
```diff
@@ -0,0 +1,200 @@
...
+
+  // Outline, not border, so the box model is untouched. Rounded because an outline only follows
+  // corners the element itself has.
+  &--editing {
+    @apply rounded-[--vc-radius] outline-dashed outline-1 outline-offset-2 outline-neutral-300 transition-opacity;
+  }
+
+  // Stat cards: the card is the handle, so the grab cursor and the selection guard go on the block
+  // itself rather than on a chrome button.
+  &--whole#{&}--editing {
```
goldenmaya: Use a Sass variable instead of duplicating the class.

client-app/modules/sales-rep/components/layout-block.vue
```diff
@@ -0,0 +1,200 @@
...
+      }
+    }
+
+    .vc-widget__header {
+      padding-inline: var(--layout-block-header-inset);
+    }
+  }
+
+  &__handle,
+  &__hide {
```
goldenmaya: These are different elements, and there's no point in duplicating them just to save five lines of code. Such constructs make the code difficult to read.

client-app/modules/sales-rep/components/layout-block.vue
```diff
@@ -0,0 +1,200 @@
...
+
+    &:hover {
+      @apply bg-danger-50 text-danger-500;
+    }
+  }
+}
+
+// `sortable-ghost` lands on the dragged element, which Sortable moves to the insertion point — so it
+// previews what will land there. Faded, not an empty slot, matching `--grabbed`.
+.layout-block.sortable-ghost {
```
goldenmaya: &.sortable-ghost

client-app/modules/sales-rep/components/layout-block.vue
```diff
@@ -0,0 +1,200 @@
...
+  @apply rounded-[--vc-radius] opacity-45 outline-dashed outline-1 outline-offset-2 outline-primary-500;
+}
+
+// The clone under the pointer stays solid, so the card being carried reads as the real one.
+.layout-block.sortable-drag {
```
goldenmaya: &.sortable-drag

client-app/modules/sales-rep/components/layout-edit-bar.vue
```diff
@@ -0,0 +1,78 @@
...
+      <b>{{ t("sales_rep.hub.layout.editing") }}</b>
+      {{ t("sales_rep.hub.layout.hint") }}
+      <!-- Visible, not just announced: the arrow-key gesture was otherwise discoverable only by
+           screen-reader users, who hear it on grab. -->
+      <span class="layout-edit-bar__keys">{{ t("sales_rep.hub.layout.hint_keyboard") }}</span>
+    </p>
+
+    <div class="layout-edit-bar__actions">
+      <!-- Disabled, not just inert: `inert` on the layout stops the interaction but greys nothing out. -->
+      <VcButton size="sm" variant="no-background" color="secondary" :disabled="saving" @click="$emit('reset')">
```
goldenmaya: variant="no-background" is deprecated, please use another one.

client-app/modules/sales-rep/components/layout-edit-bar.vue
```diff
@@ -0,0 +1,78 @@
...
+        {{ t("sales_rep.hub.layout.cancel") }}
+      </VcButton>
+
+      <VcButton size="sm" prepend-icon="save" data-layout-save :loading="saving" @click="$emit('save')">
+        {{ t("sales_rep.hub.layout.save") }}
+      </VcButton>
+    </div>
+
+    <!-- A failed save keeps the draft, so the message sits beside the still-live Save button. -->
+    <VcAlert v-if="failed" class="layout-edit-bar__error" color="danger" size="sm" variant="solid-light" icon>
```
goldenmaya: variant="solid-light" is deprecated, please use another one.

client-app/modules/sales-rep/components/layout-edit-button.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+  <!-- Wrapped so the button keeps its intrinsic width inside a stretching flex column. -->
+  <div>
```
goldenmaya: Please add class="layout-edit-button" to simplify client-side customization.

client-app/modules/sales-rep/components/layout-hidden-tray.vue
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  &__list {
+    @apply m-0 flex list-none flex-wrap gap-2 p-0;
+  }
+
+  &__restore {
+    @apply flex items-center gap-2 rounded-[--vc-radius] border border-neutral-200 bg-additional-50 px-2.5 py-1.5 text-sm font-semibold text-neutral-900 transition-colors;
+
+    .vc-icon {
```
goldenmaya: Use
--vc-icon-color: var(--color-primary-500); 
OR
prop color="primary"
OR
class &__icon {.@apply text-primary-500; }

client-app/modules/sales-rep/components/layout-region.vue
```diff
@@ -0,0 +1,233 @@
...
+    @apply flex flex-wrap gap-4;
+
+    > * {
+      @apply min-w-0 grow basis-44;
+    }
+  }
+
+  // Extra room for the chrome, which would otherwise overlap a widget's own header controls. Vertical
+  // only: stat cards drag whole and carry no chrome, and here it would fight the zone padding above.
+  &--vertical#{&}--editing {
```
goldenmaya: Use sass variable instead

client-app/modules/sales-rep/components/layout-region.vue
```diff
@@ -0,0 +1,233 @@
...
+    box-shadow: 0 0 0 var(--layout-zone-band) var(--color-neutral-50);
+  }
+
+  &__empty {
+    @apply m-0 flex min-h-24 basis-full items-center justify-center text-center text-sm text-neutral-400;
+  }
+
+  // Emptiness is read from the DOM, not state, so the hint tracks the drag rather than the drop: it
+  // appears the moment the last card is dragged out, and hides the moment one is dragged in.
+  &:has(.layout-block) &__empty {
```
goldenmaya: Why can't you use if/else constructs in your template instead of this?

client-app/modules/sales-rep/components/layout-skeleton.vue
goldenmaya, line 1: This component is too complex for a skeleton, I think it needs to be simplified.

client-app/modules/sales-rep/components/layout-stats.vue
```diff
@@ -0,0 +1,170 @@
+<template>
+  <div class="layout-stats">
+    <p v-if="editing" :id="`${scope}-visible-stats`" class="layout-stats__label">
```
goldenmaya: Prefer use div instead of p. `<p>` in HTML is the text paragraph element.

client-app/modules/sales-rep/components/layout-stats.vue
```diff
@@ -0,0 +1,170 @@
...
+<style lang="scss">
+.layout-stats {
+  // Zone frames are painted 12px outside their box, so the gap has to clear that band before it reads
+  // as space: 20px here plus the label's 4px margin leaves the same 12px the padded frame used to.
+  @apply flex flex-col gap-5;
+
+  &__label {
+    @apply m-0 mt-1 flex items-center gap-1.5 text-xs font-bold uppercase tracking-wider text-neutral-500;
+
+    &--hidden {
```
goldenmaya: hidden is an unclear name for a modifier that doesn't actually hide the element

client-app/modules/sales-rep/components/layout-stats.vue
```diff
@@ -0,0 +1,170 @@
...
+  // Decorative only — the card itself is the control. Inside `.stat-widget`, so it inherits
+  // `--stat-widget-accent` and matches the card's edge bar.
+  &__handle {
+    @apply -me-0.5 inline-flex;
+
+    color: var(--stat-widget-accent);
+  }
+
+  // Diagonal hatching marks the parked zone as inert without hiding what is in it.
+  &__hidden {
```
goldenmaya: hidden is an unclear name here as well

client-app/modules/sales-rep/components/layout-stats.vue
```diff
@@ -0,0 +1,170 @@
...
+    background: repeating-linear-gradient(
+      45deg,
+      var(--color-neutral-50),
+      var(--color-neutral-50) 9px,
+      var(--color-neutral-100) 9px,
+      var(--color-neutral-100) 18px
+    );
+
+    .stat-widget {
+      @apply bg-additional-50;
```
goldenmaya: The background color of this component is already white, which is a strange duplication of styles. It's as if something went wrong. Let's figure it out.

client-app/modules/sales-rep/pages/dashboard.vue
ivan-kalachikov: Extracted to `layout-surface.vue` — it owns the whole run from the load alert to the `aria-live` announcer, and renders the blocks too, so the pages are down to 30 and 110 lines.

No `#main-block` / `#aside-block` slots in the end: which component a block is comes from the registry (`componentOf` / `propsOf`), so a slot could only be filled by re-deriving that, or by branching per id in the page. The one genuine per-surface difference is now a typed `organizationId` prop instead of a `Record<string, unknown>` bag. The 19 keys are still listed, but once.

client-app/modules/sales-rep/components/layout-region.vue
ivan-kalachikov: Done — `composables/useBlockTitle.ts`, used by all three call sites.

client-app/modules/sales-rep/layout/registry.ts
ivan-kalachikov: Done — `layout/stat-cards.ts` is now the one table (id, caption, icon, accent). The registry derives its `statistics` blocks from it and the composables fill in only what the queries decide.

`buildStatCards` types `data` as a total record over the table's keys, so a card shipped without its value is a compile error rather than a blank tile. `stat-card-keys.test.ts` now asserts blocks and cards agree on id **and** caption.

client-app/modules/sales-rep/components/layout-skeleton.vue
ivan-kalachikov: Done, and you were right to point at it — I'd talked myself out of it over your padding caveat without reading the component.

Widget boxes are now `<VcWidgetSkeleton head size="md">` with empty divs; the kit styles each as a placeholder line and varies the widths, so `__block` / `__head` / `__body` and two bar variants all went away. The caveat turns out not to matter: at md `--placeholder-h` 1.625rem + `py-3`×2 = 3.125rem, exactly `VcWidget`'s `--header-min-h`, and `__slot` uses the identical `pt-4 pb-5 px-[--p-x]`. The delta is horizontal only, and height is what causes a load jump.

KPI card kept local as you suggested, on the real `.stat-widget` classes so its box can't drift either. File 148 → 110 lines.

client-app/modules/sales-rep/components/layout-block.vue
ivan-kalachikov: Fixed. The controls no longer overlay the header from outside: `layout-block.vue` provides them and a new `layout-widget.vue` — which every sales-rep widget renders instead of `VcWidget` — injects them into the kit's own `#prepend` / `#append` slots, composed with whatever header content the widget already had.

So their placement comes from the widget's real padding and header height, and there is **no `.vc-widget__*` rule left anywhere in the module** — the mirrored `size="md"` metrics are gone with it, and so is the "known limitation" that was in the PR description. The `.vc-widget { mx-0 }` in `customer-profile.vue` was the same category and now aims at our own class.

client-app/modules/sales-rep/components/layout-block.vue
ivan-kalachikov: Done — `$self: &`.

client-app/modules/sales-rep/components/layout-block.vue
ivan-kalachikov: Done, split into two rules.

client-app/modules/sales-rep/components/layout-block.vue
ivan-kalachikov: Done — nested as `&.sortable-ghost`.

client-app/modules/sales-rep/components/layout-block.vue
ivan-kalachikov: Done — nested as `&.sortable-drag`.

client-app/modules/sales-rep/components/layout-edit-bar.vue
ivan-kalachikov: Fixed — `ghost`.

client-app/modules/sales-rep/components/layout-edit-bar.vue
ivan-kalachikov: Fixed — `soft`. Also the two `VcAlert variant="solid-light"` on the pages, which weren't flagged.

client-app/modules/sales-rep/components/layout-edit-button.vue
ivan-kalachikov: Done — `class="layout-edit-button"` on the wrapper.

client-app/modules/sales-rep/components/layout-hidden-tray.vue
ivan-kalachikov: Done — `--vc-icon-color` on `&__restore`; the `.vc-icon` rule is gone.

client-app/modules/sales-rep/components/layout-region.vue
ivan-kalachikov: Done — `$self: &`.

client-app/modules/sales-rep/components/layout-region.vue
ivan-kalachikov: Kept this one — a `v-if` can't do the job here, so I've made the comment say why.

SortableJS moves the card into this container mid-drag and `restore()` puts it back, so `props.entries` doesn't change until the drop. Driven from state, the hint would sit visible *under* a card already dragged in, and linger after the last one was dragged out. `:has()` tracks the gesture instead of the result.

Happy to change it if you see a way I'm missing.

client-app/modules/sales-rep/components/layout-skeleton.vue
ivan-kalachikov: Simplified: 148 → 110 lines, 15 → 10 SCSS rule blocks.

No box is hand-drawn any more — the widget boxes are the kit's `VcWidgetSkeleton` and the KPI boxes the real `.stat-widget` classes. What's left is the column split and the KPI card's four placeholder lines, which have no kit equivalent. Also dropped the per-widget row-count table: matching each widget's page size was false precision, since the real widget renders as many rows as the customer has orders.

client-app/modules/sales-rep/components/layout-stats.vue
ivan-kalachikov: Done — here and the tray label. Kept `<p>` where the content is actually prose (the edit-bar text, the `aria-live` announcer).

client-app/modules/sales-rep/components/layout-stats.vue
ivan-kalachikov: Renamed → `--parked`.

client-app/modules/sales-rep/components/layout-stats.vue
ivan-kalachikov: Renamed → `__parked-zone`.

client-app/modules/sales-rep/components/layout-stats.vue
ivan-kalachikov: Removed — you were right that something was off: it was a no-op, `.stat-widget` already sets `bg-additional-50`.

What actually lets the hatching show through the cards is `opacity-70` on the zone, which that rule could never have fixed. Left as opacity — parked cards should read as set aside, and they stay legible.

client-app/modules/sales-rep/components/layout-block.vue
ivan-kalachikov: Correction to my earlier reply — `$self: &` was a cosmetic answer. It compiled to the same `.layout-block--whole.layout-block--editing`, so nothing about the duplication actually changed.

Looking properly: `layout-block--whole` had **no rule of its own** — it existed only to be compounded. So edit mode folds into the class binding (`'layout-block--whole': dragWhole && editing`) and the rule is now a flat `&--whole`. The built CSS is `.layout-block--whole{`, and `$self` is gone from the file.

Pushed in 86d6f45e3.

client-app/modules/sales-rep/components/layout-region.vue
ivan-kalachikov: Correction to my earlier reply — same as the `layout-block` one, `$self: &` only aliased the duplication; it still compiled to `.layout-region--vertical.layout-region--editing`.

This rule turned out not to be needed at all. The `pt-1` was clearance for the drag chrome that used to be overlaid *above* a widget's header; that chrome now lives inside the header, so the reason for it is gone — and the rule was itself shifting the column 4px on entering edit mode, the same reflow the zone band below it was designed to avoid. Deleted, and `$self` with it.

`--editing` stays on the region as the hook a project restyles edit mode through, now that it carries no styles of its own. There is no `#{&}`, `$self` or compound `.layout-*.layout-*` selector left anywhere in the module.

Pushed in 86d6f45e3.

client-app/modules/sales-rep/components/layout-block.vue
ivan-kalachikov: Follow-up on this rule one level further out: `layout-surface.vue` still cancelled VcWidget's mobile full-bleed with `.layout-surface__main .layout-widget { mx-0 }`. My earlier reply called that "our own class" — it isn't the file's own class, and the ban isn't only about `vc-*`.

VcWidget publishes `--vc-widget-border-color` / `-divide-color` / `-bg-color` / `-radius` and no margin seam, and the bleed comes from `.vc-container .vc-widget` (two classes) inside `@media (width < md)` — so nothing in the module outweighs it without that foreign selector or an `!important`. The override is deleted and the bleed accepted: below 768px a widget sits 6px from the screen edge where the KPI row sits at 24px. Checked at 375px, and the edit-mode outline still fits (2px offset + 1px line needs 3px of the 6px).

A `--vc-widget-margin-x` var on VcWidget would solve it properly, but that belongs in a kit PR rather than this one. Pushed in c367bc8c8.

client-app/modules/sales-rep/components/layout-region.vue
ivan-kalachikov: One more on this rule: `layout-region--editing` is gone as well. Once its rule was deleted it was bound but styled nothing, and a modifier with no rule of its own isn't a state worth a class. I'd kept it as a restyling hook, but the root already carries `.layout-region`, and a project restyling edit mode has `.layout-block--editing`, which does carry rules. Pushed in c367bc8c8.

client-app/modules/sales-rep/components/layout-region.vue
ivan-kalachikov: Revisited this one too. Reading the DOM stays — the reason is unchanged, Sortable moves the card into the container mid-drag and `restore()` puts it back, so `entries` only changes on drop. But the selector no longer names another component's class: it is now `&__empty:not(:only-child)`. The hint and the cards are siblings in one container, so "the hint isn't alone in here" is the same condition, read from this file's own output instead of `.layout-block`. Pushed in c367bc8c8.

client-app/modules/sales-rep/components/layout-skeleton.vue
ivan-kalachikov: Second pass on the KPI card you said to keep local — it was still mirroring the card's metrics, just less visibly than the widget box was. `h-4` / `h-9` / `h-4` / `h-5` were `text-xs` / `text-3xl leading-tight` / `text-xs` / `text-sm` read off `stat-widget.vue`, so changing the card's type scale would have shifted the skeleton silently.

Each bar now wears the class it stands in for (`stat-widget__label`, `__value`, `__sub`, `__delta`) and takes its height from `min-h-[1lh]` — the line box of that class — so the only thing the skeleton owns is which lines exist and how wide they are. `animate-pulse` on the card as well, since `VcWidgetSkeleton` pulses right beside it.

`<StatWidget loading>` turned out not to fit: `loading` renders a `VcLoaderOverlay` spinner rather than placeholder lines, and `label` / `value` / `icon` are required. A `skeleton` prop instead would have given the card a second loading mode next to `loading` and forced those three to become optional, so `stat-widget.vue` is untouched. Pushed in c367bc8c8.

vkrashenko (dismissed):

ivan-kalachikov:
**QA notes — VCST-5367**

**Built beyond the ticket**
- **Edit mode** — explicit Edit → Save / Cancel / Reset. Nothing persists until Save.
- **Keyboard reordering** — Tab to a handle, Space grabs, arrows move, Space drops, Esc restores, tab-out cancels.
- **Screen-reader announcements** — `aria-live` on grab / move / drop / cancel.
- **Hidden tray + parked zone** — widgets hide via ✕ and return from the tray; stat cards drag into a parked zone instead.
- **Skeleton on load** — no blocks render until the saved layout arrives, so the default arrangement never flashes.
- **Failure handling** — failed load = alert + no Edit button; failed save = error, draft kept, still in edit mode.
- **Right rail disappears** when nothing is visible in it.
- **Touch** — press-and-hold ~0.2s to drag; a quick swipe scrolls instead.

**Known gaps — please don't file these**
1. **"Editing layout" button acts as Cancel** — discards the draft, no confirmation.
2. **No unsaved-changes guard** — navigate away or reload while editing and the draft is gone silently.
3. **Keyboard park/restore appends to the end** of the target zone (a mouse drop respects position) and announces no position.
4. **Stat card in edit mode announces only "Reorder {name}"** — value/delta aren't read in that mode.
5. **No concurrency control** — two tabs or two devices: last Save wins, silently.
6. **Widget header is draggable but has no grab cursor** — only the small handle button shows one.

**What to check — lazy loading is split, don't report the second half as a bug**
- **Widgets: yes.** Hide Recent Orders → Save → reload → no `salesRepOrders` request in Network.
- **Stat cards: no.** The KPI numbers come from 3 page-level queries (order stats / cart stats / customer counts) that always run. Hiding a card doesn't remove a request — the cards are data, not components.

client-app/modules/sales-rep/components/layout-widget.vue
```diff
@@ -0,0 +1,120 @@
+<template>
+  <VcWidget :class="['layout-widget', { 'layout-widget--draggable': draggable }]" :title="heading" :size="size">
+    <!-- Through VcWidget's own header slots, so placement comes from its padding instead of metrics
+         mirrored outside it. -->
+    <template v-if="draggable" #prepend>
+      <button
```
goldenmaya: `<VcButton color="secondary" variant="ghost">`

client-app/modules/sales-rep/components/layout-widget.vue
```diff
@@ -0,0 +1,120 @@
...
+      >
+        <VcIcon name="switch-vertical" :size="16" />
+      </button>
+    </template>
+
+    <!-- Composed, not replaced: a widget's own header content (the orders "View all" link) stays put. -->
+    <template v-if="draggable || $slots.append" #append>
+      <slot name="append" />
+
+      <button
```
goldenmaya: Use VcButton as well

client-app/modules/sales-rep/components/layout-widget.vue
ivan-kalachikov: Done — `<VcButton color="secondary" variant="ghost" icon="switch-vertical" icon-size="1rem" size="xs">`. `xs` is 2rem against the old 1.75rem, and VcWidget's `md` header is 50px either way, so nothing shifts. The class keeps only `cursor: grab` and the grabbed state. Pushed in 2fab58ba9.

client-app/modules/sales-rep/components/layout-widget.vue
ivan-kalachikov: Done, same treatment — `color="neutral"` so the rest state stays the neutral-500 it was; hiding is reversible from the tray, so the danger red is hover-only now. Pushed in 2fab58ba9.

goldenmaya (approved):

vas11yev1work (approved):

PR #2405, opened 2026-07-28
feat(VCST-5332): sales rep list sharing with customers

yuskithedeveloper:
## Description

Lets a Sales Rep publish a shopping list to one of their customer organizations, notify that organization's members
by email/push with an optional custom message, and lets those members shop the list without being able to edit it.

Introduces a new `Customer` wishlist sharing scope, surfaced in the list settings modal for Sales Reps only.

### Requires backend — all three still open

The `Customer` scope does not exist in the schema without these, so this PR must not merge before they do:

- VirtoCommerce/vc-module-cart#192 — persistence for the sharing target
- VirtoCommerce/vc-module-x-cart#136 — `sharedWithId` on the sharing setting + the scope extension point
- VirtoCommerce/vc-module-sales-rep#8 — the `Customer` scope itself, plus authorization

### Sales Rep flow

1. List settings → **Sharing options** → `Customer`. Visible only when `SalesRep.Enabled` is on for the store **and**
   the user holds `sales-rep:access`.
2. Pick a customer organization from a searchable dropdown of the rep's own served customers.
3. Optionally write a message, and choose the channels (Email / Push — both on by default).
4. Save persists the list and its sharing target in a single `changeWishlist` / `createWishlist` call, then notifies.

The message is optional: left empty, a localized default carries the list name, and the shared-list link is always
appended. The rep also gets the copyable sharing link for the `Customer` scope.

Notification fires only when the save sets a *new* target — non-Customer → Customer, or a switch to a different
customer — so re-saving an unchanged list never re-notifies. Clearing both channels saves the list without notifying.
Delivery outcomes reuse the per-channel warning codes already localized for VCST-5310.

### Customer flow

An org member opening the link sees a "Recommended by your sales representative" badge and can add items to their own
cart. The list itself stays read-only: no remove, no list-quantity save.

### Notable decisions

- Copy follows the backend notification template from VCST-5310 ("your sales representative"); the
  `recommended_by_rep` string was realigned to match it. Channel labels are reused from the Sales Rep module locale
  so the two modals cannot drift apart.
- The message length cap is derived from the actual sharing-link length, because the backend enforces a 1000-char
  combined limit and the link's host varies per store.
- Leaving a targeted scope simply omits `sharedWithId`. An earlier revision of this branch sent an explicit `null` on
  the theory that omitting it would leave a stale target; that was wrong — `CartSharingService.ApplyScope` passes
  `sharedWithId: null` unconditionally for every built-in scope, so the backend detaches the target on its own.
- `Customer` stays selectable for a non-rep opening an already-shared list, so saving cannot silently overwrite the
  scope; the rep-only controls stay hidden for them.
- One target per list: the backend keeps a single sharing setting, so switching customers detaches the previous one.
  The picker says so inline.
- The rep gate is now a single exported `isSalesRepUser()` predicate instead of a third inlined copy of
  `SalesRep.Enabled + sales-rep:access`. `routes.ts` and `index.ts` can adopt it in a follow-up — left untouched here
  to avoid conflicting with the open VCST-5367 branch.
- Generated types carry **only** the sharing delta (`sharedWithId`, `Customer`). An earlier revision of this branch
  had unrelated drift from codegen run against a mismatched backend; that is reverted, and every `*Document` in core
  is byte-identical to codegen output.

### Architecture: the module contributes, core does not import it

Core does not reference the Sales Rep module. `sharedWithId` is declared by vc-module-cart / vc-module-x-cart, so it
lives in core; the `Customer` scope value is the module's (`ModuleConstants.Sharing.CustomerScope`), so it lives there,
and `WishlistScopeType.Customer` has zero hits in hand-written files. Three seams carry it across:

| Surface | Mechanism |
| --- | --- |
| Provenance badge on the shared list | New `sharedList` extension-registry category; the module registers the alert and owns its copy |
| Whether a shared list is shoppable | A `shoppable` capability on the scope — the page asks the registry, not whether a module is installed |
| The scope's controls in the list settings modal | `useWishlistSharingScopes`, a domain-local registry in the shape of the existing `useCartExtensionPoints` / `useBulkOrderExtensionPoints` |

A scope contributes `{scope, labelKey, statusKey, supportsLink, shoppable, isAvailable, element}`. The part with no
precedent in the repo is the **save transaction**: the registry is a global populated at module `init()`, while the
selected target, message and channels are per-open, so `canSave` / `dirty` / `payload` / `onSaved` come from the
rendered instance through `defineExpose` and a template ref rather than from the registration object. The modal folds
them into a single save — the scope gates Save, contributes to the command, and runs its follow-up once the list is
persisted, where a failure is logged and warned about instead of being reported as a failed save.

Shoppability stayed scope-specific rather than being opened up for link- and organization-shared lists: that is a
product decision, not a side effect of this refactor.

### Out of scope

- Multi-customer sharing, "Share" as a separate action-wheel entry, and conversion tracking — excluded from the story.
- Customer-side discovery in **My lists**. A `Customer`-scoped list is owned by the rep, with the target held only in
  the sharing setting, so the wishlist search does not return it for the target org's members. The email link is
  therefore the only entry point; listing them needs a backend story.
- Server-side keyword search in the customer picker. The first 100 served customers are fetched and filtered
  client-side; overflow is logged rather than silently hidden. To be revisited.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5332
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.55.0-pr-2405-ad5b-ad5bffe6.zip

### Unit — 73 tests, 6 files

Core's tests use a **fake contributed scope**, so they name no module and would still pass with the Sales Rep module
deleted:

- `add-or-update-wishlist-modal.test.ts` (19) — which scopes are listed, a persisted-but-unavailable scope staying
  listed while its controls stay hidden, what the provider's controls receive, Save gating, payload merging, and the
  follow-up contract including a failing `onSaved` not turning into a save error.
- `shared-list.test.ts` (10) — the provenance socket being offered with the sharing setting, shoppability driven by the
  scope's capability, and the cart paths (add / change quantity / no-op) with the read-only invariant.
- `wishlist-status.test.ts` (5) — owner vs. recipient wording, and a contributed scope's own status key.

Module-side:

- `wishlist-customer-sharing.test.ts` (23) — the picker (options, pre-selection, required, loading, failed load,
  replacement warning), the message and channels, the exposed contract, and every delivery outcome including warning
  codes the storefront cannot translate.
- `useSalesRepCustomerOptions.test.ts` (11) — query shape, `enabled` gate, mapping, name fallback, overflow warning,
  and the new failed-load flag.
- `useSalesRepsConfig.test.ts` (5) — the shared rep predicate.

Full suite: 1527 / 1534 pass. The 7 failures are load-dependent timeouts in unrelated files (analytics, theme
context, navigations, languages) — the failing set changes between runs and every one of them passes in isolation. `eslint`, `vue-tsc`, `prettier` and `check-locales` clean.

### Verified manually against a dev backend

Run as a user with `sales-rep:access` on a store with `SalesRep.Enabled`:

- List card shows "Shared with customer" for the `Customer` scope.
- Re-opening an unchanged Customer list: scope and target pre-filled, target resolved to the organization name,
  message/channels hidden, **Save disabled**.
- Picker lists exactly the rep's served customers.
- Switching the target: replacement hint appears, message field and both channel checkboxes appear, Save enables.
- Message counter reflects the derived cap (1000 − link length − 2).
- Saving with both channels cleared persists the new `sharedWithId` and issues **no** `sendCustomerCommunication`.
- Shared-list page: badge renders, add-to-cart works (new cart line), changing the quantity and updating modifies the
  same cart line rather than adding a second one, and the list's own item quantity is left untouched.

### Still needs QA with other accounts

- Actual email/push delivery to a customer organization's members.
- A member of a non-target organization must get no access (the rep owns the list, so their own session cannot prove this).
- A user without `sales-rep:access` must not see the `Customer` option.


ivan-kalachikov:
### Core now imports the sales-rep module

Direction should always be module → core. At the merge base the only file outside `client-app/modules/sales-rep/` referencing it was `app-runner.ts`. This PR adds four more, two of them in core shared code:

- [`add-or-update-wishlist-modal.vue#L146-L150`](https://github.com/VirtoCommerce/vc-frontend/blob/a6e8d04a0341a33b6e97561cb0cf0d8800e3e461/client-app/shared/wishlists/components/add-or-update-wishlist-modal.vue#L146-L150) — `useSalesRepCommunication`, `useSalesRepCustomerOptions`, `isSalesRepUser`
- [`shared-list.vue#L79-L81`](https://github.com/VirtoCommerce/vc-frontend/blob/a6e8d04a0341a33b6e97561cb0cf0d8800e3e461/client-app/pages/shared-list.vue#L79-L81) — `isSalesRepsEnabled`

Plus core logic branching on the module's scope value in [`wishlist-status.vue#L34-L36`](https://github.com/VirtoCommerce/vc-frontend/blob/a6e8d04a0341a33b6e97561cb0cf0d8800e3e461/client-app/shared/wishlists/components/wishlist-status.vue#L34-L36), `shared-list.vue:103`, and the modal at `:218`, `:229-233`, `:264`. This also makes `PORT_TO_MF.md` false — it maps only outbound imports, so a reverse remap now leaves core importing a removed directory.

**On the split, to be clear** — I checked all three backend PRs. `sharedWithId` is *core*: declared by VirtoCommerce/vc-module-cart#192 (`CartSharingSetting.SharedWithId` + migrations) and VirtoCommerce/vc-module-x-cart#136 (the GraphQL fields on `InputChangeWishlistType` / `InputCreateWishlistType` / `SharingSettingType`). VirtoCommerce/vc-module-sales-rep#8 only consumes it. So keeping `sharedWithId` in core FE is correct — leave it as is. What's module-owned is the `Customer` scope value (`ModuleConstants.Sharing.CustomerScope`, `SalesRepWishlistScopeType`). Mirror the backend's own layering.

Acceptance test: `WishlistScopeType.Customer` should have zero hits in hand-written core files. It will stay in the generated `types.ts` — that's expected, don't hand-edit it.

Three surfaces, in ascending order of new work. `sales-rep/index.ts:37-41` already uses the first mechanism for its nav links:

1. **Badge** → `useExtensionRegistry` + `<ExtensionPoint>`, per `back-in-stock/index.ts:67`. The module registers the alert and owns the `recommended_by_rep` string. Core passes `:sharing-setting` through `$attrs` and gates on `$canRenderExtensionPoint(...)`, so the module owns the condition body — same shape as `product-card.vue:66-69`. Existing mechanism, no extension needed.

2. **`addable-to-cart`** → no mechanism needed at all. Shoppability is core, but [`:addable-to-cart="isRecommendedByRep"`](https://github.com/VirtoCommerce/vc-frontend/blob/a6e8d04a0341a33b6e97561cb0cf0d8800e3e461/client-app/pages/shared-list.vue#L23-L25) gates a core capability on module install — which is also why link-shared and org-shared lists are silently non-shoppable. Just stop gating it on the module; let the provider declare `shoppable` if it must stay scope-specific.

3. **Modal** → a domain-local registry like `useCartExtensionPoints.ts` / `useBulkOrderExtensionPoints.ts` (see `quotes/index.ts:19`, `purchase-requests/index.ts:18`). Being precise about how much of this already exists, because it's only half:

   - **Has a precedent.** `ITab` (`{ id, icon, label, hidden, element }`) already contributes a labelled, conditionally-visible entry plus a component — which maps cleanly onto the scope *option*: `{ scope, labelKey, statusKey, supportsLink, shoppable, isAvailable, component }`.
   - **Is new.** Nothing in the repo participates in a **save transaction**. The scope also has to contribute `sharedWithId` to the mutation payload, take part in `canSave`, and run notify after a successful save. No existing seam does any of that.

   One constraint worth getting right up front: the registry is a global singleton populated at module `init()`, while the modal is per-open with per-instance state (selected org, message, channels). So `resolvePayload` / `canSave` / `onSaved` can't be static functions on the registration object — they have no access to that state. They should come from the rendered component instance instead, either via `defineExpose` + a template ref or emitted upward as a single contribution object.

Worth doing here rather than deferring: it also moves all 19 `share_*` keys out of the 13 core locale files, and it folds in the picker's missing `required`/error/loading affordances plus the `:182` comment that claims this gate "can't drift from the hub's own gate".

If it has to be staged, 1 and 2 are small and remove one import site outright — but 4 of the 5 inbound edges are in the modal, so deferring 3 is deferring the finding.

client-app/shared/wishlists/components/add-or-update-wishlist-modal.vue
```diff
@@ -166,32 +261,126 @@ const { value: name } = useField<string | undefined>("name");
...
+
+    if (isEditMode.value) {
+      await updateWishlist({ listId: props.list!.id, ...payload });
+    } else {
+      await createWishlist(payload);
+    }
+
+    // Notify the customer's members (reuses the VCST-5310 channel) when this edit set a genuinely new target and
+    // at least one channel is on. The rep's own message is optional — a default one carries the link on its own.
+    if (isNewCustomerTarget.value && sharedWithId && (sendEmail.value || sendPush.value)) {
```
ivan-kalachikov: Notification is skipped when the post-save refetch fails.

`updateWishlist` calls `fetchWishlists()` outside its try/catch, and that rethrows. So after `changeWishlist` already succeeded, a refetch hiccup lands in this catch — before the notify check. User sees "could not be saved"; the customer org gets access with no email/push, and that link is their only way in.

**Needed:** notify on the mutation's success, not the whole `updateWishlist` call.

client-app/shared/wishlists/components/add-or-update-wishlist-modal.vue
```diff
@@ -63,6 +64,53 @@
...
+            :disabled="saving"
+            :items="customerOptions"
+            :message="
+              replacesPreviousTarget ? $t('shared.wishlists.add_or_update_wishlist_modal.share_replace_hint') : ''
+            "
+            text-field="organizationName"
+            value-field="organizationId"
+            autocomplete
+            clearable
+          />
```
ivan-kalachikov: Persisted target renders as "nothing selected".

`VcSelect` gets its label from `items.find(...)`, else falls back to placeholder (`vc-select.vue:271`). `items` is capped at 100 — so a target that is unloaded, over-cap, or lost to a query error looks unshared. Save disables with no reason. Re-picking detaches the real target and notifies.

**Needed:** seed `items` with the persisted target so `find()` always hits. Add `required` too. Not `:error` — no touched signal, and `:message` already holds the replace hint.

client-app/modules/sales-rep/composables/useSalesRepCustomerOptions.ts
```diff
@@ -0,0 +1,53 @@
...
+      organizationId: customer.organizationId,
+      organizationName: customer.organizationName ?? customer.organizationId,
+    })),
+  );
+
+  const totalCount = computed(() => result.value?.salesRepCustomers?.totalCount ?? 0);
+
+  // Client-side filtering silently hides everything past OPTIONS_LIMIT, which would look like a missing customer
+  // rather than a truncated list. Surface it until the picker moves to server-side keyword search.
+  watch(totalCount, (count) => {
```
ivan-kalachikov: Never fires on a cache hit — the case it exists for.

No `immediate`, but `useQuery` fills `result` during setup, so `totalCount` is already over the cap at first read and never *changes*.

**Needed:** `{ immediate: true }`.

client-app/modules/sales-rep/composables/useSalesRepCustomerOptions.ts
```diff
@@ -0,0 +1,53 @@
...
+    storeId: globals.storeId,
+    first: OPTIONS_LIMIT,
+    after: "0",
+    keyword: "",
+    sort: "name:asc",
+  }));
+
+  // `enabled` keeps the query from firing for users who can't share (non-reps / create mode), avoiding a
+  // needless authorized request. Defaults to enabled for standalone callers.
+  const { result, loading, onError } = useQuery(SalesRepCustomersDocument, variables, { enabled });
```
ivan-kalachikov: Hub-table query used to fill an id+name dropdown.

Per customer it selects 2x `orderStatistics`, `lastOrder`, `address`, `accountType`. The four date vars aren't passed and are nullable, so both aggregations run all-time. ~200 unbounded aggregations for 100 orgs.

**Needed:** a narrow document — `totalCount` + `items { organizationId organizationName }`.

client-app/modules/sales-rep/composables/useSalesRepCustomerOptions.ts
```diff
@@ -0,0 +1,53 @@
...
+// Flat list of the caller rep's served customer organizations for the "share with customer" picker
+// (resolved server-side from the caller's claims). Distinct from useSalesRepCustomers, which is the paged table.
+export function useSalesRepCustomerOptions(enabled?: Ref<boolean> | boolean) {
+  const variables = computed(() => ({
+    storeId: globals.storeId,
+    first: OPTIONS_LIMIT,
+    after: "0",
+    keyword: "",
+    sort: "name:asc",
+  }));
```
ivan-kalachikov, line 21: Create mode does fetch — the only gate passed is `canShareWithCustomers`, and the Customer scope works in create mode too.

**Needed:** drop "/ create mode".

client-app/pages/shared-list.vue
```diff
@@ -127,6 +141,31 @@ function selectItemEvent(item: Product | undefined): void {
...
+      if (itemInCart.quantity !== quantity) {
+        await changeItemQuantity(itemInCart.id, quantity);
+      }
+    } else {
+      await addToCart(lineItem.product.id, quantity);
+    }
+  } finally {
+    pendingItems.value[lineItem.id] = false;
+  }
+}
```
ivan-kalachikov, line 179: Drops the analytics the mirrored function deliberately has.

`list-details.vue` also calls `trackAddItemToCart` (:342), `pushHistoricalEvent` (:343), `showResultModal` (:349) — added in 26c33dbc0 / 9464a6ef9, kept through 72e92d164. This page still fires `viewItemList` and `selectItem`, so the funnel breaks exactly at conversion.

**Needed:** fire the two tracking calls. If deliberate, say "minus analytics/result modal" instead of "Mirrors".

client-app/pages/shared-list.vue
```diff
@@ -4,6 +4,10 @@
       {{ list.name }}
     </VcTypography>
 
+    <VcAlert v-if="isRecommendedByRep" color="info" variant="solid-light" size="sm" icon class="shared-list__rep-badge">
```
ivan-kalachikov: `solid-light` is deprecated — `vc-alert.types.d.ts:8` says use `soft`. The modal in this PR already uses `soft` (:60).

**Needed:** `variant="soft"`.

client-app/shared/wishlists/components/add-or-update-wishlist-modal.vue
```diff
@@ -107,20 +158,47 @@ interface IProps {
...
+// and edit mode. Gated on the shared rep predicate, so this surface can't drift from the hub's own gate.
+const canShareWithCustomers = computed(isSalesRepUser);
+// Only fetched when the caller can share (avoids an authorized request for everyone else).
+const { options: customerOptions } = useSalesRepCustomerOptions(canShareWithCustomers);
+
+// The backend rejects a combined message over 1000 chars, and the sharing link is appended to whatever the rep
+// writes. Measuring the actual link (its host varies per store) keeps a long domain from silently pushing an
+// otherwise-valid message over the limit.
+const NOTIFICATION_MESSAGE_LIMIT = 1000;
+const SEPARATOR_LENGTH = 2;
```
ivan-kalachikov: Measures the `"\n\n"` at :359, but nothing ties them together. Correct today; change the separator and the 1000-char cap silently stops matching.

**Needed:** one `const SEPARATOR = "\n\n"` — `SEPARATOR.length` here, `join(SEPARATOR)` there.

client-app/shared/wishlists/components/add-or-update-wishlist-modal.vue
```diff
@@ -63,6 +64,53 @@
...
+              <VcLabel class="add-or-update-wishlist-modal__channels-label">
+                {{ $t("shared.wishlists.add_or_update_wishlist_modal.share_channels_label") }}
+              </VcLabel>
+
+              <div class="add-or-update-wishlist-modal__channels">
+                <VcCheckbox v-model="sendEmail" :disabled="saving" data-test-id="wishlist-share-email-checkbox">
+                  {{ $t("shared.wishlists.add_or_update_wishlist_modal.share_email_label") }}
+                </VcCheckbox>
+
+                <VcCheckbox v-model="sendPush" :disabled="saving" data-test-id="wishlist-share-push-checkbox">
```
ivan-kalachikov: Bare attribute + no `inheritAttrs: false` in `VcCheckbox` = lands on the root `div`, not the input. An E2E selector hits a non-interactive wrapper.

`VcCheckbox` has a `testId` prop that binds it to the `<input>` (`vc-checkbox.vue:27`).

**Needed:** `:test-id="..."` on both. (Tests can't catch it — their stub is a bare `<input>`.)

client-app/shared/wishlists/components/add-or-update-wishlist-modal.vue
```diff
@@ -107,20 +158,47 @@ interface IProps {
...
-const { loading, createWishlist, updateWishlist } = useWishlists();
+// Modal-owned busy flag for the save action. Driven with try/finally so the Save button can never get stuck
+// showing the loader (useWishlists' shared `loading` can leak true on error), and it guards against double-submit.
+const saving = ref(false);
 const { isCorporateMember } = useUser();
 
+const isEditMode = computed<boolean>(() => !!props.list);
+
+// "Customer" scope (VCST-5332): a Sales Rep publishes a list to a customer organization. Available in both create
+// and edit mode. Gated on the shared rep predicate, so this surface can't drift from the hub's own gate.
```
ivan-kalachikov: "can't drift from the hub's own gate" isn't true — nothing links them. `routes.ts:24` and `index.ts:68` still spell the pair out inline; they just happen to agree today.

**Needed:** drop the can't-drift claim. (The description already covers why those are untouched.)

locales/ru.json
```diff
@@ -1009,9 +1010,24 @@
...
+        "share_customers_label": "Клиент",
+        "share_customers_placeholder": "Выберите клиента",
+        "share_replace_hint": "Ранее выбранный клиент потеряет доступ к этому списку.",
+        "share_message_label": "Сообщение",
+        "share_message_placeholder": "Необязательное сообщение",
+        "share_channels_label": "Отправить через",
+        "share_email_label": "Эл. почта",
+        "share_push_label": "Push-уведомление",
+        "share_default_title": "Новый список от вашего торгового представителя",
+        "share_default_message": "Здравствуйте! Я поделился с вашей организацией списком «{listName}». Посмотрите:",
```
ivan-kalachikov: "Я поделился" is masculine, but the rep can be any gender — and this is customer-facing.

fr and de avoid gender entirely, so a neutral form works here too, e.g. "Вам доступен список «{listName}»".

locales/pl.json
```diff
@@ -1009,9 +1010,24 @@
...
+        "share_customers_label": "Klient",
+        "share_customers_placeholder": "Wybierz klienta",
+        "share_replace_hint": "Wcześniej wybrany klient utraci dostęp do tej listy.",
+        "share_message_label": "Wiadomość",
+        "share_message_placeholder": "Opcjonalna wiadomość",
+        "share_channels_label": "Wyślij przez",
+        "share_email_label": "E-mail",
+        "share_push_label": "Powiadomienie push",
+        "share_default_title": "Nowa lista od Twojego przedstawiciela handlowego",
+        "share_default_message": "Cześć! Właśnie udostępniłem Twojej organizacji listę „{listName}”. Sprawdź ją:",
```
ivan-kalachikov: "Właśnie udostępniłem" is masculine, but the rep can be any gender — and this is customer-facing.

fr and de avoid gender entirely, so a neutral form works here too, e.g. "Lista „{listName}” jest już dostępna dla Twojej organizacji".

ivan-kalachikov (commented):
Inline notes below. The import-direction point is in a separate comment above; this is the rest of the review.

One correction to that comment: it mentions the picker missing `required` / `:error` / loading affordances. `VcSelect` has no `loading` prop, and `:error` is the wrong tool here — the picker comment below says what's actually needed instead.

Also worth recording two things I flagged while reviewing and then withdrew after reading the backend PRs: the sharing link is **not** double-appended, and leaving the `Customer` scope **does** clear the stored target (`CartSharingService.ApplyScope` passes `sharedWithId: null` for the built-in scopes). Both are fine as written.

locales/pt.json
```diff
@@ -1009,9 +1010,24 @@
...
+        "share_replace_hint": "O cliente selecionado anteriormente perderá o acesso a esta lista.",
+        "share_message_label": "Mensagem",
+        "share_message_placeholder": "Mensagem opcional para incluir",
+        "share_channels_label": "Enviar via",
+        "share_email_label": "E-mail",
+        "share_push_label": "Notificação push",
+        "share_default_title": "Uma nova lista do seu representante de vendas",
+        "share_default_message": "Olá! Acabei de compartilhar a lista \"{listName}\" com a sua organização. Confira:",
+        "share_success": "Lista compartilhada com o cliente.",
+        "share_partial": "Lista compartilhada, mas não todas as notificações foram entregues.",
```
ivan-kalachikov: "mas não todas" should be "mas **nem** todas".

client-app/shared/wishlists/components/add-or-update-wishlist-modal.vue
```diff
@@ -166,32 +196,65 @@ const { value: name } = useField<string | undefined>("name");
...
+  try {
+    // One mutation for the list and its sharing. A scope without controls contributes nothing: the backend applies a
+    // null target for its own non-targeted scopes.
+    const scopePayload = scopeControls.value?.payload ?? {};
+    const payload = {
       listName: name.value?.trim(),
       description: description.value?.trim(),
       scope: sharingScope.value,
       sharingKey: sharingKey.value,
+      ...scopePayload,
```
ivan-kalachikov: New, from the refactor: a listed-but-unavailable scope loses its persisted target.

When the scope is listed (:155-158) but unavailable, `activeScopeElement` is undefined by design (:168), so `scopeControls` is null and `scopePayload` is `{}` — `sharedWithId` is not sent. Backend then rejects: `AuthorizeCustomerShareAsync` throws Forbidden on `string.IsNullOrEmpty(context.SharedWithId)`.

Trigger: a rep still serving the org, but `SalesRep.Enabled` gets switched off store-wide (or the permission is revoked). They can no longer rename their own Customer-shared lists — generic "could not be saved". Pre-refactor this worked, because `sharedWithId` came from `listSharedWithId` regardless of rep-ness.

**Needed:** when the active scope is listed but unavailable, carry `listSharedWithId` into the payload.

ivan-kalachikov (commented):
Re-reviewed at `edda6ae`. **All 13 inline findings are fixed** — threads resolved. Both acceptance tests from the architecture comment now pass:

- only `app-runner.ts` references `modules/sales-rep` (the init call, same as before this PR)
- `"Customer"` appears in exactly one hand-written place: the module's own `CUSTOMER_SHARING_SCOPE`

`useWishlistSharingScopes` came out well, including the per-instance constraint — `canSave`/`dirty`/`payload`/`onSaved` are exposed from the rendered element rather than the registration object, and the JSDoc says why. Two things I'd have missed: gating the provenance note on the scope alone (the viewer is the customer, not a rep), and firing the conversion only in the genuine-add branch. Also nice that the refactor made the styling consistent — the modal is Tailwind-only again and the new component is fully BEM.

I ran the six touched spec files locally: **77 passed**. `ci` is green.

One new low-severity thing below, introduced by the refactor. Two notes, neither blocking:

- `useSalesRepsConfig.ts`'s comment still says "so rep-only surfaces cannot drift apart" while `routes.ts:24` and `index.ts:89` keep the pair inline. Much weaker than the old claim, and the description explains why they're untouched — just flagging that the sentence is still slightly ahead of the code.
- The three `auto-tests` matrix jobs fail at "Run Auto Tests". I could not get the cause from the API, and there are two plausible unrelated explanations (that workflow checks out `vc-testing-module@dev` unpinned, and the `Customer` scope isn't in the deployed test schema until the backend PRs merge). Worth a glance rather than assuming it's this diff.

client-app/modules/sales-rep/index.ts
```diff
@@ -2,9 +2,12 @@ import { computed, defineAsyncComponent } from "vue";
 import { useNavigations } from "@/core/composables/useNavigations";
 import { useUser } from "@/shared/account/composables/useUser";
 import { useExtensionRegistry } from "@/shared/common/composables/extensionRegistry/useExtensionRegistry";
+import { EXTENSION_NAMES } from "@/shared/common/constants/extensionPointsNames";
+import { useWishlistSharingScopes } from "@/shared/wishlists";
```
ivan-kalachikov: Barrel import puts all of `shared/wishlists` into the boot chunk.

`@/shared/wishlists` re-exports `./components` — 12 SFCs, 5 with `<style>` blocks — plus `useWishlists`, whose top-level `ref()` calls make it side-effectful, so it drags the account GraphQL ops along. `package.json` declares no `sideEffects`, so none of that can be tree-shaken away.

This file is the **first eager consumer** the barrel has ever had: `app-runner.ts` imports it statically and calls `initSalesRep` at boot, while all 8 pre-existing importers were lazy pages or leaf components. Net effect is wishlist CSS/JS in the entry chunk for every visitor, plus a CSS extraction-order shift.

**Needed:** the subpath — `@/shared/wishlists/composables/useWishlistSharingScopes`. Lines 3-5 already import that way. (The `import type` from the barrel in `wishlist-customer-sharing.vue` is fine — erased at build.)

ivan-kalachikov (commented):
One more from a second review pass on the refactor. The seam itself is solid — I verified the `defineExpose` + template-ref wiring is genuinely reactive (Vue wraps `instance.exposed` in `proxyRefs`, so the exposed computeds unwrap and track), registration reliably precedes first render, and the whole scope transition table is correct. This is a bundling issue, not a logic one.

client-app/shared/wishlists/components/add-or-update-wishlist-modal.vue
```diff
@@ -166,32 +196,65 @@ const { value: name } = useField<string | undefined>("name");
...
+    if (isEditMode.value) {
+      await updateWishlist({ listId: props.list!.id, ...payload });
+    } else {
+      await createWishlist(payload);
+    }
+
+    // Saved from here on, so neither the follow-up nor the refresh may surface as a save error. Both are awaited to
+    // keep the loader up.
+    try {
+      await scopeControls.value?.onSaved?.({
```
ivan-kalachikov, line 252: Dismissing the modal mid-save silently loses the notification.

Cancel, Esc, backdrop and the header X are all ungated while `saving` is true. `close()` starts a 200ms leave transition, `@after-leave` pops the modal off the stack, and the unmount nulls `scopeControls` — so this line becomes a no-op once the mutation resolves, which for a network call is well past 200ms.

Result: the list is persisted **with** the new `sharedWithId`, so the customer has access, but no email or push is ever sent and nothing warns. Per the description that link is their only entry point.

**Needed:** don't allow dismissal during a write —

```
<VcModal :is-persistent="saving" ...>            <!-- covers Esc + backdrop (:16) and the header X (:43) -->
<VcButton :disabled="saving" @click="close">     <!-- Cancel is in this slot, needs its own guard -->
```

That keeps the tree alive through the await, so the exposed computeds and the `fetchWishlists()` refresh below also still work. Capturing `onSaved` into a local before the mutation is not sufficient on its own: the function survives unmount but the `isNewTarget` computed it reads does not.

ivan-kalachikov (commented):
One more, the only one I'd call blocking.

client-app/pages/shared-list.vue
```diff
@@ -4,6 +4,14 @@
       {{ list.name }}
     </VcTypography>
 
+    <ExtensionPoint
+      v-if="$canRenderExtensionPoint('sharedList', EXTENSION_NAMES.sharedList.provenanceNote, list?.sharingSetting)"
+      :name="EXTENSION_NAMES.sharedList.provenanceNote"
+      category="sharedList"
+      :sharing-setting="list?.sharingSetting"
```
ivan-kalachikov: This object binding ends up in the DOM as `sharing-setting="[object Object]"`.

`wishlist-rep-provenance.vue` declares no props and doesn't set `inheritAttrs: false`, and neither does `VcAlert` — so `ExtensionPoint`'s `$attrs` spread falls through to the alert's root `div`.

Nothing consumes it: the badge references `sharingSetting` zero times, and the gating condition already receives it separately as the third argument to `$canRenderExtensionPoint` on the line above.

**Needed:** drop this line. The category's `Props` type stays satisfied — it types the `props?` field on the registration object (read via `getProps`), not runtime attrs. If a provider ever needs the sharing setting for rendering, add it back then.

ivan-kalachikov (commented):
Minor one.

ivan-kalachikov (dismissed):

ivan-kalachikov (approved):

PR #2410, opened 2026-07-31
feat(VCST-4933): support Shared Components in Page Builder preview

ToxaKZ:
## Description

Adds the Storefront preview support required by Page Builder Shared Components.

The published Storefront does not resolve or fetch Shared Components at runtime. The Page Builder module expands `componentRef` placements before Pages indexing, so public Storefront routes continue to receive ordinary Page Builder blocks.

This PR implements the authoring-preview side:

- reactive VPPageBuilder content replacement;
- a strict Designer/Storefront preview protocol;
- one visual boundary around every multi-section Shared Component placement;
- hover, selection, and scroll synchronization between Designer and preview;
- reliable iframe reconnect behavior;
- trusted endpoint-origin validation.

### Companion PageBuilder module PR

https://github.com/VirtoCommerce/vc-module-pagebuilder/pull/159

### Reactive Page Builder content

`vp-page-builder.vue` reparses incoming content whenever the Pages document changes. It updates the rendered page for cache/network replacements, validates the expected Page Builder shape, clears stale content when the new payload is empty or malformed, and preserves ordinary block rendering.

### Preview protocol and security

The preview `postMessage` contract is centralized and validated.

Messages are accepted only when:

- the source is the current parent Designer window;
- the origin matches the configured Designer origin;
- the payload has the expected `source`;
- the message type and fields conform to the protocol.

Asynchronous handling is serialized so an older locale or navigation update cannot overwrite newer preview state.

The `ep` query parameter accepts only valid HTTP(S) URLs and is reduced to its clean origin. Malformed values and `javascript:`/`file:` endpoints are rejected.

### Shared Component placement overlay

Designer sends a resolved template together with placement metadata:

```ts
{
  placementId,
  componentRef,
  startIndex,
  count,
  label
}
```

The preview:

1. creates stable anchors around rendered Page Builder sections;
2. maps each boundary to its first section and following section/end anchor;
3. draws one boundary around all sections belonging to the placement;
4. recalculates geometry after content, layout, or viewport changes;
5. sends placement-level hover and selection events back to Designer;
6. scrolls to the complete placement when Designer selects it.

```mermaid
sequenceDiagram
    participant D as Page Builder Designer
    participant I as Storefront preview iframe
    participant O as Shared Component overlay

    I->>D: preview loaded and supported blocks
    D->>I: resolved template and placement boundaries
    I->>O: render anchors and boundaries
    O->>D: hover/select placementId
    D->>I: select/hover placementId
    I->>O: highlight or scroll to placement
```

Overlay elements are authoring-only and are not part of business content.

### Iframe lifecycle

A newly loaded or replaced iframe must complete its own `preview-loaded` handshake before queued messages are sent. Ready state from an old iframe is not reused.

This prevents updates from being posted before the new iframe has installed its listener.

### Published-page behavior

```mermaid
flowchart LR
    PB["Page Builder raw content"] -->|"resolve componentRef"| PI["Pages index"]
    PI -->|"ordinary expanded blocks"| SF["Storefront VPPageBuilder"]
```

There is no production-time Storefront request to the Shared Components API. A component update becomes visible after the PageBuilder propagation job causes Pages to reindex affected published documents.

### Verification

Covered scenarios include VPPageBuilder content replacement, malformed content clearing, protocol validation, message ordering, iframe reconnect, boundary normalization, overlay rendering/resize/hover/click/selection/scrolling, and endpoint validation.

### QA-test:

Assume the Page Builder module and this Storefront build are already installed and configured.

#### 1. Ordinary content regression

1. Open a Page Builder page containing only ordinary sections.
2. Verify Designer preview loads and renders every block.
3. Add, edit, remove, reorder, hover, and select ordinary sections.
4. Verify preview updates and bidirectional selection still work.
5. Publish and verify the normal public Storefront route renders correctly.

#### 2. Shared Component preview boundary

1. Open a page containing a Shared Component with at least three visible sections.
2. Verify all sections render in preview.
3. Verify one boundary surrounds the complete component rather than each individual section.
4. Resize the preview and change content height; verify the boundary realigns.
5. Place the same component twice and verify two independent boundaries.

#### 3. Selection and navigation synchronization

1. Hover the component in Designer and verify its preview boundary highlights.
2. Hover/click the boundary and verify Designer receives placement hover/selection.
3. Select an off-screen component in Designer and verify preview scrolls to the beginning of the full placement.
4. Verify ordinary sections retain section-level selection and scrolling.

#### 4. Live component editing

1. Open **Edit original** for a Shared Component.
2. Edit individual sections/blocks, including an image.
3. Add, remove, and reorder component sections.
4. Save and verify preview redraws the updated component and correct boundary.
5. Detach one placement and verify its boundary disappears while its independent blocks remain visible.

#### 5. Iframe lifecycle

1. Open Designer and wait for preview.
2. Reload or recreate the preview iframe.
3. Immediately perform several page edits.
4. Verify messages are applied only after the new iframe announces readiness.
5. Verify no edit is lost and old iframe state is not reused.
6. Navigate quickly between pages/components and verify the newest state wins.

#### 6. Published Storefront integration

1. Publish a page containing a Shared Component and wait for Pages indexing.
2. Open the regular public Storefront URL, not `/designer-preview`.
3. Verify the component renders as ordinary blocks.
4. Verify no preview boundary, blocker, anchors, or Designer messaging appears.
5. Edit the component original, wait for propagation/reindexing, and reload the public page.
6. Verify the updated component appears without republishing the page.
7. Verify ordinary page content remains unchanged.

#### 7. Security validation

1. Open Designer through the expected platform URL and verify preview connects.
2. Verify malformed, `javascript:`, and `file:` `ep` values do not establish a trusted Designer origin.
3. Verify a valid HTTP(S) endpoint is normalized to its origin.
4. Verify messages from an unexpected origin or window are ignored.

### Jira-link:

https://virtocommerce.atlassian.net/browse/VCST-4933

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.58.0-pr-2410-4cf5-4cf5c3a6.zip


ivan-kalachikov:
**Blocker: the Add-block live preview is discarded by the new protocol guard**

`isBuilderMessage` requires `model` to pass `isPageContent`, which demands a non-empty `id` (`builder-preview.protocol.ts:34`, `:112`). The Designer's preview payload has no id — `generatePreviewBySchema()` returns `{ ...schema.preview || schema.default, ...settings, type }` (vc-module-pagebuilder `editor.helpers.ts:104`), and `previewItem$` sends exactly that object as `model` (`ui/effects.ts:162`).

Clicking an item in **Add block** is what triggers this message (`add-section-item.component.html`, `(click)="raiseOnPreview()"`). The guard now rejects it, so the entire message is dropped — `template` included — and the preview keeps rendering the previous page, leaving only a `Logger.warn`. On `dev` it worked: `updatePreview` appended the model and rendered it.

Ran the guard from this branch against the real theme schemas (`sections/text.json` + `shared/title.json`) through the Designer's own model-building path:

```
model the Designer sends: {"text":"Hello, markdown!","heading":"h2","type":"text"}
model has an id?          false
PR 2410 accepts it:       false
…with an id added:        true
```

No id can appear on that path: none of the 24 section/block schemas define one in `preview`/`default`, no setting is keyed `id`, `prepareSchema()` doesn't add one, and `previewMessage` is merged at message level (the only one in the theme is `{"templateKey":"product"}`). `type` *is* present, so the feature worked before this change.

Suggested fix: validate `model` as a preview model (non-empty `type`, `id` optional) and assign a temporary id in `updatePreview`. Worth adding a protocol test with a `model` payload — none of the new cases carry one, which is why this passed CI.

client-app/plugins/builder-preview/builder-preview.protocol.ts
```diff
@@ -0,0 +1,140 @@
...
+    case "hover":
+      return value.sectionId === null || isNonEmptyString(value.sectionId);
+    case "select":
+      return isNonEmptyString(value.sectionId);
+    case "navigate":
+      return typeof value.url === "string";
+    case "settings":
+      return isRecord(value.settings);
+    case "auth":
+      return isAuthToken(value.token) && (value.userId === null || typeof value.userId === "string");
```
vas11yev1work: `userId` is optional in `TransferDataType` (`Partial<NullableBuilderMessageFieldsType>`)
and the handler already copes with it being absent — `setGlobals({ userId: data.userId || originalUserId })`.
But this guard requires the field to be present, so `{ type: "auth", token: { access_token } }`
without `userId` is rejected and the whole `auth` branch is skipped: `previewToken` stays
`undefined`, `Authorization` is never overridden in `modifyRequests`, and blocks are not
remounted — the preview keeps rendering as the previous user. Before this PR the message
was applied.

```suggestion
      return isAuthToken(value.token) && (value.userId === null || isOptionalString(value.userId));

File: client-app/plugins/builder-preview/shared-component-overlay.ts [hunk 2]
```diff
@@ -0,0 +1,349 @@
...
+    this.scheduleRender();
+  }
+
+  highlight(placementId: string | null): void {
+    this.highlightedPlacementId = placementId;
+    this.applyHighlight();
+  }
+
+  scrollToPlacement(placementId: string): boolean {
+    this.pendingScrollPlacementId = placementId;
```
vas11yev1work: `pendingScrollPlacementId` is assigned before the existence check, so a rejected call still
leaves it set. The caller reads `false` as "not mine" and falls back to `scrollToSection`,
but the id stays pending and `flushPendingScroll` re-runs on every `render()` — including
plain resizes and `ResizeObserver` height changes. Once a boundary with that id shows up,
the preview scrolls on its own.

ToxaKZ:
Fixed

client-app/plugins/builder-preview/builder-preview.protocol.ts
ToxaKZ: Fixed

client-app/plugins/builder-preview/shared-component-overlay.ts
ToxaKZ: Fixed

vas11yev1work (approved):

ivan-kalachikov (approved):

PR #2395, opened 2026-07-21
feat(VCST-5485): sales-rep hub statistics & data

ivan-kalachikov:
## Description

Adds the statistics and data layer for the Sales Rep hub, stacked on top of the dashboard branch (#2388).

**Data layer**
- **KPI statistics**: order and cart statistics composables + GraphQL queries feeding the dashboard and customer-profile widgets (counts, totals, averages, and period-over-period comparisons).
- **Customer counts**: `salesRepCustomerCounts` (assigned / ordering / new customers with month-over-month comparison).
- **Top sellers**: `top-sellers` widget with its query and composable.
- **Filter & sort rules**: rule-discovery queries (customer/order/top-seller) surfaced via `useSalesRepRules` and rendered as `sales-rep-rule-chips`; header-click sorting is bridged to the backend's named sort rules via `useSalesRepColumnSort`, with the asc/desc toggle driven by the backend `supportsDirection` flag.
- **Period filtering**: `useSalesRepPeriodFilter` and the shared `buildStatisticsWindows` date-window helper.
- Regenerated GraphQL `types.ts`, locale strings across all 13 languages, and shared helpers in `utils.ts`.

**Behavior & correctness**
- Previous-period comparison windows are clamped to the current period start, so a longer current-period-to-date span can't overlap the previous window and skew the month-over-month / year-over-year deltas at month-end and leap-year boundaries. Covered by `utils.test.ts`.
- Filter-aware empty states: the orders, top-sellers, and my-customers views distinguish "nothing matches this filter/search" from "no data at all".
- Header-click sorting works on both the dashboard and the customer-profile Orders widget (sortability is driven by backend rule availability, not the filter-chips prop).
- My customers: changing the filter or sort resets to page 1 in a single query.
- Customer-profile value cards use their own labels (`active_cart` / `purchased_mtd`) rather than reusing the dashboard's count labels, since they show monetary totals.

**Cleanup**
- Removed unused GraphQL selections (`itemsQuantity`, `lastOrderDate`, `sinceDate`, `lastCartDate`, `accountId`) and the dead `salesRepCartFilterRules` query; pruned dead locale keys.
- Table header alignment uses VcTable's defaults instead of reaching into UI-kit internals.
- Trimmed comments across the module to the minimum useful set.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5485
### Artifact URL:
[...]

PR #2414, opened 2026-07-31
fix(VCST-5592): Using local time instead of UTC for sales rep hub statistics filters

yuskithedeveloper:
## Description
The sales-rep statistics widgets on the Dashboard and Customer page built their today/week/MTD/YTD windows from UTC calendar parts, so the day boundary didn't match the user's own day: an order placed at 23:00 UTC on Jul 27 (Jul 28 02:00 for a UTC+3 rep) appeared in the recent-orders list and in the unbounded "new orders" counter, but was missing from the "created today" badge. 
`buildStatisticsWindows()` — the single place periods are computed, feeding both surfaces and the orders/top-sellers period chips — now anchors every boundary to the user's local midnight while still sending plain UTC instants to the API, so the widgets agree with the dates rendered beside them; no backend change was needed.

**Note for testing:** order numbers are still minted from the UTC date (`CO{0:yyMMdd}-…`), so a near-midnight order can be numbered `…0727…` while the UI consistently shows Jul 28 — that mismatch is expected, not a regression.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5592
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.56.0-pr-2414-332d-332d12cb.zip


client-app/modules/sales-rep/utils.ts
```diff
@@ -44,40 +49,46 @@ export type StatisticsWindowsType = {
...
+  // Monday-start week: getDay() is 0 (Sun)…6 (Sat); shift so Monday === 0.
+  const daysSinceMonday = (now.getDay() + 6) % 7;
+  const weekStart = local(year, month, day - daysSinceMonday);
+  // Calendar arithmetic rather than weekStart − 7×24h: a DST transition inside the week makes the elapsed
+  // millisecond span differ from seven days, which would land the bound an hour off midnight.
+  const prevWeekStart = local(year, month, day - daysSinceMonday - 7);
+
   // Clamped to the previous period's own end, so a longer current elapsed span (e.g. 30 days into
   // March vs a 28-day Feb) can't spill past it and double count.
   const matched = (prevStart: number, currentStart: number): string =>
```
ivan-kalachikov: `matched()` still adds raw elapsed milliseconds, which drift across a DST change. You already fixed exactly this for `prevWeekStart` five lines up — the same calendar move works here.

Measured over 2025-26: ~100 days/yr land wrong in any DST zone, about half on the wrong calendar day. `TZ=America/New_York`, Wed Apr 9 2025 gives `prevTo = Mon Mar 10 00:59:59` where Mar 9 23:59:59 is intended. UTC is unaffected, which is why the tests stay green.

Impact is small (<=1h inside a month-long baseline), so not a blocker. Fix for the month bound:

```ts
Math.min(local(year, month - 1, day, 23, 59, 59, 999), monthStart)
```

client-app/modules/sales-rep/utils.test.ts
```diff
@@ -1,65 +1,113 @@
...
-    const now = new Date(Date.UTC(2025, 2, 12, 9, 30, 0)); // a Wednesday
+    const now = new Date(2025, 2, 12, 9, 30, 0); // a Wednesday
     const w = buildStatisticsWindows(now);
 
+    // Monday-start week, and the previous week starts exactly seven calendar days earlier.
+    expect(w.weekFrom).toBe(localIso(2025, 2, 10));
+    expect(w.prevWeekFrom).toBe(localIso(2025, 2, 3));
     // The elapsed part of a week can't exceed a full previous week, so the previous-week end
     // always stays strictly before the current week start.
     expect(new Date(w.prevWeekTo).getTime()).toBeLessThan(new Date(w.weekFrom).getTime());
```
ivan-kalachikov, line 70: This invariant no longer holds — the weekly clamp does engage on a fall-back Sunday: `TZ=America/New_York`, now = Sun Nov 2 2025 gives `prevWeekTo === weekFrom` (both `2025-10-27T04:00:00.000Z`).

The test passes only because its fixed date (Wed Mar 12) dodges the case, and the runtime behavior is fine — the clamp does its job. Suggest `toBeLessThanOrEqual` and dropping "never clamps" from the name.

ivan-kalachikov (commented):
Direction is right: it matches what `core/utilities/date` already does for the order-history filter, keeps the day-rounding cache rationale, and closes the UTC-boundary issue that #2395's review had already flagged.

One fix worth making, inline. Premise checks out — all windows feed GraphQL `DateTime` instants with no date-only field, and the adjacent surfaces really do render `createdDate` in the browser zone. Suite passes 7/7 across 12 zones.

Heads-up: #2412 touches the same three statistics composables, so re-run the freshness tests if it merges second.

ivan-kalachikov (dismissed):

ivan-kalachikov (approved):

PR #2416, opened 2026-07-31
feat(VCST-5588): sales rep hub active carts

yuskithedeveloper:
## Description

Storefront half of VCST-5588: the **Active carts** stat card now shows **item quantities** instead of a cart count + money total, and the Hub dashboard and the customer page render it from **one shared source** instead of two hand-written copies.

Requires the backend PR (new `selectedItemQuantity` / `unselectedItemQuantity` fields): https://github.com/VirtoCommerce/vc-module-sales-rep/pull/10

**Before → after**

| slot | before | after |
|---|---|---|
| value | `3` (number of carts) | `3` **items** (quantity selected for checkout) |
| sub-figure | `$0.00` (cart total) | `12 not for checkout` (quantity *not* selected) |
| delta | `2 new this week` (carts created this week) | `3 items this week` (**selected** items whose line was touched this week) |
| customer page | its own card: the cart's money total, no sub, no delta | the same card, scoped to that organization |

### What the three figures mean, precisely

* **Value** — summed `Quantity` of the cart lines with `SelectedForCheckout = true`, across the carts in scope.
* **Sub** — the same sum for `SelectedForCheckout = false` (the parked remainder).
* **Delta** — the **value's own metric, windowed**: still only *selected* lines, restricted to those whose `ModifiedDate` falls in the current week. It is **not** "all items touched this week" — an unselected line touched this week is counted nowhere in this card's delta. The selected flag is read as of *now*, not as of the window, so a line selected on Monday and deselected on Thursday does not appear.
* **Scope** — the dashboard aggregates every organization the rep serves; the customer page passes `organizationId` and the backend narrows to it. Same builder, different payloads; sharing code does **not** make the two surfaces show the same numbers.

### Behaviour

*   **One card definition, two surfaces.** New `useSalesRepActiveCartsCard(organizationId?)` owns the card *and* its data source; `useSalesRepDashboardWidgets` and `useSalesRepCustomerWidgets` each drop it in. This is the ticket's "single shared calculation source — no page-specific logic" on the client side, and it removed the duplicated `useSalesRepCartStatistics` wiring. Card position on both pages is unchanged.
*   **Customer page scoping is data-only** — the composable takes the organization and the backend authorizes + narrows to it; verified against three of the rep's organizations (the one with a cart returns its quantities, the others return zero).
*   **`<StatWidget>` gained an optional `valueSuffix`** — a de-emphasized unit rendered after the value ("3 **items**") so the number stays the focal point. Generic on purpose: the same widget renders orders and customer counts.
*   **i18n:** added `items_unit`, `not_for_checkout`, `items_this_week`; removed the now-unused `new_this_week` and the singular `active_cart` label. All 13 locales, each matching that locale's existing checkout wording (de "nicht zur Kasse", fr "hors paiement", ru "не к оформлению", zh "件不结账", …). `items_unit` / `items_this_week` use vue-i18n plural forms (`"item | items"`, house style per `locales/*.json` `products_found_message`), so a count of 1 reads "1 item", not "1 items"; single form for ja/zh/de.
*   **Query trimmed** — with both cards reading only quantities, `count` and `total` are no longer selected, which also drops the storefront's last dependency on the persisted `Cart.Total` (unreliable for carts built through short-cart mutations).
*   **`types.ts` is hand-updated** to match the new schema — `yarn generate:graphql-types` needs a running backend carrying the new module build, so it could not be run here. A codegen run after the backend PR lands should be a no-op; the schema-level `CustomerCartStatisticsPeriod` / `…Comparison` types were updated in both `core` and the module for that reason.

### Breaking changes

* **Card meaning changed** on both surfaces: the customer page tile was a money total and is now a quantity. Layout **block ids are unchanged** — dashboard `active_carts`, profile `active_cart` (kept singular deliberately so saved customer-profile layouts keep matching); only the profile's caption changed to "Active carts".

### Screenshots
image

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5588
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.56.0-pr-2416-efad-efadb479.zip


client-app/modules/sales-rep/locales/ru.json
```diff
@@ -66,7 +65,9 @@
           "ordered_this_month": "{count} заказали в этом месяце",
           "new_customers": "{count} новых клиентов",
           "placed_today": "{count} сегодня",
-          "new_this_week": "{count} новых за неделю",
+          "items_unit": "товар | товаров",
+          "not_for_checkout": "{count} не к оформлению",
+          "items_this_week": "{count} товар за неделю | {count} товаров за неделю",
```
vas11yev1work: Add these two forms (one/few/many) with pluralRules registered for ru/pl, since vue-i18n's default rule only handles singular/plural and picks the wrong form for counts like 2-4 in Slavic languages.

```
"items_unit": "товар | товара | товаров",
"items_this_week": "{count} товар за неделю | {count} товара за неделю | {count} товаров за неделю"```

File: client-app/modules/sales-rep/composables/useSalesRepCustomerWidgets.ts [hunk 2]
```diff
@@ -39,8 +39,11 @@ export function useSalesRepCustomerWidgets(organizationId: MaybeRefOrGetter<stri
     // Plain green count (no chevron) — New-status orders placed today, for this customer. Formatted
     // string renders; the raw number is vue-i18n's plural selector (see the dashboard mapper).
     const placedToday = orders?.newOrdersToday?.count ?? 0;
-    // Active cart: the summed total of this customer's non-empty carts.
-    const activeCarts = carts?.activeCarts;
+    // Same metric as the dashboard "Active carts" card, scoped to this customer's organization: quantities
+    // of the cart lines picked for checkout, the parked remainder, and the lines touched this week (VCST-5588).
+    const selectedItems = carts?.activeCarts?.selectedItemQuantity ?? 0;
+    const unselectedItems = carts?.activeCarts?.unselectedItemQuantity ?? 0;
+    const weekItems = carts?.itemsThisWeek?.selectedItemQuantity ?? 0;
```
vas11yev1work: Duplicated logic between the two composables (`useSalesRepDashboardWidgets.ts` and `useSalesRepCustomerWidgets.ts`) — same `selectedItems`/`unselectedItems`/`weekItems` derivation and card-building code copy-pasted in both files instead of shared. Nothing enforces they stay in sync going forward.

Worth pulling into one shared helper both composables call — but open to leaving it as-is if you'd rather not add an abstraction for two call sites. What do you think?

ivan-kalachikov:
**Caption vs figure.** The card is captioned "Active carts", but every figure on it is now an item quantity — value, sub and delta. No cart count or cart value is left anywhere on the card.

**It can also disagree with the cart definition.** The item aggregation runs with `OnlyNonEmpty = false` (backend test `Cart_ItemQuantities_ReadTheLineItems_NotTheDenormalizedCount`: `count` 0, `selectedItemQuantity` 8), so "Active carts" can report items for a customer with no active carts.

VCST-5588's AC asks that the label match what is actually measured — worth a decision on whether the caption stays cart-worded.

client-app/modules/sales-rep/composables/useSalesRepCustomerWidgets.ts
yuskithedeveloper: We are trying to unify the appearence and logic in the global and per-customer metrics, so it absolutely makes sense to extract duplicated code into shared helpers. Done.

vas11yev1work (dismissed):

ivan-kalachikov (approved):

PR #2418, opened 2026-08-03
feat(VCST-4855): create a separate component for saved for later

vas11yev1work:
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-4855
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.55.0-pr-2418-cf47-cf47c1fa.zip


goldenmaya (dismissed):

client-app/pages/account/saved-for-later-details.vue
```diff
@@ -67,7 +193,193 @@ async function moveLineItemsToCart(lineItemIds: string[]): Promise<void> {
...
+}
+
+onBeforeRouteLeave(canChangeRoute);
+onBeforeRouteUpdate(canChangeRoute);
+
 onMounted(() => {
   void getSavedForLater();
 });
+
+watchEffect(() => {
```
ivan-kalachikov: **Blocking.** This effect dropped the `await` the original had, which changes what it tracks. Vue only registers dependencies read *before* the first `await` — in `list-details.vue` the read came after `await fetchWishList(props.listId)`, so `props.listId` was the only dep. Here the read is synchronous, so `savedForLaterList.value?.items` is now a real dependency.

`savedForLaterList` is a `createSharedComposable` singleton reassigned by `getSavedForLater()`, `updateItems()` and the delete modal's `onResult`, so this re-runs on every list change:

- unsaved quantity edits are overwritten from the server — and `isDirty` flips false too, so the user never gets the save prompt;
- the user is bounced back to page 1;
- `addAllListItemsToCart` (L201-212) awaits a move that empties the list, so by the time the continuation runs `savedForLaterItems` is `[]` — `trackAddItemsToCart` is skipped and the results modal comes up empty. I checked that ordering against the repo's own Vue: the pre-flush job wins, the continuation sees `[]`.

The second effect at L375 picks up the same new dependency, so GA `viewItemList` also refires on every move/delete/save.

client-app/pages/account/saved-for-later-details.vue
```diff
@@ -67,7 +193,193 @@ async function moveLineItemsToCart(lineItemIds: string[]): Promise<void> {
...
+async function openSaveChangesModal(): Promise<boolean> {
+  return await new Promise<boolean>((resolve) => {
+    const closeModal = openModal({
+      component: SaveChangesModal,
+      props: {
+        title: t("pages.account.list_details.save_changes"),
+        message: t("pages.account.list_details.save_changes_message"),
+        onConfirm: async () => {
+          closeModal();
+          await updateItems();
```
ivan-kalachikov, line 252: **Blocking.** `updateItems()` calls the raw `updateWishlistItems` API, which has no try/catch — and `client.ts` sets no `errorPolicy`, so Apollo's default `'none'` rejects on GraphQL errors. The original saved through `useWishlists.updateItemsInWishlist`, which catches internally, so `resolve(true)` always ran.

Because `closeModal()` runs first, a failed save leaves no modal, no error message, and a promise that never settles — `canChangeRoute` then blocks navigation off the page permanently, and `buyNow` silently aborts. There's an unhandled rejection in the console too.

Simplest fix is local: try/catch around `updateItems()` and settle the promise in a `finally`, so the route guard can never hang regardless of how the save goes.

_Edited: I originally suggested switching to `useWishlists.updateItemsInWishlist`. Withdrawing that — its refetch is `fetchWishList` → `getWishList`, which populates the wishlists composable's `list`, not the `savedForLaterList` this page actually renders, so it would pull in unrelated state without refreshing anything here. If you do want a post-save refetch, `getSavedForLater()` is the right one._

client-app/pages/account/saved-for-later-details.vue
```diff
@@ -67,7 +193,193 @@ async function moveLineItemsToCart(lineItemIds: string[]): Promise<void> {
...
+      void pushHistoricalEvent({ eventType: "addToCart", productId: lineItem.product.id });
+    }
+  } finally {
+    pendingItems.value[lineItem.id] = false;
+  }
+
+  showResultModal([lineItem]);
+}
+
+function openDeleteProductModal(values: string[]): void {
```
ivan-kalachikov, line 310: This drops the `previousPagesCount` step-back logic that `list-details.vue` still has (L347-359), which moves the user back a page when a deletion empties the current one.

Worth fixing together with the `watchEffect` above — restoring the original untracked effect without also porting this `onResult` shape would leave deleted rows on screen, since `onResult` only updates `savedForLaterList`.

client-app/pages/account/saved-for-later-details.vue
```diff
@@ -31,27 +97,87 @@
...
+);
+const isDirty = computed<boolean>(() => {
+  const originalItemsToCompare = (savedForLaterList.value?.items ?? []).map((item) =>
+    pick(item, ["productId", "quantity"]),
+  );
+  const changedItemsToCompare = (savedForLaterItems.value ?? []).map((item) => pick(item, ["productId", "quantity"]));
+  return !isEqual(originalItemsToCompare, changedItemsToCompare);
+});
+const savedForLaterListProperties = computed(() => ({
+  item_list_id: "wishlist",
```
ivan-kalachikov: Copied verbatim from the wishlist page (`item_list_id`, `item_list_name`, `related_type` all say wishlist), so saved-for-later events will be indistinguishable from wishlist ones in GA.

ivan-kalachikov (commented):
Reviewed the fork line-by-line against the merge-base `list-details.vue`. The direction matches what was agreed on #2294, and the `list-details.vue` cleanup is clean — no remaining callers of `hideSettings` / `kind`.

Two issues look blocking, both artifacts of the copy rather than the design. The rest are minor.

One non-defect thought: ~7 behaviour/bug-fix commits landed on `list-details.vue` in the last 18 months, so the fork doubles the surface those future fixes have to be applied to.

client-app/core/api/graphql/cart/fragments/savedForLaterLineItem.graphql
```diff
@@ -33,6 +36,12 @@ fragment savedForLaterLineItem on LineItemType {
       id
       name
     }
+    availabilityData {
```
ivan-kalachikov, line 39: These fields are needed — `prepareLineItem` reads both, so the old page was missing them.

One thing worth checking: this fragment also feeds `getSavedForLater`, which `cart.vue` calls for the compact saved-for-later widget. `cart-for-later.vue` / `cart-item-for-later.vue` read neither `availabilityData` nor `properties` (they don't go through `prepareLineItem`), so that widget now fetches both per line item for nothing. If it's worth avoiding, a slimmer fragment for the cart query would do it.

_(An earlier version of this comment also suggested reusing `wishlistLineItemFields.graphql` since the two bodies are currently identical — withdrawing that. They're separate domains and should stay free to diverge; coupling them on incidental shape would mean a wishlist field change silently altering the saved-for-later payload.)_

vas11yev1work: Good catch, and confirmed — `prepareLineItem` does need `availabilityData`/`properties` for the details page, but `cart-item-for-later.vue` uses neither, so the widget over-fetches on every cart page load.

Both options (an `@include` variable threaded through the query + both mutations, or a separate lean query) end up costing more than they save. The mutations that update the shared `savedForLaterList` singleton are also called from `cart.vue` (move to/from cart), so a leaner fetch either needs its own ref that goes out of sync after those mutations, or loosens the shared type to make the heavy fields optional everywhere — for a handful of extra fields on a list capped at 6 visible items.

ivan-kalachikov (approved):

PR #2421, opened 2026-08-04
fix(VCST-5533): coupons sidebar — WCAG 2.2 AA gaps

goldenmaya:
### Jira-link:

https://virtocommerce.atlassian.net/browse/VCST-5533
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.55.0-pr-2421-1e94-1e949ce0.zip

---

> [!NOTE]
> **Medium Risk**
> Changes cart coupon apply/remove semantics and shared `VcInput` ARIA merging, which can affect checkout UX and any consumer passing custom `aria-describedby`; behavior is well covered by new tests but touches core cart flows.
> 
> **Overview**
> Addresses **VCST-5533** by fixing WCAG 2.2 AA gaps in the cart coupons sidebar and tightening coupon apply/remove behavior.
> 
> **Coupon cards** now expose proper names and errors: visible labels get stable `id`s wired via `aria-labelledby` / `aria-label`, rejected codes set `aria-invalid` and `aria-describedby` on the input, and errors keep `role="alert"`. **`VcInput`** merges consumer `aria-describedby` with its internal details id instead of overwriting it.
> 
> **Coupons section** adds a screen-reader-only `aria-live="polite"` region that announces apply/remove (including swap and partial-failure cases), discount, and new total when cart figures exist. Apply/remove go through handlers that rely on **boolean results** from `useCoupon` so announcements are not skewed by shared module-level `couponError`.
> 
> **Coupon logic** introduces `isSameCouponCode` (trim + case-insensitive) for loading, applied/error UI, and swaps; re-applying the same code in a different case no longer removes the working coupon. **`applyCoupon` / `removeCoupon`** return whether the cart actually changed after mutations, with extra validation when add/remove resolves but the cart does not reflect the code.
> 
> New i18n strings for announcements and fallback input labeling; broad unit/integration tests cover a11y wiring, live region behavior, and edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e949ce0c33f77bf56cc90942feb87df5003db92. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


client-app/shared/cart/composables/useCoupon.ts
```diff
@@ -36,31 +38,39 @@ export function useCoupon() {
       const isValid = await validateCartCoupon(trimmed);
       if (!isValid) {
         couponError.value = { code: trimmed, type: "invalid" };
-        return;
+        return false;
       }
 
       await addCartCoupon(trimmed);
+
+      return true;
```
ivan-kalachikov, line 69: **Can report success when the coupon was not applied.**

`true` here means "the mutation resolved", not "the coupon applied". The UI's own truth is `appliedCouponCode` (line 13) = `coupons[].isAppliedSuccessfully`.

The pre-flight `validateCartCoupon` does run promotion evaluation, so this isn't the plain bad-code case. But it is not the same test:

- `ValidateCouponQueryHandler` clones the cart and **wipes the coupon list** first (`Cart.Coupons = new[] { request.Coupon }`), so it checks the code *alone*. The real `addCoupon` evaluates alongside whatever else is already in `Cart.Coupons`.
- It accepts any reward with `IsValid`. `isAppliedSuccessfully` needs an actual `Discount` on the cart (`CartAggregate.cs:121-133`). A valid reward that yields no `Discount` -- zero amount, gift reward, shipping reward with no shipment -- passes validation but is not applied.
- Two round-trips, so the cart can move in between.

When it happens the card stays in `default` view (no error, no green border) while a screen reader hears "Coupon X applied. Discount ..., new total ...". The announcement contradicts the screen.

The answer is already loaded -- the mutation returns the full cart including `coupons`:

```suggestion
      await addCartCoupon(trimmed);

      if (appliedCouponCode.value?.toLowerCase() !== trimmed.toLowerCase()) {
        couponError.value = { code: trimmed, type: "invalid" };

        return false;
      }

      return true;
```

Case-insensitive because the backend matches with `EqualsIgnoreCase`. Bonus: also fixes the existing hole where these coupons fail with no feedback at all.

client-app/shared/cart/composables/useCoupon.ts
```diff
@@ -36,31 +38,39 @@ export function useCoupon() {
...
-      return;
+      return false;
     }
 
     try {
       loadingCouponCode.value = trimmed;
 
       await removeCartCoupon(trimmed);
+
+      return true;
```
ivan-kalachikov: **Same unconditional `true` on the remove side.**

Backend does `Cart.Coupons.Remove(FirstOrDefault(...))` -- a silent no-op when the code isn't there, so this still returns `true` and the removal gets announced.

Low reach, since Remove only renders in `applied` view. But the symmetric guard (`appliedCouponCode` no longer equals the code) closes it for free.

client-app/shared/cart/components/coupons-section.vue
```diff
@@ -57,8 +59,38 @@ const isMarketingExperienceApiEnabled = computed(() => hasModule(MODULE_ID_MARKE
...
+
+async function handleApply(code: string) {
+  announcement.value = "";
+
+  // Failures are already announced by the card's alert.
+  if (!(await applyCoupon(code))) {
+    return;
+  }
+
+  announcement.value = t("shared.cart.coupons_section.applied_announcement", {
```
ivan-kalachikov: **Swapping coupons never announces the removal.**

`useCoupon.ts:34-35` silently drops the applied coupon when a different one is applied. A screen reader hears only "Coupon B applied" -- that A is gone is never said, though A's card visibly flips back to Apply.

Sighted users get no explicit feedback either, so it's a parity question more than a regression. But under the same 4.1.3 rationale as finding #1, losing a coupon is a status change.

client-app/shared/cart/components/coupons-section.vue
```diff
@@ -57,8 +59,38 @@ const isMarketingExperienceApiEnabled = computed(() => hasModule(MODULE_ID_MARKE
...
+    code,
+    discount: cart.value?.discountTotal?.formattedAmount ?? "",
```
ivan-kalachikov: **The `?? ""` fallbacks make a broken sentence.**

With no cart totals this reads out as "Coupon ZUR10 applied. Discount , new total ."

And `coupons-section.test.ts` ("does not break when the cart has no totals yet") sets `cart.value = undefined` and asserts only `toContain("ZUR10")` -- so the test currently locks that output in.

Unlikely in practice (the full cart is loaded before you can apply anything), but a separate figure-less string would read better than empty slots.

client-app/shared/cart/components/coupons-section.vue
```diff
@@ -57,8 +59,38 @@ const isMarketingExperienceApiEnabled = computed(() => hasModule(MODULE_ID_MARKE
...
+}
+
+async function handleRemove(code: string) {
+  announcement.value = "";
+
+  if (!(await removeCoupon(code))) {
+    return;
+  }
+
+  announcement.value = t("shared.cart.coupons_section.removed_announcement", {
```
ivan-kalachikov: **Removal doesn't announce the discount change.**

Finding #1 is about Discount **and** Total. Apply announces both (lines 75-78); this announces total only, and `removed_announcement` has no `{discount}` placeholder in any of the 13 locales.

Removing a coupon changes the Discount row too, so assistive tech still misses half of what finding #1 asked for. Needs the placeholder added to all locales plus a test.

ivan-kalachikov (commented):
Findings #2 and #3 look cleanly done. Five notes on the announcement path, inline.

Checked and fine, so nobody re-litigates these: cart figures are **not** stale (both mutations return `...fullCart` with `skipQuery: false`, Apollo broadcasts before `mutate()` resolves); no `aria`-prop regression for the other `VcInput` consumers (and `vc-select.vue:78`'s own `detailsId` was being silently dropped before, so the merge fixes that too); `useId()` is right and there is no SSR here; all 13 locales properly translated. Types clean, lint clean, suite 1453/1454 (one unrelated flake). Mutation-tested the tests with three deliberate reverts: 10 failures, so they do their job.

ivan-kalachikov (approved):

PR #2423, opened 2026-08-05
fix(VCST-5555): a11y issues

goldenmaya:
### Jira-link:

https://virtocommerce.atlassian.net/browse/VCST-5555
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.55.0-pr-2423-5c87-5c876db5.zip

---

> [!NOTE]
> **Low Risk**
> Changes are localized to UI-kit accessibility patterns, Storybook examples, theme tokens, and deployment package pins; popover trigger API now expects `triggerProps` on the consumer element for `#trigger` slots.
> 
> **Overview**
> Addresses **VCST-5555** with UI-kit accessibility fixes and small visual/token updates; **`backend-packages.json`** is bumped for platform **3.1058.0**, Azure blob prerelease Cart/XCart/SalesRep packages, and refreshed GitHub module versions.
> 
> **Popover** no longer puts `role="button"` and full trigger bindings on the wrapper for the `#trigger` slot—listeners stay on the wrapper while consumers bind **role-free** `triggerProps` on their own control (fixes nested-interactive). Enter-to-open on keyup was removed so native buttons don’t reopen after click-close.
> 
> **Scrollbar** adds conditional keyboard focus when content overflows and there are no focusable descendants or interactive container roles, with resize/mutation/load observers to keep that in sync; **focus-visible** outline styling was added.
> 
> **Table** single-selection mode gets a screen-reader label for the selection column via new **`selection_column`** locale strings. **Input** clear control gets an **`aria-label`**. **VcActionInput** delegates labeling to **VcInput** instead of a separate `<p>` label.
> 
> Theme **red** preset adds solid accent background/border tokens; loading container opacity and some **neutral-600** text tweaks improve contrast in stories and loading states.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c876db5fc4a1a0ef3e3ae5cf6946dba7340cc98. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vas11yev1work (dismissed):

client-app/ui-kit/components/atoms/scrollbar/vc-scrollbar.vue
```diff
@@ -62,6 +62,36 @@ const el = useTemplateRef<HTMLElement>("el");
...
+  hasOverflow.value = verticalOverflow || horizontalOverflow;
+}
+
+onMounted(() => {
+  void nextTick(updateOverflow);
+});
+
+onUpdated(updateOverflow);
+
+useResizeObserver(el, updateOverflow);
```
ivan-kalachikov: These four triggers miss the main case: content growing inside a **child component**.

`onUpdated` only fires when `VcScrollbar` itself re-renders, which doesn't happen across a component boundary — and the `ResizeObserver` watches *this* element's box, which stays fixed while `scrollHeight` grows.

Real example: `mobile-search-bar.vue:43` wraps `<SearchDropdown/>`. Results load async → content overflows → nothing fires → no `tabindex`. Same for images loading late.

So the axe violation this is fixing still shows up in that state. Maybe observe an inner content wrapper (or add a `MutationObserver`) instead of only `el`?

client-app/ui-kit/components/atoms/scrollbar/vc-scrollbar.vue
```diff
@@ -62,6 +62,36 @@ const el = useTemplateRef<HTMLElement>("el");
...
 provide(vcScrollbarKey, { el });
 
+// A scrollable region must be keyboard-reachable (axe: scrollable-region-focusable).
+// The tab stop is added automatically when content actually overflows on an enabled axis;
+// the `focusable` prop remains as an explicit override.
+const hasOverflow = ref(false);
+
+const isFocusable = computed(
+  () => props.focusable || ((props.vertical || props.horizontal) && !props.disabled && hasOverflow.value),
+);
```
ivan-kalachikov: This makes `VcSelect`'s option list tabbable, which breaks the combobox pattern.

`VcDropdownMenu` renders the list **as** the scrollbar (`vc-dropdown-menu.vue:26`, `tag="ul"` + `:role="listRole"`), and `VcSelect` passes `list-role="listbox"` while keeping focus on the trigger via `aria-activedescendant`. A listbox driven that way must not be a tab stop. Long list overflows → Tab lands on the `<ul>` → arrow keys do nothing there (handlers live on the options).

Also worth knowing: axe's `scrollable-region-focusable` **passes** when the region contains focusable content — which every current call site does. So the blanket tabindex mostly adds stops without clearing violations.

Could this stay opt-in via `focusable`, or at least skip containers that already carry an interactive `role`?

client-app/ui-kit/components/molecules/popover/vc-popover.vue
```diff
@@ -4,7 +4,9 @@
       <slot :open="open" :close="close" :toggle="toggle" :opened="opened" :trigger-props="emitTriggerProps" />
     </div>
 
-    <div v-else-if="$slots.trigger" ref="reference" class="vc-popover__trigger" v-bind="emitTriggerProps">
+    <!-- Only event listeners are bound here: the slot may contain a real interactive element,
+         and an interactive role on the wrapper would nest interactive controls (axe: nested-interactive). -->
+    <div v-else-if="$slots.trigger" ref="reference" class="vc-popover__trigger" v-on="triggerListeners">
```
ivan-kalachikov, line 10: Dropping the `role`/aria here is right for the current in-repo users — every one of them reaches this path through a hover `VcTooltip`, where `role="button"` + `aria-haspopup="dialog"` was wrong anyway, and it fixes real nested-interactive cases (`dark-mode-toggle.vue:4` has a real `<button>` inside; `vc-checkbox.vue:37` sits inside a `<label>`). Nothing lost on keyboard either — the wrapper never had `tabindex`.

One gap though: the slot only exposes `open/close/toggle/opened` (line 10), no `trigger-props`. So a consumer on this path can't add `aria-expanded`/`aria-haspopup` back at all without rewriting to the default slot.

That matters because the stories still document this path with click popovers — `vc-popover.stories.ts:66` (Basic) and `:171` both do `#trigger` + `<VcButton>`, and those now announce no popup and no state.

Could you expose aria-only `trigger-props` on this slot (no `role`), or move those stories to the default-slot pattern?

client-app/ui-kit/components/molecules/popover/vc-popover.vue
```diff
@@ -100,17 +102,13 @@ const { placement, strategy, flipOptions, offsetOptions, shiftOptions } = toRefs
...
-  onFocusout: props.hover ? close : undefined,
-  onClick: props.hover ? undefined : toggle,
-  onKeyup: (e: KeyboardEvent) => {
+const triggerListeners = computed(() => ({
+  mouseenter: props.hover ? open : undefined,
+  mouseleave: props.hover ? close : undefined,
+  focusin: props.hover ? open : undefined,
+  focusout: props.hover ? close : undefined,
+  click: props.hover ? undefined : toggle,
+  keyup: (e: KeyboardEvent) => {
```
ivan-kalachikov, line 121: Pre-existing, but flagging it since this block is being rewritten anyway: the Enter branch makes it impossible to **close** a popover with Enter.

On a focused button the order is `keydown` → browser fires `click` → `toggle()` closes → `keyup` → `open()` reopens. Space is fine (click fires after keyup) and Escape is fine — it's Enter specifically.

And it isn't only this slot: `onKeyup` is in `emitTriggerProps` too, so it hits every default-slot consumer — currency-selector, language-selector, mega-menu, ship-to-selector, orders-desktop-filters, facet-filter, slider-filter, category-horizontal-filters, product-price-block, the members/address/wishlist dropdown menus, and the push-messages components.

Simplest fix is probably to drop the Enter branch entirely: native buttons and links already activate on Enter via `click`, so it adds nothing and causes the reopen. Escape is the part that's actually needed.

(Traced from the code, not verified in a browser.)

client-app/ui-kit/components/atoms/scrollbar/vc-scrollbar.vue
goldenmaya: Fixed: added a MutationObserver (childList + subtree) alongside the ResizeObserver, plus a recheck on scroll.

client-app/ui-kit/components/atoms/scrollbar/vc-scrollbar.vue
goldenmaya: Fixed: the auto tab stop now skips containers with interactive roles (listbox, combobox, etc.) and regions with focusable descendants; focusable remains as an explicit opt-in.

client-app/ui-kit/components/molecules/popover/vc-popover.vue
goldenmaya: Done: the #trigger slot now exposes role-free trigger-props (aria-haspopup/aria-expanded/aria-controls), and the stories bind them on VcButton.

client-app/ui-kit/components/molecules/popover/vc-popover.vue
goldenmaya: Fixed as suggested: dropped the Enter branch entirely, keyup now handles Escape only.

ivan-kalachikov (dismissed):

client-app/ui-kit/components/atoms/scrollbar/vc-scrollbar.vue
```diff
@@ -62,6 +62,82 @@ const el = useTemplateRef<HTMLElement>("el");
...
+  }
+
+  needsAutoTabStop.value = !target.querySelector(FOCUSABLE_SELECTOR);
+}
+
+const scheduleAutoTabStopUpdate = useDebounceFn(updateAutoTabStop, 100);
+
+onMounted(() => {
+  void nextTick(updateAutoTabStop);
+});
```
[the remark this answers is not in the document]
ivan-kalachikov: Confirmed — there is no watcher on `vertical` / `horizontal` / `disabled`, so a prop-only change never recomputes the tab stop. Checking every call site though, none is currently reachable: `vc-table` passes an explicit reactive `:focusable`, and the search dropdowns always contain focusable descendants — so this is latent rather than a live defect, and not worth expanding this PR for.

Tracked as [VCST-5701](https://virtocommerce.atlassian.net/browse/VCST-5701) (Sprint 26-17) with the per-call-site analysis and the suggested `flush: "post"` watcher plus MutationObserver `attributeFilter`.

ivan-kalachikov (approved):

PR #2424, opened 2026-08-05
feat(VCST-4857): remove deprecated address composables

vas11yev1work:
## Description
- Replaces the deprecated `useOrganizationAddresses`/`useUserAddresses` (which loaded up to 9999 addresses in one shot) with `useCurrentOrganizationAddresses`/`useCustomerAddresses`, backed by real server-side pagination and search.
- Migrates the two remaining consumers: `useShipToLocation.ts` (header/mobile "Ship to" selector) and `edit-quote.vue` (quote shipping/billing address modal).
- Deletes the deprecated composables, their now-unused underlying GraphQL queries (`getMyAddresses`, `getOrganizationAddresses`), and their barrel exports.

image

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-4857
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.56.0-pr-2424-c4ef-c4ef5f57.zip


client-app/shared/ship-to-location/components/ship-to-selector.vue
```diff
@@ -180,34 +194,68 @@ const {
...
+
+watch(filter, applyKeyword);
+
+// filter (the search box) updates instantly; keyword (the server-side filter actually applied to
+// `allAddresses`) only catches up after the debounce above fires and its query resolves. In
+// between - e.g. if isPaginated flips true mid-typing because the org/personal list just loaded -
+// `allAddresses` can reflect a different keyword than what's currently typed. Treat that window as
+// loading rather than rendering a list that doesn't match the search box, so a stale address can't
+// be clicked by mistake.
+const isSearchSyncing = computed(() => isPaginated.value && filter.value !== keyword.value);
```
maksimzinchuk: `filter` and `keyword` have two different representations of "empty", so this comparison can never converge after the clear button is used.

The `VcInput` on L61-67 is `clearable`, and its `clear()` assigns `undefined` (`vc-input.vue:230-234`). The model setter only coerces `""` to `undefined` for `type="number"` (`vc-input.vue:192-199`), and no `type` is passed here, so clearing from the keyboard yields `""` while the ✕ button yields `undefined`. `applyKeyword` then normalizes with `?? ""`, leaving `filter === undefined` and `keyword === ""`.

Repro — PERSONAL account with more than `MAX_ADDRESSES_NUMBER` addresses:

1. search for something with no matches → `totalCount = 0`, the popover stays open thanks to `!!filter` on L228-230
2. click the ✕ in the search box → `filter = undefined`, `hasAnyAddresses` immediately goes false and the whole `VcPopover` (L4) unmounts, replaced by the "add new address" button
3. ~300ms later the unfiltered query resolves, `totalCount > 0`, the popover comes back — but closed
4. reopen it: `undefined !== ""` is still true, so `isListLoading` keeps `VcLoader` (L75) in place permanently instead of the address list

The state does not self-heal; the only way out is typing a character and deleting it with the keyboard. Normalizing both sides — `(filter.value ?? "") !== keyword.value` — fixes step 4.

client-app/shared/ship-to-location/components/ship-to-selector.vue
```diff
@@ -180,34 +194,68 @@ const {
...
 
-  return getFilteredAddresses(filter.value);
+  return getLocalFilteredAddresses(filter.value);
 });
 
 const hasAddresses = computed(() => addresses.value.length > 0);
+
+// Based on the true total, not the current (possibly filtered) view, so an empty search result
+// doesn't hide the whole dropdown behind the "add new address" button.
+const hasAnyAddresses = computed(() =>
```
maksimzinchuk, line 235: `totalCount` is the keyword-filtered server total, so this gate collapses the whole dropdown during every search → clear transition.

`filter` drops to falsy immediately, but `keyword` only catches up after the 300ms debounce plus the round trip. In that window `totalCount` still holds the stale filtered value — 0 after a zero-result search — so `hasAnyAddresses` is false and the `v-if` on L4 unmounts `VcPopover`, closing an open dialog and swapping the trigger for the "add new address" button. The same mechanism makes the search `VcInput` unmount and lose focus when you backspace the last character of a query that had narrowed results to `<= MAX_ADDRESSES_NUMBER`.

Two things would fix it: keep the popover mounted while the search is still syncing or loading, and base "does this user have an address book" on a total that is independent of the active keyword.

client-app/shared/ship-to-location/composables/useShipToLocation.ts
```diff
@@ -100,19 +118,128 @@ export function useShipToLocation() {
...
+  // True when accountAddresses is server-paginated (PERSONAL, or CORPORATE/CORPORATE_LIMITED with
+  // organization address access) vs. sourced from the small local-storage list (ANONYMOUS, or
+  // CORPORATE_LIMITED without organization address access).
+  const isPaginated = computed(() => {
+    if (userType.value === USER_TYPE.PERSONAL || userType.value === USER_TYPE.CORPORATE) {
+      return true;
+    }
+
+    return userType.value === USER_TYPE.CORPORATE_LIMITED && organizationsAddresses.value.length > 0;
+  });
```
maksimzinchuk, line 163: `organizationsAddresses` is now only the current page / search window, so keying off `.length` here silently switches a corporate-limited user's data source at runtime. The same assumption is used in `accountAddresses` (L106, outside the diff so I can't comment there) and in `selectedAddress` (L239).

Repro — CORPORATE_LIMITED, no `CanEditOrganization`, 10+ organization addresses:

1. type a search term that matches nothing → the org query returns `items: []`
2. `isPaginated` flips to false, `accountAddresses` falls back to `localShipToAddresses`, and `selectedAddress` starts resolving against `selectedLocalShipToAddressId` instead of the org address book

Two consequences beyond the broken dropdown:

- `canAddNewAddress` (`ship-to-selector.vue:242`) becomes true, so the user gets an "add new address" button they are not supposed to have. It opens `AddOrUpdateCompanyAddressModal`, but `handleAddressAddition` (L364-369) routes CORPORATE_LIMITED into the same branch as ANONYMOUS — so a "company" address is written to localStorage under a `local-` id and never reaches the organization.
- `category.vue:402` consumes this same `selectedAddress` and feeds it into the products query (`category.vue:497-505`), so catalog pricing and availability get silently re-requested for a different address.

On `dev` the full list stayed non-empty because filtering was client-side, so the source never switched. This decision needs to be independent of both the current page and the active keyword — arguably it should follow from permissions / user type rather than from any count.

client-app/modules/quotes/pages/edit-quote.vue
```diff
@@ -344,12 +358,28 @@ function openAddOrUpdateAddressModal(addressType: AddressType, currentAddress?:
...
+    organizationAddressesPage.value = 1;
+  } else {
+    personalAddressesPage.value = 1;
+  }
+
   openModal({
     component: SelectAddressModal,
     props: {
-      addresses: accountAddresses.value,
+      addresses: accountAddresses,
```
maksimzinchuk, line 370: Two notes on this call, plus two consequences elsewhere in the file.

**`pageSize` isn't passed.** `SelectAddressModal` computes `pages = ceil(totalCount / props.pageSize)` with `pageSize` defaulting to 6 (`select-address-modal.vue:310`), which happens to match `ADDRESSES_PER_PAGE` on L155. Changing that constant alone would silently desync the page count from the fetch size — worth passing `pageSize: ADDRESSES_PER_PAGE` explicitly.

**`addresses`, `loading` and `totalCount` are passed as refs**, while `isCorporateAddresses` on the next line is passed as `.value`. It does work: `useModal` keeps the stack in a deep `ref` (`useModal.ts:5`), so nested refs are unwrapped on property access, and `IModal.props` is `Record<string, any>` so TS won't flag the mismatch. The reactivity is also genuinely needed — without it the modal wouldn't re-render on page change. But it reads like an oversight next to the `.value` sibling; a one-line comment saying the refs are intentional would stop someone "fixing" it and breaking pagination.

**Moving this page to server pagination breaks two page-independent assumptions** further down (both outside the diff, so I can't comment inline):

- `accountAddressExists` (L306, called at L344) now only inspects the 6 addresses of the current page. Add an address from the quote that already exists on page 2 and the duplicate guard returns false, so a duplicate is persisted to the account/organization address book. On `dev` the whole list was loaded (`first: 9999`), so this was caught. Fixing it properly needs a server-side check — a targeted `keyword`/`ids` lookup before saving, or an idempotent upsert in `updateMemberAddresses`. Note that a plain client-side check-then-write still leaves a race.
- `userHasAddresses` (L272) is `!!accountAddresses.value.length`, i.e. "does the current page have items" rather than "does this user have addresses". `totalCount` is the right source here.

client-app/shared/ship-to-location/composables/useShipToLocation.ts
```diff
@@ -290,14 +375,19 @@ export function useShipToLocation() {
...
-    getFilteredAddresses,
-    getLimitedAddresses,
+    getLocalFilteredAddresses,
 
-    openSelectAddressModal,
     openAddOrUpdateAddressModal,
   };
 }
+
+export const useShipToLocation = createSharedComposable(_useShipToLocation);
```
maksimzinchuk, line 393: Two side effects of sharing this composable that are worth double-checking.

**The localStorage keys are pinned for the session.** `localShipToAddresses` and `selectedLocalShipToAddressId` (L84-92) read `userSuffix.value` once, at creation time. Previously the composable was re-created per consumer, so the key was re-evaluated on every call; now the instance lives as long as any consumer stays mounted — and the header selector always is. `useSignMeOut` reloads the page by default, which limits the blast radius, but the sign-in path is worth verifying so anonymous ship-to addresses can't stick to a signed-in user.

**Nothing covers the shared behaviour.** `useShipToLocation.test.ts:72` replaces `createSharedComposable` with identity, which is a reasonable way to keep the tests isolated — but it means the shared path is untested, and both issues I flagged above live exactly there. Two component tests on `ship-to-selector` would pin them down: the clear button after a zero-result search, and corporate-limited with an empty search result.

PR #2425, opened 2026-08-05
feat(VCST-5649): [Sales Rep] widget settings — max rows and order status tabs

ivan-kalachikov:
## Description

In layout-edit mode a sales rep can now cap how many rows each list widget shows, and choose which order status tabs the Recent Orders widget offers. Both settings persist with the layout, through the same Save / Cancel / Reset, on the Dashboard and the Customer profile.

**Stacked on #2400 (VCST-5367)** — this fills in the per-block `settings` key that ticket deliberately left empty. Review #2400 first; the base is `feat/VCST-5367-srh-layout-drag-and-drop`, not `dev`.

**No API change.** `InputSalesRepLayoutBlock.settings` has been in the contract since VCST-5367 and was going out as `[]`. Only the two `.graphql` documents change, to *select* `settings` back.

### How it works

The registry gains a third responsibility next to "which blocks exist" and "which region each lives in": which settings each block has, with its default, min and max. Every value read out of a saved document is validated against that descriptor, which makes the "fall back to defaults" requirement mechanical rather than a per-widget convention.

- **Max rows** is generic — `layout-widget.vue` renders the field in `VcWidget`'s published `#append` slot for any block whose registry entry declares a `maxRows` descriptor. A list widget shipped later gets the control by adding one registry line.
- **The cap applies on save, not as you type.** It is a query variable, so driving it from the draft refired `salesRepOrders` on every keystroke — typing "15" fetched 1 row, then 15, with a skeleton flash between. Widgets fetch with the *saved* value (`persistedSettingsOf`), the field edits the draft. The ticket asks only that settings persist with the layout; the commit-as-typed behaviour came from a prototype with no backend.
- **Status tabs** are specific to Recent Orders, which swaps its chips row for checkboxes while editing. Every tab may be unchecked — the always-present "All" baseline chip keeps unfiltered orders reachable, which is why no "at least one" constraint exists.
- **Wire format:** one scalar per setting (`settings.value` cannot hold a list), so the tab selection is flattened into sibling `tab.<rule>` keys. Only *unchecked* rules are written, so a status the backend adds later shows up checked for free.
- Settings fold into the existing layout draft rather than a parallel one, so Save / Cancel / Reset cover them with no second state machine.
- A stored tab name the backend no longer returns is dropped on entering edit mode — the one moment the live rule catalog is known. An empty catalog means the rules query has not resolved (or failed, which leaves it empty for the session), never that everything was retired.

### Deviations from the AC, confirmed with the requester

1. **Default checked set = every rule the backend returns**, not the AC's "New / On hold / Active / Fix-it". Those names come from the prototype's fictional status catalog; what ships reads `salesRepOrderFilterRules` at runtime, backend-labelled. Hardcoding four names would break on any store whose statuses differ. The widget looks exactly as it does today until a rep unchecks something.
2. **The max-rows range is per block, from the registry** — Recent Orders 1–20, Top sellers 1–10 — because `salesRepTopSellers` caps `take` at 10, so a rep can never ask for rows the backend will not return.
3. **The Activity widget is out of scope**: no Activity block exists in this repo. Its descriptor is one registry line whenever that widget ships.
4. **Stat cards get no settings** — nothing about a KPI card is configurable.
5. **AC #2's "at least one must stay checked" was dropped from the ticket**, since the "All" tab makes it redundant.

Kit components are consumed through published slots and props only — no `.vc-*` selectors, no `!important`, no inline fallbacks. The rows field's width is set by a class we own, which lands on `VcInput`'s root through its documented attribute fallthrough.

Design notes, the rejected alternatives, and two consciously deferred findings are in `client-app/modules/sales-rep/specs/VCST-5649-widget-settings/`.

### Verification

- `yarn vitest run --dir client-app/modules/sales-rep` — 23 files, 251 tests pass. New coverage: `layout/settings.test.ts` (clamping, non-numeric and unloaded-catalog fallbacks, unknown and duplicated `tab.*` keys, round-trip), `layout/document.test.ts` (settings in the payload, the echo check including a duplicated block type), `useSalesRepLayout.test.ts` (settings through cancel / reset / save / an in-flight save, draft vs saved accessors), `components/layout-widget-settings.test.ts` (the rows input's drag-filter class, clamping, draft-vs-saved binding, the toggles).
- `yarn vue-tsc --build --force` — clean.
- `yarn eslint client-app/modules/sales-rep` — no errors.
- Reviewed against the conventions in #2406, including its `check:comment-noise` and `check:style-conventions` scripts — both clean.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5649
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.55.0-pr-2425-ebfb-ebfb443a.zip


client-app/modules/sales-rep/components/sales-rep-orders.vue
```diff
@@ -155,8 +172,44 @@ const { from: periodFrom, to: periodTo } = useSalesRepPeriodFilter();
...
+
+// The backend's "All" passthrough would duplicate the baseline chip, so it is not offered as a tab
+// and cannot be unchecked either.
+const selectableRules = computed(() => selectableFilterRules(filterRules.value));
+
 // Show the filter chips only when the backend offers a real filter beyond the "All" baseline.
-const hasFilterOptions = computed(() => selectableFilterRules(filterRules.value).length > 0);
+const hasFilterOptions = computed(() => selectableRules.value.length > 0);
+
+// The rep's chosen tabs, in catalog order. Falls back to the whole catalog if a saved document hides
```
vkrashenko: There are no fallback, remove or actualize comment

client-app/modules/sales-rep/components/sales-rep-orders.vue
```diff
@@ -155,8 +172,44 @@ const { from: periodFrom, to: periodTo } = useSalesRepPeriodFilter();
...
+  }
+});
+
+function toggleTab(name: string): void {
+  chrome?.updateSettings({ hiddenTabs: toggleTabRule(effectiveHiddenTabs.value, name) });
+}
+
+// The active filter can be unchecked from under the chips — leaving it selected would filter by a tab
+// the rep can no longer see or clear.
+watch(visibleRules, (rules) => {
```
vkrashenko, line 216: unchecking the currently active filter tab clears filter, but cancel brings the tab back while the filter stays cleared. Maybe reset it only on save?

client-app/modules/sales-rep/components/sales-rep-orders.vue
```diff
@@ -155,8 +172,44 @@ const { from: periodFrom, to: periodTo } = useSalesRepPeriodFilter();
...
+// The rep's chosen tabs, in catalog order. Falls back to the whole catalog if a saved document hides
+// every rule the backend still returns — see layout/settings.ts.
+const visibleRules = computed(() => visibleTabRules(selectableRules.value, hiddenTabs.value));
+
+// What the checkboxes read: the stored list minus anything the catalog no longer contains.
+const effectiveHiddenTabs = computed(() => knownHiddenTabs(selectableRules.value, hiddenTabs.value));
+
+// Edit mode is where the widget knows the live catalog, so a name it no longer returns is cleared from
+// the store. Keyed on the catalog too, which can arrive after edit mode was entered.
+watch([editingTabs, selectableRules], () => {
```
vkrashenko: minor: a widget restored from the tray mid-edit mounts with editingTabs already true and selectableRules already settled (cache hit) - the watch then never fires and the prune is skipped for that edit session. { immediate: true } closes it; the editingTabs guard makes it a no-op on ordinary mounts

client-app/modules/sales-rep/specs/VCST-5649-widget-settings/2026-08-04-widget-settings-design.md
vkrashenko, line 1: you once flagged PORT_TO_MF.md for living inside client-app/modules/** — the specs/VCST-* dirs are the same placement (they don't reach the artifact, dist/* only, so it's purely a convention question). Is the intent to make in-repo specs the new pattern for this module?

vkrashenko (dismissed):

client-app/modules/sales-rep/components/sales-rep-orders.vue
ivan-kalachikov: Right — the fallback went with 1885c120 and the comment stayed. Rewritten in 399e5fae2.

client-app/modules/sales-rep/components/sales-rep-orders.vue
ivan-kalachikov: Confirmed, fixed in 399e5fae2. The chips now read `savedSettings`, so the tab set lands on save like the row cap already did, and a cancelled edit leaves the active chip alone. Two tests in the new `sales-rep-orders-tabs.test.ts` cover the cancel round trip and the clear-on-save.

client-app/modules/sales-rep/components/sales-rep-orders.vue
ivan-kalachikov: Verified both halves before taking it: the tray renders restore buttons only, so a restore is a fresh mount, and `@vue/apollo-composable` calls `start()` inline in setup and pushes `getCurrentResult` through on a cache hit — the catalog is settled before the watcher exists. `{ immediate: true }` added in 399e5fae2, with tests for the mid-edit mount, the ordinary one, and the unloaded catalog.

client-app/modules/sales-rep/specs/VCST-5649-widget-settings/2026-08-04-widget-settings-design.md
ivan-kalachikov: Yes, deliberate — `2b6b16c40` moved the VCST-5367 spec out of the repo root *into* the module for exactly this reason: sales-rep is meant to be portable as an MF remote, so its docs travel with it. Same as `federated/specs/`; the leftover root `specs/VCST-5469-…` is the odd one out, not this. I searched the repo's review comments for `PORT_TO_MF` and this thread is the only hit, so I think the earlier flag is a misremembering — happy to be pointed at it if it was somewhere else.

vas11yev1work (approved):

vkrashenko (approved):

PR #2426, opened 2026-08-06
feat(VCST-5450): use configurable company roles for invite, filters, and change role

DmitryGrishinVirtoworks:
## Description
Storefront now sources company member roles from the new `Customer.MembershipRolesWhitelist` store setting instead of a hardcoded role list, for the invite, change-role, and member-filter UIs.

## What was added
- **New composable `useCompanyMemberRoles`**: reads the public module setting and maps configured role ids/names to display objects; falls back to the same 3 default roles (Organization employee / Purchasing agent / Organization maintainer) the UI showed before this feature, when the setting is unconfigured or unparseable.
- **`invite-member-modal.vue`** and **`members.vue`** (Change role modal): source their role list from `useCompanyMemberRoles()` instead of the removed `B2B_ROLES` constant.
- **New composable `useOrganizationContactRoles`** + GraphQL query `getOrganizationContactRoles`: powers the "role" filter facet on the company members list with roles actually assigned within the organization, replacing a static hardcoded 5-role list.
- Removed dead code: the unused `permissions` field on role constants in `security.ts` (real permission checks go through the server session, not these constants), the now-unused `B2B_ROLES` constant, and `role-icon.vue`.
- Added `no_company_roles_available` translated message across all locales for the empty-state case.

## Breaking changes
None — this is an application, not a consumed library; no public API surface changes.

## Verification
- `vue-tsc --build --force` clean, ESLint clean
- Manually exercised invite / change-role / filter flows in the browser

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.57.0-pr-2426-8ae2-8ae22416.zip


client-app/shared/company/composables/useOrganizationContacts.ts
```diff
@@ -153,7 +154,7 @@ export function useOrganizationContacts(organizationId: MaybeRef<string>) {
     loading.value = true;
 
     try {
-      return await changeOrganizationContactRole(payload);
+      return await changeOrganizationContactRole({ storeId: globals.storeId, ...payload });
```
ivan-kalachikov, line 157: `storeId` isn't in the deployed schema. Against vcst-dev/qa this dies before it reaches the resolver:

`Variable '$command' is invalid. Unrecognized input fields 'storeId' for type 'InputChangeOrganizationContactRoleType'.`

So every "Change role" save breaks until profile-experience-api#143 ships and the envs are updated. Same coupling for the new `organization { contactRoles }` query, which 400s with `Cannot query field 'contactRoles' on type 'Organization'`.

client-app/shared/company/composables/useOrganizationContactsFilterFacets.ts
```diff
@@ -1,46 +1,45 @@
...
 
-  const appliedFacets = ref<FacetItemType[]>(cloneDeep(initialFacets));
-  const selectableFacets = ref<FacetItemType[]>(cloneDeep(initialFacets));
+    return facets.filter((facet) => facet.values.length > 0);
+  });
+
+  const appliedFacets = ref<FacetItemType[]>(cloneDeep(initialFacets.value));
+  const selectableFacets = ref<FacetItemType[]>(cloneDeep(initialFacets.value));
+
+  watch(contactRoles, resetFacets, { once: true });
```
ivan-kalachikov: This wipes filters the user already applied.

`resetFacets` rebuilds both arrays with everything unselected, but `roleIds`/`statuses` in `members.vue` are only synced by `syncFilterFromFacets()` on apply/reset. So: apply "Blocked" while the roles query is still in flight, response lands, chips and the badge clear — and the list stays filtered. Reset is disabled too (`:disabled="!numberOfFacetsApplied && !isFacetsDirty"`, both false after the wipe), so there's no way out except a reload.

Merge the role facet into the existing arrays preserving `selected`, instead of resetting.

client-app/shared/company/composables/useOrganizationContactRoles.ts
```diff
@@ -0,0 +1,23 @@
...
+  const contactRoles = ref<RoleType[]>([]);
+  const loading = ref(false);
+
+  onMounted(async () => {
+    loading.value = true;
+
+    try {
+      contactRoles.value = await getOrganizationContactRoles(organizationId);
+    } catch (e) {
+      Logger.error(`${useOrganizationContactRoles.name}`, e);
```
ivan-kalachikov, line 17: Any failure here removes the role filter with no trace. The ref keeps its initial `[]`, so the `once` watch never fires, and the facet builder drops facets with no values — the filter is simply gone for the rest of the session, log-only.

Falling back to `useCompanyMemberRoles()` would keep it usable, or keep the facet visible in a disabled state.

`loading` on line 22 is unused, by the way — the only caller destructures `contactRoles` only.

client-app/core/api/graphql/types.ts
```diff
@@ -4786,6 +4787,7 @@ export type Organization = {
   /** Business category */
   businessCategory?: Maybe<Scalars['String']['output']>;
   contacts?: Maybe<ContactConnection>;
+  contactRoles?: Maybe<Array<RoleType>>;
```
ivan-kalachikov: This file wasn't regenerated — `grep -c GetOrganizationContactRoles` returns 0, and codegen emits four artifacts per operation (`...Query`, `...QueryVariables`, `...Document`, and an `OperationNames` entry).

Nullability is off as well: the backend declares `Field<ListGraphType<RoleType>>`, which codegen renders as `Maybe<Array<Maybe<RoleType>>>` — compare `rolesInOrganization` on line 801. The next `yarn generate:graphql-types` will flip this and break `getOrganizationContactRoles`.

Run codegen against a backend that has the paired change and commit the real output.

client-app/shared/company/composables/useCompanyMemberRoles.ts
```diff
@@ -0,0 +1,44 @@
...
+
+function toRole(id: string, t: ReturnType<typeof useI18n>["t"]): ExtendedRoleType {
+  const idLower = id.toLowerCase();
+  const knownRole = ALL_ROLES.find((role) => role.id.toLowerCase() === idLower || role.name.toLowerCase() === idLower);
+
+  if (knownRole) {
+    return knownRole;
+  }
+
+  return { id, name: t(`common.roles.${id}`, id) };
```
ivan-kalachikov: Whitelist entries that aren't in `ALL_ROLES` end up with `id` = the raw setting string (which is a role *name*), while `currentRoleId` in the change-role modal is the real platform role id. For any custom role that means the radio never preselects and Save never disables.

Second thing: the fallback arg to `t()` is compiled as a message format, so a role name containing `{`, `}`, `@` or `|` renders wrong — and `missingWarn` is off, so silently.

client-app/pages/company/members.vue
```diff
@@ -670,14 +671,16 @@ function openEditCustomerRoleModal(contact: ExtendedContactType): void {
...
       currentRoleId: contact.extended.roles[0]?.id,
       loading: contactsLoading,
 
       async onConfirm(selectedRoleId: string): Promise<void> {
+        const selectedRole = companyMemberRoles.value.find((role) => role.id === selectedRoleId);
+
         const result = await changeContactOrganizationRole({
           memberId: contact.id,
-          roleIds: [selectedRoleId],
+          roleIds: [selectedRole?.name ?? selectedRoleId],
```
ivan-kalachikov: Any reason for the name over the id? The backend takes either (`FindByIdAsync ?? FindByNameAsync`), but names are renameable and not unique, and everything else here compares by id — facet values, radio values, `currentRoleId`. For the five known roles `selectedRole?.id` behaves identically.

Same at `invite-member-modal.vue:158`.

client-app/pages/company/members.vue
```diff
@@ -670,14 +671,16 @@ function openEditCustomerRoleModal(contact: ExtendedContactType): void {
   const closeEditCustomerRoleModal = openModal({
     component: EditCustomerRoleModal,
     props: {
-      roles: B2B_ROLES,
+      roles: companyMemberRoles.value,
```
ivan-kalachikov: These reach the change-role modal unnormalized, and it renders `role.name` straight out — so in DE/FR the change-role dialog shows English role names while the invite modal and the filter show translations. Map them through `t("common.roles." + id, name)` the way `invite-member-modal.vue` does.

ivan-kalachikov (commented):
Blocking on merge order: `storeId` and `organization.contactRoles` don't exist on vcst-dev or vcst-qa yet — the paired backend PRs (profile-experience-api#143, customer#314) are both still open. Every role change currently fails variable coercion:

`Variable '$command' is invalid. Unrecognized input fields 'storeId' for type 'InputChangeOrganizationContactRoleType'.`

Rest inline. Two items with no diff line to hang on:

- `client-app/assets/images/role-{org-employee,org-maintainer,purchasing-agent,store-admin,store-manager}.svg` are unreferenced now that `role-icon.vue` is gone — it was their only consumer. Delete them with it.
- `useCompanyMemberRoles` is pure parse/fallback logic and ships untested. The facet-reset bug below would have been caught by one too.

client-app/shared/company/components/edit-customer-role-modal.vue
vas11yev1work, line 14: The change-role modal renders `role.name` directly with no i18n translation, while the invite modal translates the same role list via `t("common.roles." + role.id, role.name)`.

client-app/shared/company/composables/useCompanyMemberRoles.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export function useCompanyMemberRoles() {
+  const { t } = useI18n();
+  const { getSettingValue } = useModuleSettings(CUSTOMER_MODULE_ID);
+  const roles = computed<ExtendedRoleType[]>(() => {
```
vas11yev1work, line 55: `useCompanyMemberRoles` (invite / change-role modals) sources roles from the `Customer.MembershipRolesWhitelist` (`COMPANY_MEMBER_ROLES_SETTING`) module setting, while `useOrganizationContactRoles` (the filter facet) sources roles from the `organization.contactRoles` GraphQL field. These are two independent lists with different fallback behavior (whitelist falls back to 3 default roles; contactRoles falls back to empty). Is it intentional that "assignable roles" and "filterable roles" can come from different sources and potentially disagree, or should both consume the same source of truth?

client-app/shared/company/composables/useCompanyMemberRoles.ts
```diff
@@ -0,0 +1,44 @@
...
+    const rawValue = getSettingValue(COMPANY_MEMBER_ROLES_SETTING);
+
+    if (typeof rawValue === "string" && rawValue.length) {
+      try {
+        const roleIds = JSON.parse(rawValue) as unknown;
+        if (Array.isArray(roleIds) && roleIds.every((id) => typeof id === "string")) {
+          return roleIds.map((id) => toRole(id, t));
```
vas11yev1work: `roles` has no de-dup — if the whitelist setting contains duplicate/aliased role ids (e.g. "org-employee" and "Organization employee", matched case-insensitively by `toRole`), two entries end up with the same id, breaking `:key="role.id"` in the role-selection modal. Worth deduping by id here.

client-app/shared/company/composables/useOrganizationContacts.ts
DmitryGrishinVirtoworks: Waiting on profile-experience-api#143 to merge and deploy to vcst-dev/qa - the frontend code is correct, but storeId isn't in the deployed schema yet. Will verify once the backend PR ships.

client-app/shared/company/composables/useOrganizationContactsFilterFacets.ts
DmitryGrishinVirtoworks: Done

client-app/shared/company/composables/useOrganizationContactRoles.ts
DmitryGrishinVirtoworks: Done

client-app/core/api/graphql/types.ts
DmitryGrishinVirtoworks: Done

client-app/shared/company/composables/useCompanyMemberRoles.ts
DmitryGrishinVirtoworks: Done

client-app/pages/company/members.vue
DmitryGrishinVirtoworks: Done

client-app/pages/company/members.vue
DmitryGrishinVirtoworks: Done

client-app/shared/company/components/edit-customer-role-modal.vue
DmitryGrishinVirtoworks: Done

client-app/shared/company/composables/useCompanyMemberRoles.ts
DmitryGrishinVirtoworks: Done

client-app/shared/company/composables/useCompanyMemberRoles.ts
DmitryGrishinVirtoworks: Assignable roles (whitelist) and filterable roles (contactRoles) are separate on purpose: the whitelist is what can be assigned, contactRoles is what's actually assigned - different by design, not a bug.

client-app/pages/company/members.vue
ivan-kalachikov: `members.vue:686` sends the id now, but `invite-member-modal.vue:158` still sends `selectedRole?.name ?? data.roleId`.

That half isn't cosmetic: for a custom role that has a `common.roles.<id>` translation, `toRole` puts the *translated* label into `name` (`useCompanyMemberRoles.ts:24`). So a DE/FR storefront sends "Vertriebsleiter" into `ResolveRoles`, where `FindByIdAsync ?? FindByNameAsync` matches neither — "Role not found". The five known roles are safe, their `name` is the English constant.

`selectedRole?.id`.

client-app/shared/company/composables/useCompanyMemberRoles.ts
ivan-kalachikov: The `te()` guard in `toRole` is right, but the same `t(key, role.name)` fallback is still live at three call sites: `members.vue:679`, `invite-member-modal.vue:100`, `useOrganizationContactsFilterFacets.ts:40`.

Measured against vue-i18n 11.4.6 as configured here (full build, no runtime-only alias):

```
"Sales | Manager"  ->  "Sales"        plural split
"Team {A} Lead"    ->  "Team  Lead"   interpolation
"@:admin"          ->  "admin"        linked message
"a@b"              ->  throws: Message compilation error: Invalid linked format
```

The throw lands inside a computed, so it takes the render with it, not just the label. The same `te()` guard covers all three.

DmitryGrishinVirtoworks: Done, thanks

client-app/pages/company/members.vue
DmitryGrishinVirtoworks: Done

ivan-kalachikov (dismissed):

vas11yev1work (approved):

PR #2427, opened 2026-08-06
feat(VCST-5159): extend the core facade so sales-rep can ship as a plugin

ivan-kalachikov:
## What this is for

The `sales-rep` module should ship as a separate Module Federation plugin, from its own repo. To do that it needs everything it uses from the host to come through `@vc-frontend/core` — the facade — instead of `@/...` paths.

This PR widens the facade to cover what the module actually imports, and fixes the one case that did not fit.

After this, **every host import in `modules/sales-rep` resolves to the facade** — runtime imports through the root export, and its specs' mount helpers through a new `@vc-frontend/core/testing` subpath.

---

## The one case that did not fit: the mobile nav badge

The "My customers" link in the mobile menu shows a count badge.

To render it, the plugin's component had to import `MobileMenuLink` from the host's `_internal/` folder. That is a private folder — once a plugin depends on it, the host can no longer restyle its own menu without breaking the plugin. And the plugin now lives in a different repo with its own release cycle.

So an extension entry can now contribute **data** instead of markup:

```ts
// before — the plugin renders the menu link itself
register("mobileMenu", MY_CUSTOMERS_NAV_LINK_ID, {
  component: defineAsyncComponent(() => import("./components/link-my-customers-mobile.vue")),
});

// after — the host renders its own link, the plugin supplies the number
register("mobileMenu", MY_CUSTOMERS_NAV_LINK_ID, {
  use: useSharedSalesRepCustomersCount,
});
```

`link-my-customers-mobile.vue` is deleted. The desktop link stays a component on purpose — it carries real logic (keeping the link active on the sibling customer-profile route) and never imported anything internal.

### Why `use()` is a function and not a value

The count comes from a GraphQL query, so it has a lifecycle: start it, stop it, refresh it. A plain value cannot express that.

`ExtensionPoint` calls `use()` **in its own setup**, inside a scope it stops when the entry unmounts. That is the only place Vue lets you register cleanup. `useSharedSalesRepCustomersCount` wraps the composable in `createSharedComposable` (the repo's existing idiom — `useCart`, `usePayment`, `useUserQuote`): one query for however many surfaces show the badge, disposed when the last unmounts, rebuilt fresh on the next mount.

An earlier revision of this PR passed a getter through `props` instead. It ran during render, where lifecycle hooks silently do not register, and it memoised the query for the whole session — so the badge kept showing a number from earlier while the desktop link showed the current one. `data-freshness.test.ts` asserts exactly the behaviour that was lost.

### Contributing is not allowed where the host has no fallback

`ExtensionPoint` renders the plugin's component **or** the host's fallback slot. A component-less entry only works where that fallback exists — and in 6 of 9 categories it does not: those call sites are self-closing, so such an entry would render nothing at all, with no error.

A category opts in by declaring a contributed shape:

```ts
type ExtensionEntryType<Props, Contributed, Condition> =
  [Contributed] extends [never]
    ? ReplaceEntryType<Props, Condition>                            // component required
    : ReplaceEntryType<Props, Condition> | DecorateEntryType<...>;  // either mode
```

Leave `Contributed` at `never` and the decorate branch does not exist, so registering one there is a **compile error**, not a silent no-op. Today only `mobileMenu` opts in.

`ExtensionPointList` now delegates to `ExtensionPoint` per entry, so both renderers agree on what a component-less entry means.

---

## The rest of the facade

Grouped by export shape — the tiering rule is now written down in `client-app/core-api/README.md`, so the next request answers itself.

| Tier | Shape | What went in |
| --- | --- | --- |
| 1. Capability | composable / function | `registerCacheTypePolicies`, `useWishlistSharingScopes`, `Logger`, `useUser`, `useNavigations`, `useModal`, `useNotifications`, `useBreadcrumbs`, `usePageHead`, `getProductRoute`, `EXTENSION_NAMES`, menu types, the sharing-scope controls type |
| 2. Host-rendered chrome | nothing — an extension entry | the sales-rep mobile nav link |
| 3. Frozen component | a `.vue` component | `OrderStatus` |

Plus a new subpath export alongside the existing `/federation` and `/tailwind-preset`: **`/testing`**, for the mount helpers a plugin's specs need.

**`registerCacheTypePolicies`** lets a plugin normalise its own GraphQL types without importing the Apollo cache. It warns in development when a typename is already in the cache — policies do not apply retroactively, and registering late otherwise fails silently. Its first consumer is on this branch: the layout policies from #2400 register through it, so `modules/sales-rep` no longer touches the cache instance.

**ui-kit** — all **21** `Vc*` components the module's templates use are exported, up from 7. The host registers them globally, but a remote only resolves that when it renders inside the host's app instance, and a plugin's own dev server has none.

**`@vc-frontend/core/testing`** ships `createWrapperFactory` / `createShallowWrapperFactory` as *real source*, unlike the root export: that one is types-only because the host injects the implementation at runtime, and a plugin's specs run with no host to inject anything. `@vue/test-utils` is passed in rather than imported, so the module never reaches for the consumer's copy. The host's own `@/core/utilities/tests` re-exports it, so the four in-repo specs are untouched.

The Apollo cache is deliberately *not* exported. `data-freshness.test.ts` needs *an* Apollo cache, not the host's instance — the host's type policies are cart/checkout-specific, and the ones that spec exercises come from the module's own `layout/cache-policies.ts`.

**`OrderStatus`** is exported as-is because its status → colour mapping comes from per-store `orders_statuses` settings; a plugin-side chip would drift. Its four props are contract from here on.

---

## Versioning: the contract goes to 0.1.0

It was heading for 1.2.0. It is not ready to promise stability to a third-party plugin author, so it moves to a pre-1.0 line instead.

Nothing breaks by doing so: **no `core-v*` release has ever been published**, so `0.1.0` is the first version anyone can actually install.

Under `^0.x` the **minor** is what a caret range refuses, so the levels shift down one: additive is a **patch**, breaking is a **minor**. The versioning automation now picks its levels from the release line it is on; the 1.x mapping is untouched and takes over at 1.0.0.

`core-v0.1.0` still has to be published by the *Core Facade Release* workflow before a plugin can install against it.

A plugin pins the facade by tarball URL, so anything added to the package later re-pins every plugin. That is why the two remaining gaps — ui-kit coverage and the specs' mount helpers — are closed here rather than deferred to the move.

---

---

## Defects found and fixed while reviewing this branch

Each one below is a real defect in this PR's own additions, with a test that fails without the fix.

### `@vc-frontend/core/testing` did not work for the only consumer it exists for

`testing.mjs` imports `lodash-es` at runtime. It was a peer only, and `create-plugin` drew its
`devDependencies` from `CONTRACT_TYPE_PEERS` — where `lodash-es` never appeared. So the first spec in
a scaffolded plugin importing the mount helpers died on `Cannot find module`. `vue-i18n` survived by
happening to be MF-shared.

Same root cause, wider blast radius: `peerDependencies` declared **2 of the 14** packages the
published files import. The real list lived in build-time JS objects no package manager reads, so a
consumer installing the tarball by hand — a supported path, `PORT_TO_MF.md` §4 — was missing five type
peers, and `skipLibCheck: true` in the scaffold's own tsconfig kept it silent: only the individual
members degraded to `any`, never the top-level exports, so `mergeMenuSchema(42)` and
`mergeMenuSchema(null)` compiled clean on the very call `PORT_TO_MF` says to keep.

The declaration is now **derived, not maintained**. `build-types.mjs` scans every file in `files` —
not just the rolled contract, which is why `testing.mjs` went unchecked — for `from`, `require()` and
the `import("x")` form the `.d.mts` files use, and fails with the exact object to paste when
`package.json` disagrees. Required vs optional follows reachability; `create-plugin` merges the
facade's own `peerDependencies` so there is no second list left to drift.

### Five holes in `registerCacheTypePolicies`, the new plugin-facing cache seam

- **Ownership was per typename, but Apollo merges `fields` per field name.** So the first plugin to
  register any `Query.fields.*` owned all of `Query`, and every later plugin's policy for a *different*
  root field was refused wholesale — a refusal protecting nothing, since Apollo merges the two
  additively. Claims are keyed per field now, with a deliberately **asymmetric** collision rule:
  `keyFields` re-keys the whole entity, so a type-level claim collides with the typename *and* every
  field under it. Without that asymmetry the host's `Product: { fields: { properties } }` —
  fields-only, like most host policies — would leave `Product` formally unclaimed and a plugin could
  redefine how the host identifies it. A spec pins that case.
- **The late-registration warning could not see the two shapes that matter.** A root-type policy never
  produces a `Query:` key — its fields live under `ROOT_QUERY` — so a late `Query.fields.x.keyArgs`,
  the commonest hazard there is, orphaned everything already stored under the old field key and warned
  about nothing. A typename stored inline (no `id`) was invisible for the same reason.
- **Re-registering your own claim reported a collision with yourself**, which is what HMR and a second
  `init()` both look like, sending the reader hunting for a plugin that does not exist.
- **`priority` had no ceiling, so the docblock's central promise was false.** `by.priority >= priority`
  is the whole gate and `priority` came from the caller, so `{ owner: "x", priority: 101 }` took the
  host's `CouponType: { keyFields: ["code"] }` over with zero refusals and zero warnings, and the cache
  really re-keyed by `id`. A caller's value is clamped below `HOST_PRIORITY` now. The host seeds its own
  claims straight into the ownership map, never through the public function, so the cap costs it nothing
  and plugin-vs-plugin escalation is untouched. A non-finite priority is sanitized in the same step:
  every `>= NaN` comparison is false, so `NaN` refused nothing at all.
- **`owner` defaulted to `"unknown"`, which turned the same-owner exemption into a hole.** Two plugins
  that both omitted it were indistinguishable, so neither blocked the other and the loser got no signal —
  in production none at all, since `Logger` is a no-op there. `owner` is required now, so every refusal
  names someone. No `core-v*` contract is published, so tightening the signature costs no consumer;
  after the first release it would not be free, and the contract gate compares export *names*, so it
  would not have flagged it.

And the docs stop overclaiming: `cache.ts` promised a plugin "cannot quietly take over a typename the
host normalizes", but the ownership map is seeded from the policies the host *declares*, which is not
the set it stores — anything Apollo normalizes by its default `id` rule is unclaimed. Stated as a known
gap now rather than a guarantee.

### The typing CI step was checking zero files

`vitest --typecheck` spawns `tsc --noEmit` from the repo root without `-p` unless `typecheck.tsconfig`
is set, so it inherited the root `{"files": []}` config: every `.test-d.ts` reported "no errors"
regardless of content — a planted `TS2322` and an unused `@ts-expect-error` both stayed green. Which
matters here, because `useExtensionRegistry.test-d.ts` is where this PR's central type guard lives, so
anyone verifying that claim by running the type test got a false pass. (The Build step's
`vue-tsc --build` *was* catching them — the step named after typing was not.)

Pointing it at a project surfaced 26 pre-existing errors across the repo and red-lit CI for things
this PR is not about, so it now uses a dedicated `tsconfig.typecheck.json` scoped to `**/*.test-d.ts`.
TypeScript still pulls in what those files import. `yarn test:typing --run` exits 0 clean and 1 on a
planted error — checked both, exit code included.

### Smaller, each with a test that fails without it

- **A released contract line only moves forward.** `escapesCaretRange` asked "is the current version
  outside `^base`", which is also true *below* the baseline — so `1.2.0 → 1.1.0` with an export removed
  returned "major already bumped" and would have minted a release whose contract has fewer exports than
  the earlier version number. A regression is refused outright now, gated on `released` so the
  deliberate 1.2.0 → 0.1.0 reset this PR performs stays legal. The bump guard also stops printing
  "passed" when it compared the contract with itself.
- **Both dependency-cruiser boundary rules are blind to `import type`.** `tsPreCompilationDeps` is off,
  so type-only edges never enter its graph — proven with probes: three value imports caught, the same
  three as `import type` produce zero violations. That is exactly the coupling that breaks when
  `modules/sales-rep` moves to its own repo. Turning the flag on is not a drop-in (`true` → 19 errors,
  `"specify"` → 15), so the bans are restated in `eslint.config.js`, which sees `import type`. **Honest
  limit:** CI runs `yarn build` → `validate:types` + `validate:dependencies`, not eslint, so the
  type-only half holds in the IDE and `yarn lint`, not the pipeline. Recorded in the config; closing it
  properly is a separate decision (scope the `tsPreCompilationDeps` work, or add lint to `validate`).
- **Replacing a contribution under the same name kept the old one running** and never started the new
  one — the child was keyed on `name`, and it reads `use` once in setup. Not reachable today (nothing
  calls `unregister` but specs, and both triggers the reviews assumed hit the duplicate guard instead),
  but the registry README teaches the unregister-then-register pattern, so it is a trap for the first
  plugin author to update a contribution. Keyed on the contribution's identity now.
- **A released line's "already bumped" reason named the wrong level.** `breaking` comes from
  `policyFor(baseVersion)`, so promoting `0.9.0 → 1.0.0` with an export removed reported `"minor already
  bumped"` when the human moved the major. The decision was right, the wording was not; it reads the
  bump that HAPPENED off `semver.diff` now. The `require-` path still names the level a human is *asked*
  for, which is the baseline's.
- **Two paths lost the `_internal` ban entirely.** The sales-rep block waives `NO_RAW_APOLLO_QUERY` for
  its specs and `useSalesRepHubQuery.ts`, but the two boundary blocks below skip all of `sales-rep/**`,
  so those files matched no `no-restricted-syntax` block at all — value *and* type imports of a host
  `_internal/` component went uncaught, and dependency-cruiser sees neither type-only edge. A trailing
  block restates just that ban for them, raw-Apollo waiver intact. A spec travels with the module when
  it is extracted, so a host component mounted in one breaks the extraction exactly as a source import
  would.
- **`EXTENSION_NAMES.mobileMenu.myCustomers`** — the registry README's canonical example — is
  `undefined` at runtime while type-checking as `string`, so copying that line registers under the
  literal string `"undefined"`. `EXTENSION_NAMES` is typed as its own declared names now, making that a
  compile error.
- The README told authors a late registration runs `use()` **outside** a component context where
  `inject()` does not resolve — the opposite of what the code does and of this PR's own passing spec.
  Believing it means avoiding every injection-dependent composable, i.e. all of Apollo, which is the
  entire reason `use()` is a function.
- `@vc-frontend/core/tailwind-preset` shipped with no `types` condition, so a TypeScript
  `tailwind.config.ts` failed with TS7016. `window.modulesCacheDebug` was declared always-present
  while only development assigns it. `default-menu.test.ts`'s "renders no count" asserted
  `toContain("my customers:")`, which `"my customers:3"` also satisfies — the test could not fail.
- **The tier table authorised one `.vue` export and the facade ships 22 `Vc*` plus `uiKit`.** They are
  there for a real reason (a plugin's own dev server has no host app instance, so nothing resolves a
  globally-registered component), but a reviewer had no written basis to refuse the 23rd. Now a tier of
  its own with one admission rule, and the unenforced half is stated: the gate compares exported
  *names*, so a renamed ui-kit prop ships as a patch and every plugin pinned to the range keeps
  resolving it. A prop-level diff would make that promise real; until then a ui-kit prop rename is a
  manual breaking bump. (Correcting this PR's own earlier claim: the facade went from **3** `Vc*` at
  the merge base to 22, and from 10 public names to 53 — not "up from 7".)

### Sonar

The failing quality gate was one `javascript:S2871` BUG on `readdirSync(dir).sort()` — a false
positive (S2871 exists for numeric arrays; the default comparator on strings already *is* UTF-16 code
units, identical to the file's own helper), fixed anyway since a bare `.sort()` reads as a latent
numeric-sort bug. The MAJOR `S6661` "use object spread" on `extension-point.vue` is **stale**: there is
no `Object.assign` in any revision of that file on this branch, none in its compiled output, and the
issue carries `"line": null` with a creation date eight days before the file was rewritten. The
remaining seven smells introduced by these fixes are cleared, and the `String.raw` ones turned out to
sit on a real defect: `new RegExp(`\b${name}\b`)` interpolated identifiers raw, and the
ambient-globals collector accepts `$` in identifiers — which is a regex *end anchor*, so `\b$VcTheme\b`
could never match and such a declaration was skipped with no diagnostic.

## Verification

- `yarn test:typing --run` (what CI runs) — **2188 tests / 143 files**, no type errors, exit 0
- `yarn vitest run --dir client-app` — **2150 tests / 140 files**, exit 0
- `yarn build` — exit 0
- `vue-tsc --build --force` — 0 errors
- `yarn validate:dependencies` — 0 errors
- `yarn validate:core-types` — contract matches the facade; peer declaration covers 10 required + 4
  subpath-only packages
- `eslint` — 0 errors on the changed files
- Merged `dev` twice. The first merge (6 commits) mattered because `dev` had independently rewritten
  three files this branch touches, including `data-freshness.test.ts` — the spec this PR offers as its
  freshness guarantee — changing its GraphQL fixture shapes.
  The second (4 commits, up to #2439) conflicted for real, in `modules/sales-rep`, and both conflicts
  were resolved in this branch's favour where the two disagree on purpose:
  - dev calls `registerLayoutTypePolicies(cache)`; this branch renamed that export and routes it
    through `registerCacheTypePolicies`, so the module no longer imports the cache instance;
  - dev registers the mobile My customers link as a component — the very component this PR deletes —
    so `registerContribution` stays and dev's permission-gated documents block sits ahead of it;
  - dev's new documents route mounts via `ROUTES.COMPANY.NAME`, like its siblings here.
- Regenerated the GraphQL types from `vcptcore-dev`. Core types only; the eight module schemas came
  back byte-identical. This picks up the document-library surface #2439 added to `Query`, which the
  core snapshot had not caught up with, and drops `CartType.loyaltyPoints` — an environment
  difference, not a schema change. Nothing reads that field, and the facade contract is unaffected:
  the `loyaltyPoints` it inlines belongs to `Product`.

---

## Notes for the reviewer

- `yarn validate:core-types` is stale on `dev` itself — it does not run in `yarn validate` during the pilot — so regenerating here also pulls in the `sharedList` category from #2405 that `dev`'s committed contract predates.
- `isRegistered()` answers "has a component", not "is registered" — a decorate entry returns false. That reading is what hid the dropped-contribution bug. The name is left alone here (renaming it is churn this PR does not need) and noted where it is defined.
- `headerMenu` and `accountMenu` have fallback markup but their slots are unscoped, so contributed data would be discarded, and both leave `Contributed` at `never` — the compile gate keeps them closed until those slots are scoped. `mobileMenu`'s two call sites are both scoped here: `default-menu.vue` is where today's badge renders, since the My customers link is an account-section child; `main-menu.vue` opens decorate mode for top-level mobile items, which nothing uses yet.
- `modules/push-messages` still imports two `_internal/` components. The same fix applies to it, out of scope here — and both files are the documented exemption in the new eslint rule as well as in dependency-cruiser.
- Two things about the badge that look like defects and are not, so they need not be re-raised. The badge is not stale across mount/unmount: staleness is bounded to the account-shell visit, because `account-shell.vue` renders the desktop rail `max-md:hidden` — mounted, not unmounted — which is the trade-off that makes the two surfaces agree. And `data-freshness.test.ts` does catch the regression it guards: reintroducing the session-lifetime memo, or switching to `cache-first`, turns it red.
- One real behaviour delta in that area, stated rather than fixed: a failed count fetch hides the badge for the rest of the account-shell visit with no retry. Pre-PR each surface had its own query, so reopening the mobile menu re-issued it. The designed failure mode is a hidden badge and the consistency win is the point, but it is a behaviour delta worth knowing.
- `sortablejs` is deliberately **not** in `MF_SHARED_RANGES`: a leaf DOM library with no cross-copy state, so a duplicate costs bundle size, not correctness. The HOWTO documents that rule, plus the exception that does force sharing — a library whose state crosses the boundary via `provide`/`inject`. `vee-validate` is the near-miss.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5159
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.58.0-pr-2427-8344-8344665b.zip


client-app/core/api/graphql/config/registerCacheTypePolicies.ts
```diff
@@ -0,0 +1,253 @@
...
+ *
+ * The same owner re-registering its own claim is never a collision: HMR and a second `init()` both
+ * land here, and reporting that as a conflict points the reader at a second plugin that does not
+ * exist.
+ */
+function blockedBy(typename: string, key: string, typeLevel: boolean, owner: string, priority: number) {
+  const prefix = `${typename}.`;
+  for (const [held, by] of owners) {
+    const collides = typeLevel ? held === typename || held.startsWith(prefix) : held === key || held === typename;
+    if (collides && by.owner !== owner && by.priority >= priority) {
```
goldenmaya, line 87: `priority` has no ceiling, so a plugin passing anything above `HOST_PRIORITY` takes a host policy over — which is the one thing the docblock on `registerCacheTypePolicies` promises cannot happen:

> The host's own policies sit at {@link HOST_PRIORITY}, so no plugin can take one over.

`by.priority >= priority` is the whole gate, and `priority` arrives from the caller.

Verified, against the host's real `CouponType: { keyFields: ["code"] }` (`cache.ts:51`):

```ts
registerCacheTypePolicies({ CouponType: { keyFields: ["id"] } }, { owner: "evil-plugin", priority: 101 });
```
```
owner of CouponType after: {"owner":"evil-plugin","priority":101} | rejections=0 | warnings=0
cache.policies.identify({ __typename: "CouponType", id: "X", code: "C" })  ->  CouponType:{"id":"X"}
```

No refusal, no warning, and the cache really re-keys by `id`. Control case passes: the same registration at the default priority **is** refused and recorded in `rejected`, so the mechanism works — it just has no upper bound.

This matters more than an internal API would, because the signature is in the published contract (`core-api/contract/index.d.ts:2561`) and in the root export list, so any third-party plugin author can reach it:

```ts
declare function registerCacheTypePolicies(policies: TypePolicies, { owner, priority }?: {
    owner?: string;
    priority?: number;
}): void;
```

Either clamp a caller-supplied `priority` below `HOST_PRIORITY` (the host seeds its own claims directly, so it does not need the public path), or drop the guarantee from the docblock and say host claims are outrankable. The first seems closer to the intent — the escalation the spec exercises is plugin-vs-plugin, which a clamp leaves untouched.

client-app/core/api/graphql/config/registerCacheTypePolicies.ts
```diff
@@ -0,0 +1,253 @@
...
+ * more typenames than it writes policies for — anything Apollo normalizes by its default `id` rule
+ * holds no policy and is therefore unclaimed, so a plugin registering `keyFields` for one of those
+ * is accepted. The check is a collision guard between declared policies, not a fence around the
+ * host's whole cache surface.
+ *
+ * In development the ownership map and every refusal are readable as `window.modulesCacheDebug`.
+ */
+export function registerCacheTypePolicies(
+  policies: TypePolicies,
+  { owner = "unknown", priority = 0 }: { owner?: string; priority?: number } = {},
```
goldenmaya: `owner` defaulting to `"unknown"` turns the same-owner exemption into a hole: two plugins that both omit it are indistinguishable, so `by.owner !== owner` is false and neither blocks the other.

The exemption itself is right — HMR and a second `init()` both need it. But it keys on a string that has a shared default, and forgetting an optional argument is the likeliest way to end up here.

Verified:

```ts
registerCacheTypePolicies({ VerifyAnon: { keyFields: ["a"] } }); // owner -> "unknown"
registerCacheTypePolicies({ VerifyAnon: { keyFields: ["b"] } }); // owner -> "unknown"
```
```
first={"owner":"unknown","priority":0} second={"owner":"unknown","priority":0}
collisions=0 warnings=0 identify=VerifyAnon:{"b":"2"}
```

Second registration wins, `rejected` stays empty, `Logger.warn` is never called. Both plugins believe they own the typename, and the one that lost has no signal at all — in production none of this is visible anyway, since `Logger` is a no-op there.

Two options, either works: make `owner` required (it is the argument that makes a refusal nameable, and every in-repo caller already passes it — `sales-rep/index.ts:100` does), or treat the sentinel as never matching itself, e.g. `by.owner !== owner || owner === UNKNOWN_OWNER`. The first also removes the `"unknown"` label from refusal messages, which points at nobody.

eslint.config.js
```diff
@@ -442,15 +472,32 @@ export default defineConfigWithVueTs(
       "client-app/modules/sales-rep/**/*.test.ts",
```
goldenmaya, line 475: These two `ignores` entries were meant to waive `NO_RAW_APOLLO_QUERY`, but because the two blocks below also skip all of `sales-rep/**`, these files end up matching **no** block that sets `no-restricted-syntax` — so they lose the `_internal` ban too.

The residual gap is exactly the case this eslint restatement exists for. dependency-cruiser still catches a value import, but not a type-only one, and neither catches a type-only import in these files. Probed all six combinations:

| probe | eslint | depcruise |
| --- | --- | --- |
| module, `import` `_internal` | caught | caught |
| module, `import type` `_internal` | caught | not seen |
| **sales-rep spec, `import` `_internal`** | **not caught** | caught |
| **sales-rep spec, `import type` `_internal`** | **not caught** | **not seen** |
| **`useSalesRepHubQuery.ts`, `import` `_internal`** | **not caught** | caught |
| host → `@/modules/sales-rep/...` | caught | caught |

(The `import type` rows are what the block comment above predicts for dependency-cruiser, and it is right — the value/type asymmetry reproduced exactly as described.)

Worth closing because specs travel with the module when it is extracted: a spec importing host `_internal/` breaks the extraction the same way a source file does, and it is the file most likely to reach for a host component to mount.

Smallest fix is to narrow the waiver to the rule it was for, rather than to the file — keep these paths inside the block and give them their own entry with just `NO_HOST_INTERNAL_IMPORT`:

```js
{
  files: ["client-app/modules/sales-rep/**/*.test.ts", "client-app/modules/sales-rep/composables/useSalesRepHubQuery.ts"],
  rules: { "no-restricted-syntax": ["error", NO_HOST_INTERNAL_IMPORT] },
},
```

placed after the sales-rep block. Given how carefully the replacement semantics are documented here, this reads like an oversight rather than a decision — but say so if it was deliberate.

client-app/shared/layout/components/header/_internal/mobile-menu/menus/main-menu.vue
```diff
@@ -21,7 +21,14 @@
...
         >
-          <LinkDefault :item="item" @close="$emit('close')" @select-item="$emit('selectItem', item)" />
+          <template #default="{ extensionProps }">
+            <LinkDefault
+              :item="item"
+              :count="toValue(extensionProps?.count)"
+              @close="$emit('close')"
+              @select-item="$emit('selectItem', item)"
+            />
+          </template>
```
goldenmaya, line 31: This contradicts a note in the PR description, which a reviewer will read as the authority on where decorate mode reaches:

> `headerMenu` and `accountMenu` have fallback markup but their slots are unscoped, so contributed data would be discarded. Same for `main-menu.vue` within `mobileMenu`. Decorate mode stays closed for them until those slots are scoped.

But this diff scopes `main-menu.vue`. Enumerated the call sites to be sure: `mobileMenu` has exactly two — `default-menu.vue:11` and this one — and both receive `extensionProps` and bind `:count` after this PR. So the badge reaches both mobile surfaces, not one.

The `headerMenu` / `accountMenu` half of the note is accurate — `bottom-header.vue:45` and `account-navigation.vue:10` both render a bare `<LinkDefault :item="…" />` with no scoped slot, and both categories leave `Contributed` at `never` in the map, so the compile gate keeps them closed. Consistent.

Only the description needs the edit; the code is fine. Flagging it because the note understates what shipped, and the next person deciding whether to scope a slot will start from it.

client-app/core-api/contract-versioning.mjs
```diff
@@ -37,35 +38,86 @@ function majorOf(version) {
...
+
   if (removedExports.length > 0) {
-    // A removal is only acceptable once the MAJOR version has moved past the baseline;
-    // a minor/patch bump (or no bump) is not enough. NaN comparisons fail closed here,
-    // so an unparseable version still demands an explicit major bump.
-    if (majorOf(currentVersion) > majorOf(baseVersion)) {
-      return { action: "none", reason: "major already bumped" };
+    if (escapesCaretRange(currentVersion, baseVersion)) {
+      return { action: "none", reason: `${breaking} already bumped` };
     }
```
goldenmaya, line 119: Cosmetic only: `breaking` comes from `policyFor(baseVersion)`, so crossing out of the 0.x line reports the wrong level in `reason` — `0.9.0 → 1.0.0` with an export removed says `"minor already bumped"` when the human bumped a major. The decision is right, just the wording.

I probed the table rather than reading it, and everything else lands correctly:

```
additive on 0.x          -> bump-patch
breaking on 0.x          -> require-minor
breaking, minor done     -> none (minor already bumped)
breaking, patch only     -> require-minor
additive on 1.x          -> bump-minor
breaking on 1.x          -> require-major
REGRESSION released      -> require-forward-version
REGRESSION unreleased    -> require-major
0.9 -> 0.10 breaking     -> none (minor already bumped)
0.9 -> 1.0  breaking     -> none (minor already bumped)   <- says "minor", was a major
garbage version          -> require-major                  <- fails closed
```

Also traced the consumers, since the action strings all changed: `build-types.mjs` dispatches on `startsWith("require-")` / `startsWith("bump-")` with `levelOf()` peeling the suffix, and `require-forward-version` is checked explicitly at `:762` **before** the generic `require-` branch — so nothing falls through to `levelOf("require-forward-version")`. The ordering is load-bearing and correct as written.

One aside, not a defect: `satisfies` excludes prereleases by default, so `1.2.0 → 1.3.0-rc.1` with removals also reads as "bump done". Formally right for an npm range, and the facade is pinned by tarball URL anyway — mentioning it only so the behaviour is a known one rather than a surprise.

vitest.config.ts
```diff
@@ -11,6 +11,12 @@ export default defineConfig((env) =>
         environment: "jsdom",
         exclude: [...configDefaults.exclude, "client-app/e2e/*"],
         root: fileURLToPath(new URL("./", import.meta.url)),
+        // Without an explicit tsconfig, vitest spawns `tsc --noEmit` from the repo root with no
+        // `-p`, so it inherits the root tsconfig — `{"files": []}` with project references, which
+        // --noEmit does not follow. Zero files were checked and every .test-d.ts reported green
+        // regardless of content. The Build step's `vue-tsc --build` caught them; the step named
+        // after typing did not.
+        typecheck: { tsconfig: "./tsconfig.typecheck.json" },
```
goldenmaya, line 19: Confirming this one rather than questioning it, because the claim is strong enough to be worth independent proof and it holds exactly.

Premise — with `dev`'s `vitest.config.ts`, a planted `TS2322` inside `useExtensionRegistry.test-d.ts`:

```
Test Files  139 passed (139)
Type Errors  no errors
exit=0
```

Same planted error on this branch:

```
exit=1
```

And the clean run here: `139 files / 2120 tests`, `Type Errors: no errors`, exit 0 — the numbers in the PR description reproduce.

So the step really was reporting green over zero checked files, and it really checks them now. That also means the type gate this PR's central mechanism rests on (`useExtensionRegistry.test-d.ts`, `[Contributed] extends [never]`) was unverifiable by the step named after it until this commit — worth calling out, since anyone who validated that gate by running `yarn test:typing` before this got a false pass.

Only note: `tsconfig.typecheck.json` scoping to `**/*.test-d.ts` + `**/*.d.ts` leaves the 26 pre-existing repo errors unaddressed by design, which the config comment says plainly. Fine as a scope decision — just means "Typing Tests" is a narrower gate than its name suggests, and `vue-tsc --build` in the Build step stays the broad one.

client-app/modules/sales-rep/composables/useSalesRepCustomersCount.ts
```diff
@@ -8,21 +9,35 @@ import { useSalesRepHubQuery } from "./useSalesRepHubQuery";
...
-  const { result, onError } = useSalesRepHubQuery(
-    SalesRepCustomersCountDocument,
-    () => ({ storeId: globals.storeId }),
-    {
-      fetchPolicy: HUB_FETCH_POLICY,
-    },
-  );
+  // Must not throw: the shared wrapper below counts a subscriber before it runs this, so a throw
+  // leaves it holding an undefined state forever and every later surface reads `undefined`.
+  try {
```
goldenmaya, line 14: This comment is precisely right, and I want to record the proof so nobody later reads the `try/catch` as defensive noise and removes it.

`createSharedComposable` in `@vueuse/shared`:

```js
function j(e){ if(!h) return e;
  let n=0,r,i, a=()=>{--n, i&&n<=0&&(i.stop(), r=void 0, i=void 0)};
  return (...o)=>( n+=1, i||(i=effectScope(!0), r=i.run(()=>e(...o))), s(a), r );
}
```

`n += 1` happens before `i.run(...)`, and `i` is assigned before the composable runs. So a throw leaves `i` truthy with `r === undefined`, and every later call short-circuits on `i ||` and returns `undefined` — permanently, for the process. Exactly what the comment says.

Two related things I checked while I was in there, both fine:

- `s(a)` is `tryOnScopeDispose`, which returns false and registers nothing when there is no active scope. `ExtensionContribution` calls `use()` inside `scope.run(...)`, so the decrement does register — which is the reason `use()` has to be a function invoked in setup rather than a value, and it holds.
- `effectScope()` without `detached` attaches to `activeEffectScope` and `stop()` recurses into `scopes` (`@vue/reactivity` 3.5.40, ctor lines 39-50 and `stop` lines 147-152), so the child scope in `ExtensionContribution` really is disposed when the component unmounts. The comment there is accurate too.

The `catch` returning `computed(() => 0)` rather than rethrowing is the right shape given the above: the badge hides, the nav keeps working, and the shared slot is left holding a usable state instead of `undefined`.

client-app/shared/common/components/extension-point.vue
```diff
@@ -5,23 +5,116 @@
...
+
+function idOf(use: object): number {
+  const known = contributionIds.get(use);
+  if (known !== undefined) {
+    return known;
+  }
+  lastContributionId += 1;
+  contributionIds.set(use, lastContributionId);
+  return lastContributionId;
+}
```
goldenmaya, line 53: Mutation-checked this and the sibling scope, since the PR claims 8 of 8 spec mutations were caught and both of these are subtle enough to be worth an independent sample. 2 of 2 caught, each by a test named after the behaviour:

* reverting `contributionKey` to just `String(props.name)` (i.e. undoing this WeakMap identity) → `× runs the replacement when an entry is swapped under the same name in one tick`
* replacing `scope.run(...)` with a bare IIFE in `extension-contribution.vue` → `× stops the scope of a use() that threw, so its effects do not keep running`

Both from `45 files / 430 tests` in `shared/common/components` + `modules/sales-rep`, one failure each, rest green. So the specs really do pin the mechanism rather than just exercising it.

Also traced the `extension-point-list.vue` rewrite for behaviour drift, since delegating per entry changes more than it looks: slot props are added additively (an existing consumer ignoring them is unaffected), `getProps` + `$attrs` reach the component by the same merge as before, and `v-if="$slots.default"` preserves the old "no slot ⇒ render nothing" outcome. And the replace branch still gates on `isRegistered` without `passesCondition` — same as `dev`, with the call sites doing `canRender` themselves, so not a regression introduced here.

Only nit in the file: two `import/order` warnings (`:30` empty line between groups, and one in `extension-point-list.vue:29`) — `yarn lint:fix` clears them. Lint over all changed files is 0 errors / 15 warnings.

goldenmaya (commented):
Reviewed on the branch. Everything below was verified by execution rather than by reading: throwaway specs against the real Apollo cache and a real `vue-router`, mutation tests on both the PR's own specs and its config, type probes with `@ts-expect-error`, and reading `@vueuse` / `@vue/reactivity` / `vue-router` in `node_modules` for each library claim. Scratch files removed; the branch is untouched.

My own baseline, all re-run rather than taken from the description: `validate:types` exit 0 · `validate:dependencies` 0 errors / 5 warnings · `validate:core-types` passed · `test:typing --run` **139 files / 2120 tests**, exit 0 · `test:unit` 137 / 2114 green · eslint over all changed files **0 errors / 15 warnings**.

No blocker. Two findings worth fixing, both in the new plugin-facing cache API; one real coverage gap; one description edit.

**Worth fixing**
* `registerCacheTypePolicies.ts:79` — `priority` has no ceiling, so a plugin passing `101` takes a host policy over: `CouponType` moved to `{owner: "evil-plugin", priority: 101}` with 0 refusals, 0 warnings, and the cache re-keyed by `id`. The docblock promises this cannot happen, and the signature is in the published contract, so a third-party plugin can reach it. Control case (default priority) is refused correctly.
* `registerCacheTypePolicies.ts:147` — `owner` defaults to `"unknown"`, so two plugins that both omit it are the same owner: second registration wins, `rejected` empty, no warning. The same-owner exemption is right; keying it on a shared default is the hole.

**Coverage gap**
* `eslint.config.js:472` — the two `ignores` entries drop the `_internal` ban as well as the Apollo one, because the blocks below skip all of `sales-rep/**`. Probed six combinations: a type-only `_internal` import in a sales-rep spec is caught by **neither** eslint nor dependency-cruiser — which is precisely the coupling the eslint restatement was added for, in the module being extracted, in files that travel with it.

**Description**
* `main-menu.vue:24` — the reviewer note says decorate mode stays closed for `main-menu.vue`; this diff scopes it. Both `mobileMenu` call sites are scoped now. The `headerMenu` / `accountMenu` half of the note is accurate.

**Nits**
* `contract-versioning.mjs:111` — `0.9.0 → 1.0.0` reports `reason: "minor already bumped"` when the bump was a major. Wording only; the decision is right.
* two `import/order` warnings in the new extension-point files.

---

**Claims I checked and confirmed**, listing them because several are load-bearing and I did not want to relay them untested:

* **The typing-CI fix is real and matters more than it looks.** With `dev`'s `vitest.config.ts`, a planted `TS2322` in a `.test-d.ts` reports `Type Errors: no errors` and exits 0; on this branch the same plant exits 1. So the gate that this PR's central type mechanism rests on (`[Contributed] extends [never]`) was unverifiable by the step named after it until now — anyone who validated it with `yarn test:typing` beforehand got a false pass.
* **The derived peer declaration is enforced.** Removing `lodash-es` from `peerDependencies` makes `validate:core-types` exit 1 and print the exact object to paste. `create-plugin.mjs:232` merges `corePkg.peerDependencies` wholesale, optional entries included, so `lodash-es` does reach a scaffolded plugin — the failure story and the fix line up.
* **`createSharedComposable` increments its subscriber count before running the composable** and assigns the scope first, so a throw pins `undefined` for the process. The `try/catch` in `useSalesRepCustomersCount` is necessary, not defensive noise. Related: `tryOnScopeDispose` registers nothing without an active scope, which is exactly why `use()` must be a function invoked in setup.
* **`effectScope()` without `detached` attaches to the parent and `stop()` recurses** (`@vue/reactivity` 3.5.40), so `ExtensionContribution`'s child scope really is disposed on unmount.
* **The `EXTENSION_NAMES` type gate works**: `.mobileMenu` and `.productCard.nope` are compile errors, `.productCard.cardButton` is not — proven with `@ts-expect-error`, which would itself error if unused. Every existing call site uses a name present in `CUSTOM_EXTENSION_NAMES`.
* **dependency-cruiser is blind to `import type`** as described, and the eslint restatement does catch it — three probes each way.
* **The versioning table is correct** on all 12 cases I probed, including `require-forward-version` on a released regression and `require-major` on an unparseable version. Consumers dispatch on `startsWith("require-"/"bump-")` with `require-forward-version` checked first, so no action string falls through.
* **The specs pin the new mechanisms**: 2 of 2 independent mutations caught by named tests (WeakMap contribution identity; the contribution effect scope).
* **The CI limitation is stated honestly.** `yarn validate` is `validate:types` + `validate:dependencies`, and no workflow runs `yarn lint` — `theme-ci.yml` runs `check-locales`, `build`, `test:coverage`, `test:typing`. So the new boundary bans hold in the IDE and in a manual `yarn lint`, not in the pipeline, exactly as the config comment says.
* `extension-point-list.vue`'s rewrite introduces no behaviour drift I could find: slot props are additive, `getProps`/`$attrs` merge identically, `v-if="$slots.default"` preserves the no-slot outcome, and the replace branch still omits `passesCondition` just as on `dev` (call sites do `canRender`), so that is not a regression from this PR.

client-app/core/api/graphql/config/registerCacheTypePolicies.ts
ivan-kalachikov: Fixed — a caller-supplied `priority` is now clamped below `HOST_PRIORITY`, with a spec that goes red without it. Non-finite values are sanitized in the same step, since `>= NaN` was refusing nothing at all.

client-app/core/api/graphql/config/registerCacheTypePolicies.ts
ivan-kalachikov: Fixed — `owner` is required now. No `core-v*` contract is published, so tightening the signature costs no consumer.

eslint.config.js
ivan-kalachikov: Not deliberate — fixed. A trailing block restates `NO_HOST_INTERNAL_IMPORT` for both paths and the raw-Apollo waiver still holds. Your table was missing one cell: the type-only import in `useSalesRepHubQuery.ts` was uncovered too.

client-app/core-api/contract-versioning.mjs
ivan-kalachikov: Fixed — the reason now comes from `semver.diff`, so `0.9.0 → 1.0.0` reads "major already bumped". A spec had the wrong wording pinned.

client-app/shared/common/components/extension-point.vue
ivan-kalachikov: Both mutations reproduced here too. Nit fixed — it was three warnings, not two, and `--fix` could not clear them until the component imports moved into the first `<script>` block, matching `extension-contribution.vue`.

client-app/shared/layout/components/header/_internal/mobile-menu/menus/main-menu.vue
ivan-kalachikov: Description updated. One nuance: the badge reaches `default-menu.vue` only — My customers is an account-section child, while `main-menu.vue` iterates top-level `mobileMainMenuItems`. So decorate mode is open there, just unused.

vitest.config.ts
ivan-kalachikov: Reproduced both ways on the tracked file — exit 1 here, exit 0 with `dev`'s config. Thanks for proving it independently.

client-app/modules/sales-rep/composables/useSalesRepCustomersCount.ts
ivan-kalachikov: Confirmed against the `@vueuse/shared` source. One addition: the throw skips `tryOnScopeDispose` entirely, so the subscriber count never returns to 0 — the slot cannot recover at all, not just until the next unmount.

client-app/core-api/tailwind-preset.d.cts
```diff
@@ -0,0 +1,12 @@
+/**
```
goldenmaya, line 1: **Merge blocker: this file makes `yarn lint` abort, so the whole lint gate is dead on this branch.**

This is the repo's first `.cts` file. typescript-eslint's typed-rule config matches `**/*.ts,tsx,mts,cts`, but no block in `eslint.config.js` gives `.cts` a TS project, so eslint dies on the first one it reaches:

```
Error: Error while loading rule '@typescript-eslint/no-misused-promises': You have used a rule
which requires type information, but don't have parserOptions set to generate type information
for this file.
Occurred while linting .../client-app/core-api/tailwind-preset.d.cts
```

`[verified: executed]` `yarn eslint . --no-cache` at `7eac3875` → **exit 2, zero findings printed**. Same command at the merge base `36f99c1a8` → completes with `500 problems (36 errors, 464 warnings)`. It is the extension, not this file's contents: a scratch `.cts` and a scratch `.d.cts` anywhere under `client-app/` reproduce it, a scratch `.d.mts` does not (`tsconfig.app.json` includes `client-app/**/*`, so the file *is* in the program — the project service just doesn't associate `.cts`).

Two consequences worth separating. The gate itself is gone for this branch and every branch after it, and CI won't tell anyone, because `theme-ci.yml` runs `yarn build` → `validate:types` + `validate:dependencies`, not eslint. And it hid this PR's own new warnings: with the crash fixed, head reports `509 problems (36 errors, 473 warnings)` — the same 36 errors as base, plus 9 new warnings, all in files this PR adds (`registerCacheTypePolicies.ts` ×4, `extension-point-list.vue` ×2, `extension-point.vue` ×2, `extension-contribution.vue` ×1). Worth a look once lint runs again; `import/order` accounts for 3 of them, which suggests the `--fix` pass in the last round didn't reach these files.

**Fix**, `[verified: executed]` — add `.cts` to the existing `disableTypeChecked` block in `eslint.config.js` (line 521):

```js
files: ["**/*.js", "**/*.cjs", "**/*.cts"],
```

With that one string, `yarn eslint .` at head completes and reports the 509 above. I'm describing it rather than offering a `suggestion` block because that line sits outside this PR's diff hunks, so GitHub can't anchor a suggestion there.

client-app/core-api/package.json
```diff
@@ -19,7 +22,45 @@
...
+      "types": "./testing.d.mts",
+      "import": "./testing.mjs"
+    },
+    "./tailwind-preset": {
+      "types": "./tailwind-preset.d.cts",
+      "require": "./contract/tailwind-preset.cjs",
+      "default": "./contract/tailwind-preset.cjs"
+    }
+  },
+  "peerDependencies": {
```
goldenmaya, line 35: **Merge blocker: `yarn.lock` wasn't regenerated after this block landed, so the Core Facade Release workflow can't install.**

Deriving the peer declaration instead of maintaining it is a good change — but yarn records a portal workspace's peers in the lockfile, and the committed `@vc-frontend/core@portal:...` entry is unchanged from the merge base.

`[verified: executed]`

- `yarn install --immutable` at `7eac3875` → **exit 1**: `The lockfile would have been modified by this install, which is explicitly forbidden.`
- `yarn install --mode=update-lockfile` at `7eac3875` → `yarn.lock | 24 ++++`, and the 24 lines are exactly the `peerDependencies` / `peerDependenciesMeta` from this block, added under `"@vc-frontend/core@portal:./client-app/core-api::locator=..."`.
- The same command at the merge base `36f99c1a8` → no change at all, so this is a regression rather than pre-existing drift.

Why CI is green: `theme-ci.yml:97` and `storybook-ci.yml:102` both use a plain `yarn install`, which happily rewrites the lock. The only workflow that uses `--immutable` is `.github/workflows/core-facade-release.yml:95` — the *Core Facade Release* job, i.e. the one the description names as the required next step ("`core-v0.1.0` still has to be published by the Core Facade Release workflow before a plugin can install against it"). It would fail at the install step, before `validate:core-types` ever runs.

**Fix:** `yarn install` and commit the resulting `yarn.lock`.

One caveat on the rest of my review that belongs here: every green gate I report on this PR (`yarn build`, `validate:types`, `validate:core-types`, `test:typing`, full `vitest`) was measured in a worktree where I had **already applied that lockfile update**, because `yarn install` won't complete otherwise. They confirm the code, not that the tree as committed installs.

client-app/core/api/graphql/config/registerCacheTypePolicies.ts
```diff
@@ -0,0 +1,265 @@
...
+ *
+ * In development the ownership map and every refusal are readable as `window.modulesCacheDebug`.
+ */
+export function registerCacheTypePolicies(
+  policies: TypePolicies,
+  { owner, priority = 0 }: { owner: string; priority?: number },
+): void {
+  // Sanitized once, at the boundary: a non-finite priority would make every `>=` comparison in
+  // blockedBy() false and so refuse nothing at all.
+  const effectivePriority = Math.min(Number.isFinite(priority) ? priority : 0, MAX_PLUGIN_PRIORITY);
```
goldenmaya: **The clamp closes the priority route, but `owner` is the key the same-owner exemption actually turns on, and it isn't validated here.**

First, credit where it's due — I checked the clamp independently rather than taking it on trust. `[verified: executed]` At `8e30bb1` a plugin at `priority: 9999` took `MoneyType` from the host with no warning at all; at head it is refused. Both new specs go red if I replace `effectivePriority` with `priority`, and the `NaN` case is genuinely closed. That route is shut.

Two values of `owner` still walk past it, because `blockedBy` short-circuits on `by.owner !== owner` *before* it ever compares priorities.

**`"host"`.** `HOST_OWNER` is a plain string literal, so:

```ts
registerCacheTypePolicies({ CouponType: { keyFields: ["id"] } }, { owner: "host" });
```

is accepted at the default `priority: 0`, with `Logger.warn` called zero times. `cache.policies.identify({ __typename: "CouponType", id: "X", code: "C" })` moves from `CouponType:{"code":"C"}` to `CouponType:{"id":"X"}` — the cache really re-keys. It also rewrites the host's own entry to `{ owner: "host", priority: 0 }`, so an ordinary, honest plugin at `priority: 1` then takes `CouponType` as well (`CouponType:{"code":"C","id":"X"}`), again silently. The whole ladder runs with 0 warnings. Same result for a type-level claim on `CartType`, where the host holds only field policies.

**Absent.** A Module Federation plugin is plain JS at this boundary — the same reasoning `extension-contribution.vue` uses for `use()` possibly returning a Promise. Two plugins that pass `{ priority: 0 }` and no `owner` both get `undefined`, are mutually exempt, and the second silently replaces the first: `identify` flips, `rejected` stays empty, no warning. That is the shared-default hole the comment on `blockedBy` says removing the default closed — the default's *value* changed, its *existence* didn't. Related symptom from the same missing check: `registerCacheTypePolicies(policies)` with one argument now throws `TypeError: Cannot destructure property 'owner' of 'undefined'`, where at `8e30bb1` it fell back to `{}`. And a refusal against the host prints `"undefined" (priority 0) cannot take over "CouponType"`, which is the "every refusal names someone" promise this change was made for.

Validating at the same boundary the priority is sanitized closes both:

```suggestion
export function registerCacheTypePolicies(
  policies: TypePolicies,
  options: { owner: string; priority?: number },
): void {
  const { owner, priority = 0 } = options ?? {};

  // A plugin is plain JS at the boundary, and both an absent `owner` and HOST_OWNER itself slip
  // through the same-owner exemption in blockedBy().
  if (typeof owner !== "string" || !owner || owner === HOST_OWNER) {
    Logger.error(
      `registerCacheTypePolicies: "owner" is required and must not be "${HOST_OWNER}". Nothing was registered.`,
    );
    return;
  }

  // Sanitized once, at the boundary: a non-finite priority would make every `>=` comparison in
  // blockedBy() false and so refuse nothing at all.
  const effectivePriority = Math.min(Number.isFinite(priority) ? priority : 0, MAX_PLUGIN_PRIORITY);
```

`[verified: executed]` I ran exactly this diff (only the comment wording differs from what I tested): the impersonation ladder is refused end to end — the host entry stays `{ owner: "host", priority: 100 }` and `identify` stays `CouponType:{"code":"C"}` — the owner-less pair registers nothing, the one-argument call no longer throws, the legitimate `plugin-a` vs `plugin-b` collision still refuses with its warning, all 11 specs in `registerCacheTypePolicies.test.ts` stay green, and `validate:types` is clean. The only in-repo caller is `modules/sales-rep/index.ts:149` (`owner: "sales-rep"`), and the exported signature is unchanged, so `contract/index.d.ts` regenerates identically.

**One caveat, and I want to be precise about how well I checked it:** the guard can't make the docblock sentence above true, because there is a route that doesn't come through this function at all. `apolloClient` is exported from the same facade (`core-api/index.ts:68`) and is constructed with this exact `cache` instance (`client.ts:30`), so `apolloClient.cache.policies.addTypePolicies(...)` bypasses the ownership map entirely. `[reasoned — read from source, not executed]`: I confirmed the export and the shared cache instance by reading those two files, but I did **not** run that bypass, so treat it as a strong reading rather than a measurement. If it holds, the guarantee is real inside this function and advisory outside it, and the sentence at line 143 reads as absolute today.

client-app/core-api/contract-versioning.mjs
```diff
@@ -37,35 +38,89 @@ function majorOf(version) {
...
-    // A removal is only acceptable once the MAJOR version has moved past the baseline;
-    // a minor/patch bump (or no bump) is not enough. NaN comparisons fail closed here,
-    // so an unparseable version still demands an explicit major bump.
-    if (majorOf(currentVersion) > majorOf(baseVersion)) {
-      return { action: "none", reason: "major already bumped" };
+    if (escapesCaretRange(currentVersion, baseVersion)) {
+      // What the human actually DID, which is not always the level they were asked for: a bump out
+      // of the 0.x line satisfies a `require-minor` by moving the major, so reporting `breaking`
+      // here would call a 0.9.0 -> 1.0.0 release a minor.
+      return { action: "none", reason: `${diff(baseVersion, currentVersion)} already bumped` };
```
goldenmaya: **Any prerelease above the baseline reads as "the breaking bump happened", so a removal can pass this gate under a patch-level version — and the new reason string is what makes it visible.**

`semver` excludes prereleases from `^` ranges, so `escapesCaretRange` is true for *every* prerelease above the base, regardless of which component moved.

`[verified: executed]` — I imported this module at head and ran it directly, `changed: true`, `removedExports: ["VcButton"]`, `released: true`:

| base | current | result |
| --- | --- | --- |
| `1.0.0` | `1.0.1-beta.1` | `{ action: "none", reason: "prepatch already bumped" }` |
| `1.0.0` | `1.1.0-beta.1` | `{ action: "none", reason: "preminor already bumped" }` |
| `0.1.0` | `0.1.1-beta.1` | `{ action: "none", reason: "prepatch already bumped" }` |
| `1.0.0` | `1.0.1` | `{ action: "require-major", … }` ← correct, for contrast |
| `0.1.0` | `0.1.1` | `{ action: "require-minor", … }` ← correct |

So a breaking removal ships as a *patch* prerelease on a line that asks for a major, and the gate says "already bumped". A consumer on `^1.0.0` would not resolve `1.0.1-beta.1`, so the immediate range breakage is contained — but the gate's job is to force the version that signals the break, and here it accepts one that signals a patch.

**Reachable in the intended workflow, not just in theory:** `bump:core` only accepts `major|minor|patch` (`bump-version.mjs:33`), so a prerelease has to be hand-written — which is exactly the case the comment at `build-types.mjs:750` calls the "safety net for anyone who edited the version by hand". And `core-facade-release.yml:55` validates the version with `^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?$`, i.e. it explicitly admits a prerelease and would mint `core-v1.0.1-beta.1` from it.

**Blame, stated honestly:** this predates the diff — `8e30bb1` returned `action: "none"` for the same inputs too, just with the wrong-but-different wording. Raising it here because this hunk rewrote this exact line, and the level it now prints (`prepatch`) is what names the bug out loud.

**Fix direction** — gate the branch on the level that actually moved rather than on escaping the range alone, e.g. a small map of which `semver.diff` results satisfy which asked level (`major` ← `major`/`premajor`; `minor` ← `minor`/`preminor`/`major`/`premajor`), and require membership alongside `escapesCaretRange`. `[not verified by me]` — the diagnosis above I ran; this particular fix shape I did not. It came out of a reviewer pass that reports having applied it and kept all 23 tests in `contract-versioning.test.ts` green, and I'm passing that on as a suggestion to evaluate, not as a measurement. No effort estimate from me, since I didn't build it.

client-app/core/api/graphql/types.ts
```diff
@@ -443,8 +443,6 @@ export type CartType = {
   itemsCount: Scalars['Int']['output'];
   /** Quantity of items */
   itemsQuantity: Scalars['Int']['output'];
-  /** Get total points amount */
-  loyaltyPoints?: Maybe<MoneyType>;
```
goldenmaya, line 447: **The regeneration reverts a field that #2435 deliberately committed, and the description's "environment difference" framing understates that.**

The mechanism in the description is right, and I checked it rather than assuming. `[verified: executed]` `git show` on `client-app/core/api/graphql/types.ts` across revisions, looking only inside `export type CartType`:

- before `3b84728d5` ("chore: update backend packages and types", #2435) — absent
- at `3b84728d5` — present
- at the merge base `36f99c1a8` — present
- at `7eac3875` — absent

And `backend-packages.json` is untouched by this PR. So the pinned backend does have the field and `vcptcore-dev` doesn't — an environment difference, exactly as stated. What follows from that, though, is that regenerating against a non-pinned environment silently reverts another PR's committed types.

Nothing reads it today, which I also checked instead of trusting the claim: `[verified: executed]` no `.graphql` in the repo selects `cart { loyaltyPoints }`; the loyalty module reads the root field `loyaltyPointsHistory`; `LineItemType.loyaltyPoints` and `Product.loyaltyPoints` both survive this regeneration; and no module `types.ts` declares a `CartType` at all, so nothing goes out of sync. The facade contract is unaffected — `CartType` isn't in `contract/index.d.ts`.

So this is a minor, but the cost isn't zero: `core/api/graphql/types.ts` is a file every client fork merges, the next regeneration against the pinned environment will re-add these two lines, and if any future feature or fork selects `cart { loyaltyPoints }` it now fails to type-check for a reason that isn't visible from the diff.

Worth either regenerating against the environment that matches `backend-packages.json`, or keeping the two lines and saying so — I'd avoid the third option of hand-editing this file, since a hand edit here needs a whole-file, all-fragments consistency check to stay trustworthy.

client-app/core/api/graphql/config/registerCacheTypePolicies.ts
```diff
@@ -0,0 +1,265 @@
...
+ * `@vc-frontend/core` so a plugin can normalize its own GraphQL types without importing the cache.
+ *
+ * Call before the plugin issues its first query — policies do not apply retroactively to data
+ * already in the cache. Registering late is warned about in development, not refused.
+ *
+ * One claim, one owner, at the granularity Apollo merges at (see {@link claimsOf} and
+ * {@link blockedBy}). A claim held at an equal or higher priority is refused, and only that claim —
+ * the rest of the policy, and the rest of the batch, still applies. `owner` is required so a refusal
+ * names someone; pass `priority` only when a plugin is deliberately meant to outrank another. It is
+ * capped at {@link MAX_PLUGIN_PRIORITY}, so the host's own policies at {@link HOST_PRIORITY} cannot
```
goldenmaya: **Minor: the ceiling is undiscoverable from the published contract, and applying it is silent.**

This sentence is the one a plugin author reads, and it points at two names the contract doesn't ship. `[verified: executed]` `grep` for `MAX_PLUGIN_PRIORITY|HOST_PRIORITY` across the repo returns only the module-private `const`s in this file, plus this one rolled-up line in `core-api/contract/index.d.ts`; neither name appears in that file's `export { … }` list. Same for the two other targets in this docblock — `{@link claimsOf}` and `{@link blockedBy}` are also module-private, so the published JSDoc carries four dead links and never states that the ceiling is **99**.

The runtime half is silent in the matching way: `[verified: executed]` `{ owner: "loud", priority: 9999 }` against a host-held typename warns

```
registerCacheTypePolicies: "loud" (priority 99) cannot take over "MoneyType" — "host" holds it at priority 100.
```

and records `priority: 99` in `rejected`. Nothing says a cap was applied, so an author sees a number they never passed. A non-finite value becomes `0` the same way. It also has a quiet consequence between plugins: two that both pass `priority: 9999` both land on 99, where `by.priority >= priority` makes whichever registered first win — which is defensible, but not what either author asked for.

Two small things would fix both halves: state the number in this sentence rather than linking a private const ("capped at 99, one below the host's 100"), and warn once when `priority !== effectivePriority`, naming both values.

client-app/core/api/graphql/config/registerCacheTypePolicies.ts
```diff
@@ -0,0 +1,265 @@
...
+    } else {
+      const field = claim.key.slice(typename.length + 1);
+      const fieldPolicy = fields?.[field];
+      // An explicitly-undefined field policy claims nothing, so it takes no ownership either.
+      if (fieldPolicy === undefined) {
+        continue;
+      }
+      acceptedPolicy.fields = { ...acceptedPolicy.fields, [field]: fieldPolicy };
+    }
+    owners.set(claim.key, { owner, priority });
```
goldenmaya, line 160: **Minor: this `set` takes the newest priority rather than the highest, so re-registering lower hands the claim to a third plugin.**

`[verified: executed]`, against the real `InMemoryCache`:

1. `plugin-a` claims `ZzpDowngrade` at `priority: 99` → record `{ plugin-a, 99 }`
2. `plugin-a` re-registers the same policy at the default `priority: 0` → record `{ plugin-a, 0 }`
3. `plugin-b` at `priority: 50` now takes it → record `{ plugin-b, 50 }`, `identify` becomes `ZzpDowngrade:{"b":"2"}`

Step 2 is not a contrived path: it's the one the docblock on `blockedBy` explicitly blesses ("HMR and a second `init()` both land here"), and it's easy for a plugin to pass its priority on the first registration and forget it on a later one. The docblock calls a same-owner re-registration a no-op; here it isn't quite — it's a no-op for the *policy* and a downgrade for the *rank*.

Pre-existing rather than introduced by this round (identical at `8e30bb1`), and I'd have left it alone except that this PR is what documents and advertises the re-registration path.

Keeping the higher of the two priorities on the same-owner path would match what the docblock already promises.

goldenmaya (commented):
Second round, reviewed at `7eac38757ebdbc9486a2902938a57cb2ca0f7338`. Everything below was run, not read, unless a comment says otherwise — and where I only read the source I've labelled it `[reasoned]` inline rather than dressing it up.

**What I did not check, first, so nothing reads as more complete than it is.**

- **No browser walk this round.** I had the app running on my own port against the dev backend, but the walk needs an authenticated Sales Rep session and I don't type passwords. The round's delta contains no change to markup this PR authors — the mobile badge is untouched since round 1 — so the exposure is the merge interaction only: dev's new Documents nav link now sits in the same account section the badge renders from. Unverified in a browser either way.
- **Styles and polish passes not run**, per the repeat-round rule. The delta touches no styles at all, so I did not measure the `red`/`coffee` contrast matrix — there is no colour or visual change to measure.
- **Every green gate was measured with the `yarn.lock` fix already applied** (see the `core-api/package.json` comment) — `yarn install` won't complete without it. They confirm the code, not that the tree as committed installs.
- Two specific claims are labelled in place: the `apolloClient` bypass route is `[reasoned]` from reading two files, not executed; and the fix shape I pass along for the prerelease gate is `[not verified by me]`.

**A finding I withdrew, because it isn't yours.** I had `data-freshness.test.ts`'s `documents()` fixture down as a defect — it omits `displayName` and `isPinned`, both selected by `SalesRepDocumentsDocument`, which produces four Apollo "Missing field" errors per run and, more importantly, makes the `documents list` case pass even with the fetch policy mutated to `cache-first` (an incomplete cache write can never satisfy a cache-first read, so the remount refetches regardless — the case is green for the wrong reason; adding the two fields makes it correctly fail). That defect is real and I proved it three times in isolation. It is also **byte-identical at the merge base** — it came in with #2439, and this PR's only change to that file is the `useSharedSalesRepCustomersCount` rename. Wrong PR; I'll raise it against #2439 separately. Worth mentioning because the same wrong attribution is easy to make from the `8e30bb1..HEAD` range, which includes two `dev` merges.

I also have to correct myself on the mirror image of that: my first pass credited the `documents list` case as a working guard, on the strength of a whole-file run in which it went red. Run in isolation with `-t`, under the same mutation, it passes 3/3 — the earlier cases fail first, skip their `stop()`, and their module-level counters carry into it. The whole-file run was the wrong measurement.

**Prior round's findings — all four genuinely closed, and I re-verified rather than assuming:**

- *priority had no ceiling* → closed for the numeric route. At `8e30bb1` a plugin at `priority: 9999` took `MoneyType` from the host silently; at head it's refused, and both new specs go red if the clamp is removed. `NaN` closed too. (The owner route is still open — separate comment.)
- *the eslint waiver left two paths with no `_internal` ban* → closed. A probe import of a host `_internal/` component into both `useSalesRepHubQuery.ts` and a sales-rep spec is flagged at head (2 errors); with `8e30bb1`'s `eslint.config.js` and the same probes, **0** violations. Regression control both directions.
- *`0.9.0 → 1.0.0` reported the wrong level* → closed. Reverting to `policyFor(baseVersion).breaking` turns exactly one test red.
- *`owner` defaulting to `"unknown"`* → closed at the type level.

**And the PR's own central tooling claim holds, independently.** A planted `TS2322` in `useExtensionRegistry.test-d.ts` gives `yarn test:typing --run` **exit 1** at head and **exit 0** with the merge base's `vitest.config.ts`. The typing step really was checking zero files. Same for the corrected export counts: merge base 10 public names / 3 `Vc*` → head 53 / 22, nothing removed.

**Gates at head:** `validate:types` 0 · `build` 0 · `validate:core-types` 0 (including its honest `⚠ bump guard NOT enforced` line) · `test:typing` 0 (166 files / 2467 tests) · full `vitest --dir client-app` 0 (163 files / 2429 tests) · `check-locales` silent (no locale files touched) · SonarCloud gate OK with 0 unresolved issues · `yarn eslint .` **exit 2** and `yarn install --immutable` **exit 1** (the two comments above).

**Dismissed, with reasons,** so they don't come back next round: the `TS2742` pair in `useCart.ts` was an artefact of a symlinked `node_modules` (the error path literally resolved through another checkout) and is gone after a real install; the red `auto-tests (mysql)` check is infra flake — the merge base has `postgres` red instead; and the `String.raw` → template-literal change in `NO_HOST_INTERNAL_IMPORT` is behaviour-identical, since that selector contains no escapes, and the rule still fires.

The merge resolution in `modules/sales-rep/index.ts` reads correct to me: both sides survive, `registerLayoutTypePolicies` has no remaining references, and the deleted `link-my-customers-mobile.vue` has no importers.

Two blockers and one open seam; the rest is small. Nice work on the typing gate in particular — that one was worth finding.

client-app/core-api/tailwind-preset.d.cts
ivan-kalachikov: Fixed — `.cts` added to the `disableTypeChecked` block, with a comment saying why. `yarn eslint .` completes again: 509 problems, then 506 after clearing the three `import/order` warnings (`--fix` alone could not — the imports sat in the second `<script>` block). Leaving the other six: three are `sonarjs/null-dereference`, which fires 102 times repo-wide.

client-app/core-api/package.json
ivan-kalachikov: Fixed — lockfile regenerated, exactly the 24 lines you named.

client-app/core/api/graphql/config/registerCacheTypePolicies.ts
ivan-kalachikov: Fixed, with one correction: the signature change is not free. `rollup-plugin-dts` prints the parameter *pattern*, so `{ owner, priority }` → `options` rewrote `contract/index.d.ts` and `validate:core-types` went red until I regenerated it — no version bump, the export set is identical. Also ran your `apolloClient` bypass rather than leaving it reasoned: `apolloClient.cache === cache` is true and `addTypePolicies` re-keys with no ownership record, so the docblock now scopes the guarantee to this function.

client-app/core-api/contract-versioning.mjs
ivan-kalachikov: Fixed — gated on the level that actually moved (`SATISFIES` map + `escapesCaretRange`), failing closed on an unknown level. One correction to the blame: `dev` returns `require-major` for these inputs, so the caret rewrite introduced it; `8e30bb1` was already this PR. Four tests added, 27 green.

client-app/core/api/graphql/types.ts
ivan-kalachikov: Fixed — regenerated against vcst-qa, which has the field, rather than hand-editing. Three additive fields came along (`Organization.isLockedForCurrentUser`, `salesRepCustomerOrder`, `salesRepCustomerOrders`); nothing selects them and none reach the contract.

client-app/core/api/graphql/config/registerCacheTypePolicies.ts
ivan-kalachikov: Fixed, but not by keeping the higher of the two — a rank is now a property of the owner, not of the claim. Omitting `priority` inherits it, so the blessed re-registration path cannot demote anyone; passing a different value is announced and moves every claim that owner holds together. Ranks live in their own map, since a plugin can hold nothing and still have declared one.

client-app/core/api/graphql/config/registerCacheTypePolicies.ts
ivan-kalachikov: Half fixed, half declined. The docblock now states the ceiling as 99 against the host's 100 and drops the four dead `{@link}`s, so the published contract carries the number instead of private names.

Not adding the warn on `priority !== effectivePriority`: it would fire on every over-aiming registration including the ones refused straight after, so one call gets two lines, and with the number in the docblock the surprise it exists to explain is gone. Leaving this open in case you disagree.

client-app/core-api/tailwind-preset.d.cts
ivan-kalachikov: Correction to my reply above: I reverted the `import/order` part. Moving those imports out of `<script setup>` makes Sonar read the components as unused (S1128 ×2 on the same two files) — eslint's warnings are not enforced anywhere, the Sonar issues are. So the count stands at the 509 you measured; the gate fix itself is unaffected.

vas11yev1work (dismissed):

vas11yev1work (approved):

PR #2430, opened 2026-08-10
fix(VCST-5683): pluralize sales-rep count messages

ivan-kalachikov:
## Description

The **My Customers** table rendered `"1 orders"`. The call site already passed vue-i18n's plural-choice index, but `sales_rep.my_customers.table.orders_count` carried a single form (`"{count} orders"`), so there was no singular to select. The dashboard stat sub-lines (`ordered_this_month`, `new_customers`, `new_this_week`, `of_recent_orders`) had the same shape.

**What changed** — `client-app/modules/sales-rep/locales/*.json` and nothing else: 11 files, 44 lines. No component change; the plumbing was already correct.

| Forms | Locales | Why |
|---|---|---|
| one | `ja`, `zh` | Neither marks number — untouched. Same for keys that read the same either way (en `"{count} ordered this month"`, de `"{count} in diesem Monat bestellt"`). |
| two | `de`, `en`, `es`, `fi`, `it`, `no`, `sv` | The built-in selector picks between them, and their plural is what zero takes. |
| three | `fr`, `pt` | A leading **zero** slot, whose copy is the singular — see below. |
| four | `ru`, `pl` | The width at which `createIntlPluralRule` (VCST-5487, now on `dev`) consults the CLDR data. Three forms would fall back to index counting and render `"5 заказа"`. Their leading slot keeps the numeric many form, which is what both languages take at zero. |

**Agreement at zero (QA reopen).** French renders `"0 nouveau client"`, not `"0 nouveaux clients"` — it files zero under CLDR `one`. A two-form message can't express that: below four forms the rule defers to vue-i18n's locale-blind `count === 1 ? 0 : 1`, so zero can only ever land on the plural. Three forms give zero a slot of its own, which is how `reviews_count` is already written in these locales.

`pt` gets the same treatment, as QA asked. CLDR files zero under `one` for region-less `pt` and for `pt-BR`, and under `other` only for `pt-PT`. It reads correctly on the QA store purely because that store is configured `pt-PT` — switching to `pt-BR` would make `"0 novos clientes"` the same defect with no code change. One file serves both spellings and the rule short-circuits zero to slot 0 for every locale, so this picks a side: the file's wording is Brazilian (`pedido`, `Meu último pedido`), so the zero slot is singular. `pt-PT` now reads `"0 novo cliente"` where CLDR would say plural.

**Merged `dev` in.** VCST-5587 (#2433) replaced the "placed today" card with `of_recent_orders`, deleting `placed_today`; its plural forms went with the key. Everything `dev` added to these files since — `ytd` (#2429), `table.load_failed` (#2432), `of_recent_orders` (#2433) — is preserved: key sets are identical to `dev` in all 13 files, and the only values that differ are the 44 above.

`of_recent_orders` came in with that merge carrying the same defect, so it is fixed here too: a participle agreeing with the elided "orders" in `es`/`fr`/`it`/`pt`/`sv`, the genitive after `из`/`z` in `ru`/`pl`. `de`, `en`, `fi` and `no` keep one form — nothing in them agrees.

**No tests.** This is locale data; the repo has 117 locale JSONs and no tests over any of them, and `client-app/i18n.test.ts` already covers the rule these messages feed — including `rule(0, 3) === 0`, the path the `fr`/`pt` zero slot rides on. Asserting that `"{count} order | {count} orders"` renders `"1 order"` would be testing vue-i18n, not this change.

One property of the convention worth knowing, since this PR is the first to lean on it: a **Slavic** message written with exactly three forms silently loses the CLDR path and renders `"1 заказа"` with no error, because `createIntlPluralRule` only consults `Intl` at four. A repo-wide guard would belong alongside the rule, not here.

**Verification** — every message rendered through the real `createI18n` at counts 0 / 1 / 2 / 5 / 21 in all 13 locales (65 cells, no `|` leak, no raw key): `fr` `0 nouveau client`, `pt` `0 novo cliente`, `ru` `21 заказ`, `pl` `2 zamówienia`. Full `vitest` suite green (125 files / 1995 tests), `vue-tsc --build --force` clean, `eslint` and `prettier` clean.

**Note for reviewers:** the non-English forms are my best-effort translations. `ru`/`pl` would particularly benefit from a native check — `new_this_week` assumes the elided subject is the cart (корзина / koszyk), which is what the widget counts.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5683
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.56.0-pr-2430-469e-469ed074.zip


vas11yev1work (dismissed):

Aleksandra-Mitricheva:
**QA re-test on vcptcore-qa @ `2.56.0-pr-2430-f5ae-f5aec2e9` — the French zero form is fixed. One cell to fold in before merge: `pt` at zero.**

Full results on [VCST-5683](https://virtocommerce.atlassian.net/browse/VCST-5683). Everything else verified: STR 3/3 with the literal `1 order`; `orders_count` correct at 1/2/4/7/11 in en, de, fr, pt, ru, pl; all four slots of the new 4-form ru/pl messages select correctly (`0 заказов` / `1 заказ` / `2, 4 заказа` / `7, 11 заказов`); `of_recent_orders` correct at 0, 1 and 13. No console or GraphQL errors, no leaked `|`, no raw keys.

### The issue

`pt` renders **`0 novo cliente`** at zero. European Portuguese wants **`0 novos clientes`**.

`pt` got **three** forms, so it takes the rule's `formCount < 4` branch — the locale-blind `Math.min(|n|,2)` — and zero lands on slot 0, the singular. That is correct for `pt-BR` but not for `pt-PT`, and the CLDR data splits on exactly this point (ICU 77.1, the same data the browser uses):

| Locale tag | CLDR category of 0 | Correct at 0 |
| --- | --- | --- |
| `fr` | `one` | singular — so the fr fix is right |
| `pt`, `pt-BR` | `one` | singular |
| **`pt-PT`** | **`other`** | **plural** |

The store under test serves `pt-PT`, so `new_customers`, `new_this_week`, `of_recent_orders` and `ordered_this_month` are all one step off at zero. Note this is a change of behaviour: before this PR `pt` had two forms, zero fell to the plural, and it read correctly on a pt-PT store.

<details>
<summary><b>Why the two Portuguese variants differ — it is the width of the <code>one</code> bucket, not a zero special-case</b></summary>

Probing ICU 77.1 across fractional values exposes the whole rule, not just the zero cell:

| n | `pt` / `pt-BR` | `pt-PT` | `fr` | `en` |
| --- | --- | --- | --- | --- |
| 0 | `one` | `other` | `one` | `other` |
| 0.5 | `one` | `other` | `one` | `other` |
| **1** | `one` | `one` | `one` | `one` |
| 1.5 | `one` | `other` | `one` | `other` |
| 2 | `other` | `other` | `other` | `other` |

So the rules are `pt`/`pt-BR` → **`one: i = 0..1`** (the entire 0-to-1 range, hence 0, 0.5 and 1.5 as well) versus `pt-PT` → **`one: i = 1 and v = 0`** (exactly 1, nothing else). Brazilian Portuguese behaves like French here; European Portuguese behaves like English.

Linguistically this is agreement with quantities below two: Brazilian takes the singular (*0,5 metro*, *zero grau*), European the plural (*0,5 metros*, *zero graus*). CLDR carries a separate `pt_PT` ruleset deliberately for this.

To be straight about how firm each half is: `pt-PT` → `0 novos clientes` is solid on both CLDR *and* ordinary usage (*zero novos clientes*), which is why I am raising it. `pt-BR` → `0 novo cliente` is what CLDR *prescribes*, though real Brazilian UI copy often uses the plural at zero without reading as wrong — so for pt-BR the appeal is to the standard rather than to native intuition. The standard is the right yardstick because the rule in the code is itself built on `Intl.PluralRules`, i.e. on this same CLDR data: if the strings are authored against a different intuition than the selector consults, the two drift apart — which is what happened here.

</details>

### Two ways to close it

**1. Narrow — in this PR.** Give `pt` four forms with slot 0 = plural, so the `>= 4` branch's zero slot is used:

```
"new_customers": "{count} novos clientes | {count} novo cliente | {count} novos clientes | {count} novos clientes"
```

Correct for a pt-PT store, wrong for a pt-BR one — a single `pt.json` cannot serve both, because the `>= 4` branch hardcodes zero via `n === 0 ? slot0` rather than asking CLDR.

**2. Structural — in `createIntlPluralRule` (VCST-5487's code, not this PR).** Drop the zero shortcut and let CLDR select zero as well. Then `pt-PT` → `other` → plural and `pt-BR` → `one` → singular both resolve with no per-store guessing, and `pt`/`fr` just need four forms.

**The shortcut is already redundant for every locale shipped today**, so removing it should be safe — I checked each cell against ICU 77.1:

* `fr`: 0 → `one` → slot 1, the singular. Unchanged.
* `ru`/`pl`: 0 → `many` → slot 3, which is **byte-identical** to their current slot 0 (`заказов`, `zamówień`). Unchanged.
* `en`/`de`: 2 forms, never reach this branch.

Happy to re-verify whichever route you take.


[VCST-5683]: https://virtocommerce.atlassian.net/browse/VCST-5683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

yuskithedeveloper (approved):

PR #2431, opened 2026-08-10
fix(VCST-5654): recover from failed chunk loads

ivan-kalachikov:
## Description

**Deploy while someone has the tab open → lazy components break for the rest of their session, silently.**

### What happens today

1. **10:00** — a user has the storefront open and is browsing.
2. **10:05** — we deploy. Every chunk is renamed by its new content hash, so the filenames their already-loaded document knows about start returning **404**.
3. **10:06** — cart full, they click **Checkout**. `/checkout` lazy-loads `@/pages/checkout/index.vue`; that chunk 404s, the navigation dies, and they are left sitting on the cart with no message. Same for `/account/orders`, `/account/lists`, and every other lazy route.
4. They click **Checkout** again. Same nothing — `defineAsyncComponent` caches the rejected loader promise and replays it on every later mount.
5. Only a full page reload recovers, and nothing on screen tells them that.

Worst case is a lazy component that gates a control: the wishlist sharing-scope element renders an empty area and **Save stays disabled**, because the parent reads `canSave` off a component instance that never mounted.

There is no retry path today: `retry()` is only reachable through `onError`, and none of the 25 `defineAsyncComponent` call sites pass it.

### The fix — one listener

Vite dispatches `vite:preloadError` for a failed dependency preload **and** for the failed base `import()`. So a single handler covers every lazy site at once: 45 lazy routes, 25 `defineAsyncComponent` call sites, plain `import()` calls.

New `client-app/chunk-load-recovery.ts`, installed from `main.ts` before the app runner so nothing can slip past it:

| When it fails | What the user gets |
|---|---|
| App is rendered | Toast — *"Part of the site could not be loaded. Please reload the page to continue."* — with a **Reload page** button. Nothing they were typing is discarded. |
| Before the app renders (blank page) | One automatic reload. A timestamp guard stops a permanently missing chunk from looping. |

The event is left un-prevented on purpose: preventing it makes the failed import resolve with `undefined` instead of rejecting, which would hide the failure from Vue and the router.

The message says what happened, not why. A failed chunk is *usually* a deploy, but it is also a flaky network, a proxy, or a poisoned cache — so the copy states the fact and the fix rather than guessing at a cause.

### Cost and blast radius

Measured against `origin/dev` (same base commit, both built clean):

| | Base | This PR | Delta |
|---|---|---|---|
| Entry chunk, gzip | 333,964 B | 334,219 B | **+255 B (+0.08%)** |
| Entry chunk, raw | 1,850,260 B | 1,851,010 B | +750 B |
| Chunk count | 362 | 362 | 0 |
| Locale chunk (one per session) | — | — | +~150 B |

No runtime cost on the happy path: Vite already dispatched this event, nothing listened. Registration is one `addEventListener`; everything else runs only after a chunk has already failed. The module ships in exactly one chunk — no duplication.

Existing behaviour is untouched: the event is not prevented, so a failed import rejects exactly as it does today. This only adds a message and a recovery path on top.

Storage access is wrapped: `sessionStorage` throws where it is denied (sandboxed preview iframes, blocked cookies), and without a record of the previous attempt the automatic reload cannot be kept from looping — so it is skipped there rather than risking one.

### Deliberately not included

- **`onError` retry wrapper** — retrying cannot help the deploy case, the hashed filename is gone. A wrapper with no call sites is dead code.
- **Federated remotes** — they load through the Module Federation runtime, not Vite's preload helper, so they never emit this event.

### Testing

- 6 new tests in `client-app/chunk-load-recovery.test.ts`: toast when rendered, single reload when blank, no second reload inside the cooldown, reload again after it, no reload when storage is denied, event left un-prevented.
- `yarn vitest run --dir client-app` — 116 files / 1877 tests pass.
- `vue-tsc --build --force`, `eslint`, `prettier --check` on changed files — clean.
- `yarn build` — succeeds; `depcruise` reports no new violations.

Verified live against the production build (`yarn preview` + real backend), by deleting a route chunk from `dist/` while the tab was open — the actual deploy scenario:

- **Client-side navigation** — clicked a link to the missing route: navigation aborted, page stayed put, toast appeared with the **Reload page** button, no automatic reload. Console: `TypeError: Failed to fetch dynamically imported module: …/sign-in-BrTah6Dc.js`.
- **Boot-time failure** — hard-loaded the same route: `#app` stayed empty, the handler reloaded once (`performance.getEntriesByType("navigation")[0].type === "reload"`), then the guard held — no loop.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5654
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.55.0-pr-2431-14ba-14ba8d48.zip


client-app/chunk-load-recovery.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    // The event is left un-prevented: preventing it makes the import resolve with `undefined`
+    // instead of rejecting, which hides the failure from Vue and the router.
+    if (isApplicationRendered()) {
+      notifyLoadFailure();
+    } else {
+      reloadOnce();
+    }
+  });
+}
```
vas11yev1work, line 54: `window.addEventListener("vite:preloadError", ...)` is global and fires for every failed dynamic import during boot — including ones that already have their own graceful-degradation `.catch()`, e.g. the builder-preview/builder-io-preview plugins and the federated-module loader, explicitly documented as "a loader failure degrades to 'no plugins' and can never break boot").

A transient failure loading an optional plugin chunk (CDN blip, remote intentionally absent) now force-reloads the whole page instead of the app booting without that plugin, as originally designed.

Could we scope this to skip chunks that already have local recovery — e.g. call `event.preventDefault()` in those `.catch()` handlers, or otherwise exclude them from triggering a reload here?

ivan-kalachikov: Fixed in fa7ba13a5 — and it was broader than the two sites: the ui-kit and module locale loaders degrade the same way, and because they are fire-and-forget (`void loadModuleLocale(...)`) they can settle after mount, where the failure showed the user a toast for a bundle designed to fall back to English.

Two details went differently than suggested. `event.preventDefault()` inside those `.catch()` handlers can't work — `handlePreloadError` dispatches, checks `defaultPrevented` and rethrows synchronously, so the handler runs after the decision and receives the error, not the event. Instead each of the 5 sites claims its own failure via `ignoreChunkLoadFailure(error)` (Vite rethrows the same error instance it put on the event, so the match is exact, no chunk-URL matching), and the listener now decides one macrotask later so that claim lands first. A pre-mount failure additionally waits `BOOT_SETTLE_DELAY` and reloads only if nothing rendered — so a boot that finishes without the optional chunk costs no page load.

On federated remotes: those load through the MF runtime, not Vite's preload helper, so an absent or blipping remote never emits this event — only the local loader chunk fetch does, and that path is claimed now too.

vas11yev1work (approved):

PR #2432, opened 2026-08-11
fix(VCST-5682): no page-level error toast for a failed Sales Rep hub widget

ivan-kalachikov:
## Description

A failing Sales Rep statistics query rendered its inline "Couldn't load…" state **and** raised the global _"Apologies for the inconvenience. Our server is currently experiencing technical issues."_ toast, so one degraded card read as a page-level outage.

**Why the module couldn't fix this itself.** Every sales-rep composable already declares "No toast; the widget/page names the failure instead" in its own `onError` — but the toast is emitted from `errorHandlerLink`, an Apollo `onError` link in the shared chain, below the composable. Four escape hatches were checked and all dead-end:

| Attempt | Why it doesn't work |
|---|---|
| `errorPolicy: "ignore"` / `"all"` on the query | Only changes what the *observable* surfaces; the link still runs beneath it |
| Module-scoped link / second `ApolloClient` | The chain is assembled in `core/api/graphql/config/links/index.ts`, the client provided at bootstrap — both core |
| Intercepting the broadcast event | `broadcast.on` only *adds* listeners; it can't cancel core's `unhandledErrorEvent` → toast handler |
| Muting the `"GenericError"` notification group while the hub is mounted | Would also swallow genuinely page-level failures from unrelated queries (cart, user) |

So the seam has to be in core, which is what the ticket suggests. Core footprint is kept to two files.

**The change.** A per-operation opt-out carried in the Apollo operation context:

```ts
// client-app/core/api/graphql/consts.ts
export const SUPPRESS_ERROR_NOTIFICATIONS_CONTEXT: ErrorNotificationsContextType = {
  suppressErrorNotifications: true,
};
```

`errorHandlerLink` honours it in its **two generic-toast branches only** — `ServerError.Unhandled` and the unrecognized-`graphQLErrors` fallback. `Unauthorized`, `Forbidden`, `UserLocked` and `PasswordExpired` drive navigation (sign-in redirect, no-access page, `/blocked`, change-password) rather than a notification, so they still fire for an opted-out operation.

**Scope.** Set on all twelve hub reads, not just the three statistics queries named in the ticket — QA reproduced the toast on `/company/my-customers` and on the search-active path too, and every one of these composables already claims its failure inline: order/cart statistics, customer counts, my-customers list, customers-count badge, orders, top sellers, customer profile header, share-customer picker, reps list, rule discovery, saved layout.

Mutations are deliberately untouched — a failed user action still deserves a toast. That includes `saveSalesRepLayout`, which shows `saveFailed` inline as well; worth a separate UX call if the double signal there is unwanted.

### Review follow-up

Three review findings, all verified against the code before fixing:

1. **Auth outcomes under suppression.** `toServerError()` maps `code === ""` to `Unhandled` and resolves it ahead of the auth codes, so an opted-out response carrying both hit the early return and emitted nothing — contradicting this file's own comment. An opted-out operation now drops the two inputs that can only mean `Unhandled` (a network error, and the `""`-coded errors) and classifies what is left, so all four auth outcomes survive: `Unauthorized`, `Forbidden`, `UserLocked`, `PasswordExpired`. That includes a 401/403, which arrives as a network error carrying its GraphQL errors on `result` — the only shape where the two co-occur, and the one the test harness previously could not produce. With notifications on, the computation is the untouched `toServerError(networkError, graphQLErrors)`, so `dev` behaviour outside the opt-out is byte-identical.
2. **My Sales Reps had no failure signal.** The page rendered `variant="empty"` for a dead fetch and for a legitimately empty list alike, and this PR removed the fallback toast that was its only other signal. `useSalesReps` now returns `error` and the page shows the `variant="error"` failure view with a new `sales_rep.table.load_failed` (13 locales), the same shape My Customers got in VCST-5586.
3. **The opt-in was a bare constant.** All twelve reads now go through `useSalesRepHubQuery`, which bakes the context in (a caller context merges, but cannot switch the opt-out off), and `no-restricted-syntax` keeps direct `useQuery` / `useLazyQuery` out of `modules/sales-rep` — the wrapper and the tests are the only exemptions. The ambiguous mutation is resolved rather than documented: `sendCustomerCommunication` opts out too, because its modal already raises an error toast naming the channel that failed, so the generic one only stacked a second toast on one action. `saveSalesRepLayout` keeps the generic toast and now records why.

### Verification

- New `client-app/core/api/graphql/config/error-handler.test.ts` (4 tests) drives the real link: notifies on a network failure and on an unrecognized GraphQL error, stays silent for all three suppressible shapes when opted out, and still emits all four auth events when opted out. Proven red on `dev`'s handler — it emitted `unhandled_error`, `unhandled_error`, `graphql_error` where the fix emits nothing.
- New `client-app/modules/sales-rep/composables/error-notifications.test.ts` (13 tests) mounts each of the twelve hub composables against a real `ApolloClient` whose link chain is `[errorHandlerLink, alwaysFailingLink]` and asserts no broadcast emit. A thirteenth case runs the same query *without* the context and asserts the toast still fires, so the suite can't pass by never firing at all.
- `yarn vitest run --dir client-app` — 119 files, 1898 tests passed (new: `pages/sales-reps.test.ts` for the four states, `useSalesRepHubQuery.test.ts` for the option merge, plus the mixed-code auth case and the opted-out mutation in the two suites above; each new case was proven red against the code it guards)
- `yarn vue-tsc --build --force` — clean
- `yarn eslint <changed files>` — no new errors or warnings (one pre-existing `no-unnecessary-type-assertion` on `serializeError(networkError as Error)` was auto-fixed by the pre-commit hook)
- `yarn build` — clean, no new circular-dependency or chunk warnings

Not verified in a browser — reproducing needs a Sales Rep account on a live backend with one `POST /graphql` operation forced to fail.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5682
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.56.0-pr-2432-fbea-fbead07b.zip


client-app/core/api/graphql/config/error-handler.ts
```diff
@@ -1,25 +1,33 @@
...
+  // See SUPPRESS_ERROR_NOTIFICATIONS_CONTEXT: only the generic toasts are opt-out, the auth outcomes below are not.
+  const { suppressErrorNotifications } = operation.getContext() as ErrorNotificationsContextType;
 
   if (serverError !== undefined) {
-    const errorData = networkError ? serializeError(networkError as Error) : graphQLErrors;
+    if (serverError === ServerError.Unhandled && suppressErrorNotifications) {
+      return;
+    }
+
+    const errorData = networkError ? serializeError(networkError) : graphQLErrors;
```
vas11yev1work: The new `serverError === ServerError.Unhandled && suppressErrorNotifications` early-return can silently swallow `UserLocked`/`PasswordExpired` when they co-occur with an `Unhandled`-coded error in the same response, since `toServerError()` matches `Unhandled` (`extensions.code === ""`) via `.some()` before the `userLockedError`/`passwordExpired` branches are ever checked — contradicting this file's own comment that auth outcomes always apply regardless of suppression. Worth a defensive reorder (check `userLockedError`/`passwordExpired` before the suppress return) even though the trigger (mixed-code error array) is unconfirmed against the real backend.

client-app/modules/sales-rep/composables/useSalesReps.ts
```diff
@@ -33,6 +34,7 @@ export function useSalesReps() {
   const { result, loading, onError } = useQuery(CustomerSalesRepsDocument, variables, {
     keepPreviousResult: true,
     fetchPolicy: HUB_FETCH_POLICY,
+    context: SUPPRESS_ERROR_NOTIFICATIONS_CONTEXT,
```
vas11yev1work: `useSalesReps.ts` only logs on error (`onError` → `Logger.error`, "no toasts by design") and never returns a `failed`/`error` state — `sales-reps.vue`'s empty view renders the same `variant="empty"` for a real fetch failure as for a legitimately empty list. Before this diff, a failure at least surfaced the global fallback toast; now that this composable also opts into `SUPPRESS_ERROR_NOTIFICATIONS_CONTEXT`, that toast is gone too, leaving zero user-facing signal that anything failed.

client-app/core/api/graphql/consts.ts
```diff
@@ -4,3 +4,13 @@ export const WEBSOCKETS_ENDPOINT_TIMEOUT = 55 * 1000; // 55 seconds
...
+
+export type ErrorNotificationsContextType = { suppressErrorNotifications?: boolean };
+
+// Apollo operation context for an operation whose own component names the failure (an inline card error, an
+// empty view, a not-found page): the global error handler then skips the generic error toast for it. Auth
+// outcomes — sign-in redirect, locked user, expired password, no-access page — are navigation rather than
+// notification, so they still apply.
+export const SUPPRESS_ERROR_NOTIFICATIONS_CONTEXT: ErrorNotificationsContextType = {
+  suppressErrorNotifications: true,
+};
```
vas11yev1work, line 16: Consider hardening the suppress-context opt-in so it can't silently be missed. Today it's a bare constant every `useQuery` call site has to remember to pass — nothing catches a future widget that forgets it (there's already an ambiguous case in this same module: `useSalesRepCommunication.ts`'s `useMutation` has no suppress context and no comment saying whether that's intentional). One option worth considering: a small `useSalesRepHubQuery()` wrapper that bakes in context: `SUPPRESS_ERROR_NOTIFICATIONS_CONTEXT` by default, optionally paired with a lint rule restricting direct `useQuery` in `modules/sales-rep/composables` — mirroring the `no-restricted-imports` pattern this codebase already uses to force `useFetch`/`useAxios` through a wrapper.

client-app/core/api/graphql/config/error-handler.ts
ivan-kalachikov: Fixed. The suppression now skips the generic-toast branch instead of returning early, so a response that mixes an `Unhandled`-coded error with `UserLocked`/`PasswordExpired` still emits the auth event — with a test that was red before the change (it emitted nothing where it now emits `user_locked`/`password_expired`). Precedence when nothing is suppressed is untouched: `serverError` still wins over the auth branches, exactly as on `dev`.

client-app/modules/sales-rep/composables/useSalesReps.ts
ivan-kalachikov: Confirmed and fixed. `useSalesReps` returns `error` now and the page renders the `variant="error"` failure view (`sales_rep.table.load_failed`, all 13 locales), mirroring My Customers from VCST-5586. New `pages/sales-reps.test.ts` covers empty vs failed vs search-empty, including the stale-rows case `keepPreviousResult` creates.

client-app/core/api/graphql/consts.ts
ivan-kalachikov: Done. All twelve reads go through `useSalesRepHubQuery`, which bakes in the context (a caller context merges, but cannot turn the opt-out off), and `no-restricted-syntax` blocks direct `useQuery`/`useLazyQuery` in `modules/sales-rep` — wrapper and tests exempt. On the ambiguous case: `sendCustomerCommunication` opts out too, since its modal already raises an error toast naming the failed channel, so the generic one was a second toast on one action; `saveSalesRepLayout` keeps the generic toast and now says why in a comment.

client-app/core/api/graphql/config/error-handler.ts
ivan-kalachikov: Extended in `cc763c4a8`: my first pass only rescued `UserLocked`/`PasswordExpired`, but `Unauthorized`/`Forbidden` were still swallowed the same way, so the opt-out now classifies what is left once the generic errors are dropped and all four auth outcomes survive. That also covers the one shape where a network error and GraphQL errors co-occur (a 401/403 carrying `result.errors`), which the test harness now reproduces faithfully. Both new cases were red against `98e911b11`.

vas11yev1work (approved):

PR #2436, opened 2026-08-12
fix(search-bar): show only one category scope indicator while the scope is loading

muller39:
## Description

**Symptom.** On a category page, clicking a parent breadcrumb link makes the search bar's category-scope area show **two indicators at once** for the whole duration of the new category's fetch: the loading chip *and* the previous (child) category's stale chip. Only one indicator should ever be visible — the chip for the category being left should not linger until the new category's data arrives.

**Root cause.** `search-bar.vue` rendered the two indicators independently, with no mutual exclusion:

```html
<VcButton v-if="preparingScope" ... />          <!-- loading chip -->
<VcButton v-for="item in searchScopeData.searchScope" ... />  <!-- scope chips -->
```

Both conditions are true at the same time during navigation. In `category.vue`, the `watch(() => props.categoryId, ...)` handler sets `preparingScope.value = true` synchronously when `categoryId` changes, but the previous category's scope item is removed only later — by the separate `watch(() => currentCategory.value?.id, ...)` → `changeSearchBarScope()` → `clearCategoryScope()`, which runs after `fetchCategory` resolves. So for the entire fetch window `preparingScope` is `true` while the old scope item is still in `searchScope`, and both render.

**Fix.** Make the two indicators mutually exclusive in the template: the scope-item `v-for` moves into a `<template v-else>` of the existing `v-if="preparingScope"`. Template-only change — no script changes, no new imports.

**Why fixed in the template rather than by clearing the scope state earlier.** The alternative — calling `removeScopeItemByType("category")` in `category.vue` where `preparingScope` is set — looks smaller but is not behaviour-preserving. Today `changeSearchBarScope()` clears and re-adds the scope item **synchronously in one block**, so the `isCategoryScope` computed never observably changes and the `watch(isCategoryScope, ...)` in `search-bar.vue` never fires. Clearing early and re-adding only after the `await` breaks that atomicity and makes the computed flip `true → false → true`, which would fire that watch, flip `searchScopeFilterExpression` to the whole-catalog fallback mid-navigation, and flicker the placeholder. The template fix leaves `searchScopeData`, `isCategoryScope` and `searchScopeFilterExpression` semantics completely untouched.

## Update: fixes for review feedback

@ivan-kalachikov reviewed the first version and found the template-only fix left two things unaddressed. Both are now fixed:

1. **Placeholder still named the old category.** `searchPlaceholder` (`search-bar.vue`) read `isCategoryScope` alone, so during the fetch window the chip was hidden but the input placeholder still read "Search within Child category" — same stale info, different pixels. Now gated the same way as the chip: `isCategoryScope.value && !preparingScope.value`.

2. **The scope became unreachable after a failed category fetch.** `useCategory.fetchCategory` rethrows on error, and `category.vue` reset `preparingScope.value = false` only after the `await`, with no `try`/`finally`. A failed fetch left `preparingScope` stuck `true` forever — and now that the chips live in `v-else`, that meant the chips never render again while the category filter is still live, with no way for the user to clear it (the remove button lives in the now-permanently-hidden branch). Fixed by wrapping the `fetchCategory` call in `try/finally` so `preparingScope` always resets, success or failure.

3. **The tests couldn't see either bug.** The `VcInput` stub rendered only the `prepend` slot (placeholder is a separate attr, never rendered), and `SearchDropdown` was stubbed to `null` (so `filterExpression`/`isCategoryScope` props were never observable). Reworked both stubs — `VcInput`'s stub now renders the placeholder it's given, and `SearchDropdown`'s stub is a real component that surfaces its scope-related props as data attributes — and added tests asserting on both. Also fixed the "shows the scope chip once preparation finished" test, which previously never set `preparingScope` to `true`, so it only asserted the idle default rather than the actual loading → finished transition.

4. **Test file location.** Moved from `__tests__/search-bar.test.ts` to a colocated `search-bar.test.ts` next to `search-bar.vue`, matching the repo convention (111 of 116 spec files are colocated, only 4 live under `__tests__/`).

## Tests

`client-app/shared/layout/components/header/_internal/search-bar/search-bar.test.ts` (6 tests, no existing test or story modified):

- `hides the scope chips while the category scope is being prepared`
- `hides the scope chips as soon as preparation starts`
- `shows the scope chip once preparation finished` — now genuinely drives `preparingScope` `true → false` and asserts both sides of the transition
- `falls back to the generic placeholder while the category scope is being prepared` — new, covers fix #1
- `shows the category placeholder once the scope is no longer being prepared` — new
- `keeps the dropdown scoped to the category filter while preparing the next category` — new, asserts `SearchDropdown`'s `filterExpression`/`isCategoryScope` props directly instead of a `null`-rendering stub

Verified the placeholder test is genuinely red→green by reverting the `search-bar.vue` fix locally and re-running: it fails with the placeholder rendering `enter_keyword_placeholder_category` (the stale-category string) instead of the generic fallback.

## Verification

- `vitest` (updated spec): 6/6 green.
- `eslint` + `prettier --check` on all changed files: clean.
- `vue-tsc --noEmit`: 0 errors.

## Notes for reviewers

- **Not visually re-verified on a running storefront** — the change is unit-proven only. Worth a quick look at the search bar's scope area during category navigation (including a simulated fetch failure) before merge.
- `mobile-search-bar.vue` does not use `useSearchScore` or `preparingScope` (it builds a plain catalog-wide scope expression), so there is no mobile variant of this bug left unfixed.

> **DO NOT MERGE until human review.**

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.57.0-pr-2436-f3bf-f3bffc67.zip


client-app/shared/layout/components/header/_internal/search-bar/search-bar.vue
```diff
@@ -29,20 +29,22 @@
...
-          append-icon="delete-2"
-          size="xs"
-          variant="soft"
-          :aria-label="$t('shared.layout.search_bar.scope_remove_label', { label: item.label })"
-          :data-search-scope="item.label"
-          @click.stop="onScopeItemClick(item.id)"
-        >
-          {{ item.label }}
-        </VcButton>
+        <template v-else>
```
ivan-kalachikov, line 32: **The chip is gone, but the placeholder still names the old category.**

I mounted the real `VcInput` and read the DOM during the fetch window:

```
chips rendered:    0
loading chip:      1
input placeholder: "Search within Child category"
```

So you get a spinner sitting right next to text naming the category you just left. Same stale info, different pixels.

`searchPlaceholder` reads `isCategoryScope` and `getCategoriesNames()`. Both still come from the old `searchScopeData`.

Gate it the same way you gated the chip:

```ts
isCategoryScope.value && !preparingScope.value ? categoryPlaceholder : genericPlaceholder
```

client-app/shared/layout/components/header/_internal/search-bar/search-bar.vue
```diff
@@ -29,20 +29,22 @@
...
+            v-for="item in searchScopeData.searchScope"
+            :key="item.id"
+            class="search-bar__button"
+            color="secondary"
+            append-icon="delete-2"
+            size="xs"
+            variant="soft"
+            :aria-label="$t('shared.layout.search_bar.scope_remove_label', { label: item.label })"
+            :data-search-scope="item.label"
+            @click.stop="onScopeItemClick(item.id)"
```
ivan-kalachikov, line 43: **This X button becomes unreachable if the category fetch fails.**

`fetchCategory` rethrows (`useCategory.ts:36`). `category.vue:671` resets `preparingScope` *after* the await, not in a `finally`, and there is no `try`/`catch` anywhere in that file. One failed fetch leaves `preparingScope` stuck `true` forever.

Now that the chips live in `v-else`:

- before this PR: stale chip visible, user clicks X, done
- after: chip never renders, but the filter is still live

Props the dropdown actually gets while the chips are hidden:

```
filterExpression: "category.subtree:child price.USD:(0 TO) inStock_variations:true"
isCategoryScope:  true
chips:            0
```

Search stays scoped to a category nobody can see or clear.

It follows them off the page too. `onBeforeUnmount` calls `clearCategoryScope()`, which clears the scope items but not `preparingScope`, so the stuck spinner shows in the header everywhere.

Two lines: wrap the fetch in `try/finally`.

On the alternative you rejected in the description: you're right to reject it. I tested clearing the scope before the await, and it fires `watch(isCategoryScope)` twice (`true -> false -> true`), which wipes the user's typed search phrase.

client-app/shared/layout/components/header/_internal/search-bar/__tests__/search-bar.test.ts
```diff
@@ -0,0 +1,145 @@
...
+vi.mock("../../search-dropdown.vue", () => ({ default: { name: "SearchDropdown", render: () => null } }));
+vi.mock("../barcode-scanner.vue", () => ({ default: { name: "BarcodeScanner", render: () => null } }));
+
+/** Renders only the `prepend` slot — that is where the scope indicators live. */
+const VcInputStub = defineComponent({
+  name: "VcInput",
+  inheritAttrs: false,
+
+  setup(_props, { slots }) {
+    return () => h("div", { class: "input" }, slots.prepend?.());
```
ivan-kalachikov: **This stub is why the tests miss both problems above.**

It renders only the `prepend` slot, so the test can never see the placeholder. `SearchDropdown` is stubbed to `null` just above, so it can never see `filterExpression` either.

Those are the exact two places the stale category survives. The tests pass happily with both bugs in place.

Worth exposing `placeholder` on the stub and asserting the dropdown's `filterExpression` prop.

client-app/shared/layout/components/header/_internal/search-bar/__tests__/search-bar.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    expect(wrapper.findAll("[data-search-scope]")).toHaveLength(1);
+
+    preparingScope.value = true;
+    await nextTick();
+
+    expect(wrapper.findAll(LOADING_INDICATOR_SELECTOR)).toHaveLength(1);
+    expect(wrapper.findAll("[data-search-scope]")).toHaveLength(0);
+  });
+
+  it("shows the scope chip once preparation finished", () => {
```
ivan-kalachikov: Small one: this never sets `preparingScope = true`, so nothing actually finishes. It's checking the idle state.

Either rename it, or drive `true -> false` so it covers the real transition.

ivan-kalachikov (commented):
Verified this locally. The template fix does work: I confirmed the new tests are genuinely red -> green by reverting the hunk, and lint / prettier / build are clean.

But it hides one of the two stale indicators and leaves the other one visible, and it makes the failed-fetch path worse than before. Details inline.

client-app/shared/layout/components/header/_internal/search-bar/search-bar.test.ts
```diff
@@ -0,0 +1,145 @@
+import { mount } from "@vue/test-utils";
```
ivan-kalachikov, line 1: Nit: this repo keeps tests next to the file. 111 colocated `*.test.ts` vs 4 inside `__tests__` dirs. `search-bar.test.ts` beside `search-bar.vue` would match.

ivan-kalachikov (approved):

PR #2437, opened 2026-08-12
feat(VCST-5647): SRH queries optimization

yuskithedeveloper:
## Description

Storefront half of VCST-5647: the Sales Rep Hub's three statistics queries are now **shaped from the cards the rep can actually see**, instead of one hand-kept union selection serving both surfaces. Each card declares its data sources (`needs`) in `layout/stat-cards.ts`; a pure mapping turns the visible card set into `@include(if:)` flags plus a per-query enabled/disabled decision. Declare a card's `needs` and both the document and the run/don't-run decision follow — the two surfaces can no longer pay for each other's slices.

Pairs with the backend PR (money figures aggregated only when selected): <backend PR link>

**Before → after**

| | before | after |
|---|---|---|
| customer profile, order buckets | 8 — including the `week` bucket and the previous-week / previous-month baselines, for cards the profile does not have | 5 |
| dashboard, `mtd.average` | fetched, read by nothing | deleted outright |
| dashboard, `ytd.average` | fetched on both surfaces | selected only where an Avg-order-value card exists (the profile) |
| a hidden card | its slices fetched anyway | its slices drop out of the document; a query with no visible card left **does not run at all** (cart statistics and customer counts each exist for one card) |
| cart statistics selection | quantities + `currencyCode` | quantities only — with the backend PR, one grouped SQL per bucket instead of two |

### Behaviour

*   **Per-card `needs` declarations.** Tokens name whole aggregation buckets — the unit of backend cost, since the backend's batch loader runs one grouped query per distinct date range. `averageOrderValue` is the one field-level token: it rides inside the `ytd` bucket, so it is gated for bytes, not for a query.
*   **`@include(if:)` with required `Boolean!` flags, not separate per-surface documents** — a visibility-driven selection is variable at runtime, and N static documents can only express N static selections. One document also keeps the two surfaces from drifting. Precedent: `searchProductsQuery`; unlike there the flags are required, so a forgotten one is a compile error rather than a surface quietly over-fetching.
*   **One-way publish/subscribe, not provide/inject.** `useSalesRepLayout` publishes the visible card set (`useStatDataNeeds`, module-local and scope-keyed like `layout/registry.ts`); the statistics composables read it. The layout is resolved inside `<LayoutSurface>`, a child of the page that owns the composables, so it cannot be injected upwards.
*   **No new wait.** Queries hold until the layout read settles, but `salesRepLayout` is `no-cache` (~4 ms) and `<LayoutSurface>` already renders nothing but a skeleton until it lands — figures fetched earlier could not have been painted. A failed layout read settles too (registry defaults), so the cards stay fed rather than empty behind an alert. `loading` includes the not-yet-settled phase, so cards never read as zeros in the meantime.
*   **Layout-edit mode widens the set back to every card**, because the parked zone renders the hidden ones.
*   **Honest scale** (measured while researching the ticket, localhost against the perf dataset): trimming buckets alone is hygiene, not a speed-up — ~0.4 ms marginal per bucket, response payload 972 → 569 bytes, and buckets run under `Task.WhenAll`. The prize is the dropped round trips: hiding one card can delete an entire query, and up to 8 aggregations.
*   **`types.ts` is hand-updated** to match the new schema — `yarn generate:graphql-types` needs a running backend carrying the new module build. A codegen run after the backend PR lands should be a no-op.
*   **Tests**: the visible-cards → flags mapping has its own suite (`layout/stat-data-needs.test.ts`), including the exact VCST-5647 finding (profile leaves week + month-over-month out) and a guard that every registered card declares at least one source — a card nothing fetches would render a permanent zero. Sales-rep module suite: 355 passing.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5647
https://virtocommerce.atlassian.net/browse/VCST-5648?focusedCommentId=105908 (follow-up #2)

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.56.0-pr-2437-3b30-3b305f41.zip
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.56.0-pr-2437-315c-315cd30c.zip


client-app/i18n.ts
```diff
@@ -21,6 +21,47 @@ export function getDefaultNumberFormats(currency: string): IntlNumberFormat {
...
+  }
+
+  if (count === 0) {
+    return 0;
+  }
+
+  const lastDigit = count % 10;
+  const lastTwoDigits = count % 100;
+
+  if (lastDigit === 1 && lastTwoDigits !== 11) {
```
ivan-kalachikov: **Polish gets the Russian rule here — wrong for 21, 31, 41...**

Russian: a count *ending* in 1 is singular. Polish: only 1 itself is.

| count | Polish should be | this gives |
|---|---|---|
| 1 | pozycja | pozycja |
| 21 | pozycji | **pozycja** |
| 101 | pozycji | **pozycja** |

`Intl.PluralRules` agrees with this rule on every Russian count 1-120, and disagrees on 9 Polish ones: 21, 31, 41, 51, 61, 71, 81, 91, 101.

So a Polish rep with 21 cart items reads "21 pozycja".

**Fix:** keep this branch for `ru`; for `pl`, "ends in 1 but isn't 1" is *many*. Or drop the hand-rolled table and map `new Intl.PluralRules(locale).select(n)` onto the four slots.

client-app/modules/sales-rep/composables/useStatDataNeeds.ts
```diff
@@ -0,0 +1,57 @@
...
+  const ready = computed(() => Boolean(visibility.value?.settled));
+
+  const needs = computed<ReadonlySet<StatDataNeedType>>(() => {
+    const current = visibility.value;
+    if (!current?.settled) {
+      return new Set<StatDataNeedType>();
+    }
+
+    // Edit mode also renders the hidden cards, in the parked zone, so their figures are needed back.
+    return current.editing ? allStatDataNeeds(scope) : statDataNeeds(scope, current.visible);
```
ivan-kalachikov, line 53: **Toggling edit mode blanks the whole KPI row.**

Widening to every card changes the `@include` flags, so apollo sees new variables:

`applyVariables` (index.js:546) -> `restart()` -> `nextTick(stop(); start())` -> `start()` sets `loading = true` (:357).

`result` is not cleared, but `StatWidget` puts loading ahead of the value — so every order-fed card drops to "—" for a full round trip. `cache-and-network` keeps it there until the network answers.

Twice per edit session (open, then save/cancel), and only for reps who hid a card — the ones this PR is for.

**Fix:** make the pending flag per-card instead of per-query. Each card already declares `needs`, so a card is pending only when its query is in flight **and** its own slices are missing from the result. Cards that already have their data keep rendering.

client-app/modules/sales-rep/composables/useStatDataNeeds.ts
```diff
@@ -0,0 +1,57 @@
...
+ * What the statistics queries on `scope` should fetch.
+ *
+ * `ready` is false until the surface's layout has been read, and the statistics composables hold their
+ * queries until then. This is not a new wait: <LayoutSurface> already renders nothing but a skeleton
+ * until the same read lands, so figures fetched earlier could not have been painted anyway.
+ *
+ * A scope must therefore have a <LayoutSurface> mounted for it, which both hub pages do — the stat row
+ * is what LayoutSurface renders.
+ */
+export function useStatDataNeeds(scope: SalesRepLayoutScopeType) {
```
ivan-kalachikov, line 42: **No test covers this half.**

`stat-data-needs.ts` (the pure part) has a 90-line suite. This file has none, and these are the load-bearing bits:

- a query must not fire before `settled`
- edit mode must widen back to every card
- `clearStatVisibility` must make the next mount wait for its own read
- a **failed** layout read still settles, so cards stay fed

`data-freshness.test.ts` stubs it out with `showEveryCard()` and never tears it down, so both scopes stay `settled: true` for the rest of that file. A future "shouldn't fetch yet" test would pass there without testing anything.

ivan-kalachikov (commented):
Two things worth fixing before merge: the Polish plural rule (wrong for 21, 31, 41...) and the unscoped composable calls that the layout tests now break on. The rest is cleanup or a design trade.

One thing outside the code: this branch also carries the content of #2414, #2433 and #2416, all still open against `dev`. Only 3 of the 16 commits are VCST-5647 — merging this first swallows the other three.

client-app/modules/sales-rep/composables/useSalesRepLayout.ts
```diff
@@ -83,6 +84,26 @@ export function useSalesRepLayout(scope: SalesRepLayoutScopeType) {
...
+  // they are created by the PAGE, above the <LayoutSurface> that owns this layout.
+  watchEffect(() => {
+    publishStatVisibility(scope, {
+      settled: settled.value,
+      visible: visibleIn("statistics"),
+      editing: editing.value,
+    });
+  });
+
+  onScopeDispose(() => {
```
ivan-kalachikov, line 105: **This needs an owning effect scope, and `useSalesRepLayout.test.ts` has none.**

That file calls the composable bare 33 times — no `effectScope`, no mount. Vue 3.5.40 logs once per call:

```
[Vue warn] onScopeDispose() is called when there is no active effect scope to be associated with.
```

The `watchEffect` just above leaks silently as well: 33 effects that never stop, all writing into the shared `visibilityByScope` map.

**Fix**, in the test file:

```ts
let scopes: EffectScope[] = [];

function withLayout(scope: SalesRepLayoutScopeType) {
  const s = effectScope();
  scopes.push(s);
  return s.run(() => useSalesRepLayout(scope))!;
}

afterEach(() => {
  scopes.forEach((s) => s.stop());
  scopes = [];
});
```

then swap the 33 call sites. It also makes `clearStatVisibility` actually run, which is what you need to test the publish/clear path.

client-app/modules/sales-rep/api/graphql/queries/salesRepCustomerOrderStatistics/salesRepCustomerOrderStatisticsQuery.graphql
```diff
@@ -77,7 +93,8 @@ query SalesRepCustomerOrderStatistics(
       countChange
```
ivan-kalachikov: **Nothing reads `totalChange`, `totalChangePercent` or `countChange`** — not here, not in the other two comparison blocks. `countChangePercent` is the only one consumed, at 4 call sites across the two mappers.

The note above parks them pending "the product decision in VCST-5647" — the ticket this PR closes, so they can't wait on it.

Worth being clear on the cost though: the backend builds the whole `comparison` object regardless, so dropping these saves wire bytes, not queries. Cleanup, not performance.

client-app/modules/sales-rep/utils.ts
```diff
@@ -30,6 +30,11 @@ export function formatCustomerLocation(address: LocationPartsType, options?: { w
 
 // ISO date windows shared by the statistics ops: current bounds end at NOW (not calendar end);
 // previous-period windows are elapsed-matched for a fair "vs last X" delta.
```
ivan-kalachikov: **These comments now say the opposite of what the code does.**

This line says previous-period windows are "elapsed-matched". `matched()` below does a calendar move instead — "the same calendar position one period back". Same stale wording at lines 39, 44 and 49.

"current bounds end at NOW (not calendar end)" is stale too: every `To` is now end-of-day.

ivan-kalachikov (dismissed):

ivan-kalachikov (dismissed):

vkrashenko (dismissed):

vkrashenko (approved):

PR #2439, opened 2026-08-13
feat(VCST-5730): sales rep hub documents library

yuskithedeveloper:
## Description

Adds the storefront **Document Library** to the Sales Rep Hub: a dashboard widget of the latest documents plus a full **`/company/documents`** browse page (category tabs with counts, name search, a card grid, an inline details panel, and open / download). Sales Reps find and share approved sales materials without leaving the Hub. The whole surface is gated by the new `sales-rep-documents:read` permission.

Requires the backend PR (new `salesRepDocuments` / `salesRepDocument` / `salesRepDocumentCategories` queries): https://github.com/VirtoCommerce/vc-module-sales-rep/pull/12

## Behaviour

- **Dashboard widget** (`sales-rep-documents.vue`): the top documents, **pinned-first then newest**, each a row with file-type icon, name, `PDF · 96 pages · Published <date>` meta, and an Open button; a "Browse all →" link to the page. Row cap is a widget setting (default 5). Registered only when the user has `sales-rep-documents:read`, and — because the layout mounts only visible blocks — a **hidden widget issues no document request**.

- **Browse page** (`documents.vue`): a **featured document panel** at the top (the pinned document, else the newest, with a "Latest publication" badge), then **category tabs** whose counts are keyword-aware (empty categories drop out and the "All" count reflects the filter), then a **name search**, then a responsive **card grid** (preview image when available, else a large file-type icon + type badge). Selecting a card opens an **inline details panel** (no separate route), deep-linked via `?doc=`.

- **Open / Download never expose the raw asset URL**: both go through an authenticated fetch (bearer token) → Blob → object URL, so `__customer` sales-rep tokens work against the authorized endpoint. Open views in a new tab; Download saves under the real file name (display name is UI-only).

- **Permission-gated everywhere**: the widget, the page route (guard = Sales Rep access **and** `sales-rep-documents:read`), and the Hub nav link are all hidden without the permission.

- **Reuse**: the `VcFile` content-type icon set and `downloadFile` helper; the shared `SalesRepRuleChips` tabs, extended with an optional per-chip `count` (existing consumers unchanged).

- **i18n**: new `sales_rep.documents.*` keys in all 13 locales.

## Screenshots
image
image

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5730
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.57.0-pr-2439-b59f-b59f7bc8.zip


vkrashenko (commented):
Ran the gate locally on the branch: eslint clean, `validate:types` clean, `vitest run client-app/modules/sales-rep client-app/i18n.test.ts` 467 green, `check-locales` clean. The feature itself lines up with the AC — widget, browse page with pagination, permission gating on all three surfaces, preview-or-icon, 13 locales — and the hidden-widget-makes-no-request guarantee is genuinely pinned by a test.

One thing about the diff before the file comments: ~26 files / ~1.4k lines here come from #2437 and #2416, which are both still open (stat-data-needs, useStatDataNeeds, active-carts-card, stat-cards, i18n.ts, core types.ts, …). Nothing wrong with stacking, but the description doesn't say so, so a reviewer reads all of it as the document library. Worth a line in the description saying what it sits on, or retargeting the base — otherwise every change to those PRs means re-reviewing this one.

client-app/modules/sales-rep/pages/documents.vue
```diff
@@ -0,0 +1,564 @@
...
+const failed = computed(() => Boolean(error.value));
+
+// Category tabs reuse the shared rule chips: each category is a "rule" named after the subfolder,
+// with its document count rendered by the chips as a highlighted counter; the synthetic baseline
+// tab is "All" carrying the library total.
+const categoryRules = computed<SalesRepRuleType[]>(() =>
+  categories.value.map((entry) => ({ name: entry.name, label: entry.name, count: entry.count })),
+);
+
+const totalInLibrary = computed(() => categories.value.reduce((sum, entry) => sum + entry.count, 0));
```
vkrashenko: This under-counts. `GetCategoriesAsync` in the backend does `.Where(x => !string.IsNullOrEmpty(x.Category))`, so documents sitting at the library root never land in any category bucket — and the type comment here even says category is "empty when the file sits at the root". In a mixed library the All tab will show fewer than the grid actually lists.

`useSalesRepDocuments` already exposes the exact number: with no category selected, `totalCount` is the keyword-filtered library total. Capturing it sticky the same way `newestDocument` is captured would fix it without a new query — or ask the backend to return an uncategorized bucket / a total alongside the categories.

client-app/modules/sales-rep/files.ts
```diff
@@ -0,0 +1,36 @@
...
+// path instead (auth.plugin attaches the Authorization header), then open the blob's object URL.
+// Same pattern as downloadFile (@/shared/files), minus the `download` attribute so the browser
+// renders the blob instead of saving it.
+export async function openAuthorizedFile(fileUrl: string, contentType?: string): Promise<void> {
+  try {
+    const { data } = await useFetch(fileUrl).blob();
+    let blob = toValue(data) as Blob;
+
+    // Re-type when needed: a typeless blob prompts a download instead of rendering.
+    if (contentType && blob.type !== contentType) {
```
vkrashenko: This is the part I'd want to reconsider. The download endpoint returns `File(stream, contentType, document.Name)`, i.e. always `Content-Disposition: attachment` — the backend deliberately never renders these inline. Here we take the server-provided content type, stamp it onto the blob and open it in a tab, and a blob URL inherits the storefront's origin. So an `.html` or `.svg` in the library (upload only checks the platform extension allow-list) executes script same-origin with the SPA, where the access and refresh tokens live in `localStorage["auth"]`.

The threat model is admin-tier (needs `sales-rep-documents:write` to upload), so it's not urgent, but it's a cheap fix: allow-list what may render inline — pdf, png/jpeg, plain text — and send anything else through `downloadFile` instead of trusting the type we were handed.

client-app/modules/sales-rep/files.ts
```diff
@@ -0,0 +1,36 @@
...
+      blob = new Blob([blob], { type: contentType });
+    }
+
+    const url = URL.createObjectURL(blob);
+    const link = document.createElement("a");
+    link.href = url;
+    link.target = "_blank";
+    link.rel = "noopener noreferrer";
+    document.body.appendChild(link);
+    link.click();
```
vkrashenko: Two failure paths end in silence here. The synthetic click happens after an `await`, so it's outside the user gesture — Safari and Firefox tend to block that, and the rep just sees nothing happen after pressing Open. And a failed fetch (403, network) only reaches `Logger.error`, so again: nothing. `downloadFile` has the same habit, but it at least has the browser's own download UI behind it; Open has nothing.

Worth opening the tab synchronously on click and pointing it at the object URL once the blob is there, and surfacing a toast on failure. Did you get a chance to try Open in Safari?

client-app/modules/sales-rep/composables/useSalesRepPinnedDocument.ts
```diff
@@ -0,0 +1,29 @@
...
+import { computed } from "vue";
+import { Logger } from "@/core/utilities";
+import { SalesRepDocumentsDocument } from "../api/graphql/types";
+import { HUB_FETCH_POLICY } from "../constants";
+import { mapSalesRepDocument } from "./useSalesRepDocuments";
+import type { SalesRepDocumentType } from "../types";
+
+// The library's pinned document (at most one is pinned) — the browse page's default featured
+// document. `undefined` when nothing is pinned; the page then falls back to the newest document.
+export function useSalesRepPinnedDocument() {
```
vkrashenko: The browse page fires three queries on load, and this one is mostly redundant: the server's default sort (empty `sortInfos` in `ApplySort`) is already `IsPinned desc, CreatedDate desc`, so the pinned document is `documents[0]` of the unfiltered first page — exactly where `newestDocument` is captured from. The only case this covers is the pinned doc being out of view because of a filter, which is the same sticky-capture problem already solved above it.

Same root cause makes the explicit `sort: "isPinned:desc;createdDate:desc"` a no-op, and it leaves a few comments saying the server default is `createdDate:desc` (`useSalesRepDocuments.ts`, plus two test comments) — those read as promises about ordering that aren't what the backend does.

client-app/modules/sales-rep/constants.ts
```diff
@@ -82,4 +96,7 @@ export const ORDERS_MAX_ROWS = 20;
 export const TOP_SELLERS_DEFAULT_ROWS = 5;
 // The salesRepTopSellers API caps `take` at 10, so the input must not offer more.
 export const TOP_SELLERS_MAX_ROWS = 10;
+// Documents widget row cap (VCST-5730) — follows top-sellers: default 5, max 10.
+export const DOCUMENTS_DEFAULT_ROWS = 5;
```
vkrashenko, line 100: The AC asks for TOP 10 in the widget; this ships 5 by default (configurable up to 10). I think 5 is the better default in that rail, but the deviation should be in the PR description with the reasoning — the description mentions the setting, not that it departs from the AC.

client-app/modules/sales-rep/api/graphql/types.ts
```diff
@@ -961,19 +1082,22 @@ export const SaveSalesRepLayoutDocument = {"kind":"Document","definitions":[{"ki
...
...e"}}}],"selectionSet":{"kind":"SelectionSet","selections":[{"kind":"Field","name":{"kind":"Name","value":"salesRepOrderFilterRules"},"arguments":[{"kind":"Argument","name":{"kind":"Name","value":"storeId"},"value":{"kind":"Variable","name":{"kind":"Name","value":"storeId"}}},{"kind":"Argument","name":{"kind":"Name","value":"cultureName"},"value":{"kind":"Variable","name":{"kind":"Name","value":"cultureName"}}},{"kind":"Argument","name":{"kind":"Name","value":"organizationId"},"value":{"kind":"Variable","name":{"kind":"Name","value":"organizationId"}}},{"kind":"Argument","name":{"kind":"Name","value":"period"},"value":{"kind":"ObjectValue","fields":[{"kind":"ObjectField","name":{"kind":"Name","value":"from"},"value":{"kind":"Variable","name":{"kind":"Name","value":"periodFrom"}}},{"kind":"ObjectField","name":{"kind":"Name","value":"to"},"value":{"kind":"Variable","name":{"kind":"Name","value":"periodTo"}}}]}}],"selectionSet":{"kind":"SelectionSet","selections":[{"kind":"Field","name":{"kind":"Name","value":"name"}},{"kind":"Field","name":{"kind":"Name","value":"localizedName"}}]}}]}}]} as unknown as DocumentNode<SalesRepOrderFilterRulesQuery, SalesRepOrderFilterRulesQueryVariables>;
```
vkrashenko, line 1099: Codegen drift: four pre-existing documents got rewritten here (`SalesRepOrderFilterRules`, `SalesRepTopSellerFilterRules`, `SalesRepCustomerCartStatistics`, `SalesRepCustomerOrderStatistics`) — the generator dropped the empty `"directives":[]` / `"arguments":[]` entries. Semantically inert, but it's churn in a file where an unrelated change is hard to spot, and `CustomerOrderStatisticsComparison.averageChange` picked up a different description from whatever the codegen backend has deployed. Easiest is `git checkout HEAD --` on the file and re-adding only the three document-library documents; the pre-existing `*Document` constants should come out byte-identical.

client-app/modules/sales-rep/constants.ts
yuskithedeveloper: I think that, ideally, the `top-N` should be a configurable setting via the dashboard-editable layout for documents, as well as for top sellers, recent orders, and other similar blocks.
I updated the Jira task description to mention that the number should be consistent with other similar blocks.

client-app/modules/sales-rep/files.ts
```diff
@@ -0,0 +1,67 @@
...
+  // same-origin. downloadFile fetches through the same authenticated path.
+  const normalizedType = contentType?.toLowerCase();
+  if (!normalizedType || !INLINE_RENDERABLE_TYPES.has(normalizedType)) {
+    await downloadFile(fileUrl, fileName ?? "");
+    return;
+  }
+
+  // Open the tab synchronously, inside the click gesture: a tab opened after `await` is outside the
+  // gesture and Safari/Firefox block it. It stays blank until the blob is ready.
+  const tab = window.open("", "_blank");
```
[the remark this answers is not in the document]
yuskithedeveloper: **False positive — reverse tabnabbing is not reachable here.**

`window.open("", "_blank")` is intentionally called without a `noopener` feature. The tab is opened **synchronously inside the click gesture** (so Safari/Firefox don't block it as a popup) and its `location` is set only *after* the authenticated `fetch` resolves — which requires keeping the returned window handle. Passing `"noopener"` in the features argument makes `window.open` return `null`, removing that handle and breaking the flow.

The opener reference is instead severed on the very next line — `tab.opener = null` — giving the same protection `rel="noopener"` would. On top of that, the tab only ever renders a **same-origin `blob:` URL of an allow-listed inert type** (`application/pdf`, `image/png|jpeg|gif|webp`, `text/plain`); none of these execute script, so the opened document cannot read `window.opener` in the first place.

Dismissing as a false positive.

client-app/modules/sales-rep/files.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    // Re-type when needed: a typeless blob prompts a download instead of rendering.
+    if (blob.type !== contentType) {
+      blob = new Blob([blob], { type: contentType });
+    }
+
+    const url = URL.createObjectURL(blob);
+
+    if (tab) {
+      tab.location.href = url;
```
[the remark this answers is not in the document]
yuskithedeveloper: **False positive — the navigation target is a locally-minted, same-origin blob URL.**

`tab.location.href = url` navigates to `url = URL.createObjectURL(blob)`, i.e. a `blob:https://<storefront-origin>/<uuid>` value created locally from the fetched response. It is same-origin by construction and never user-controlled, so there is no open-redirect / phishing target.

The only user-derived value in this function, `fileUrl`, is the **fetch** target (a first-party API path, `/api/sales-rep/documents/{id}`) — *not* the navigation target. The tab is pointed at the resulting blob, not at `fileUrl`. The rule's same-origin sanitizers (`getReturnUrlValue` / `toSameOriginPath`) don't apply here because they normalize path-style URLs and would mangle the `blob:` scheme.

Dismissing as a false positive.

client-app/modules/sales-rep/files.ts
```diff
@@ -1,36 +1,67 @@
...
+      setTimeout(() => URL.revokeObjectURL(url), OBJECT_URL_REVOKE_DELAY_MS);
+    } else {
+      // Popup blocked: fall back to a download so the click isn't silently lost.
+      URL.revokeObjectURL(url);
+      await downloadFile(fileUrl, fileName ?? "");
+    }
   } catch (error) {
+    tab?.close();
     Logger.error(openAuthorizedFile.name, error);
+    useNotifications().error({ text: globals.i18n.global.t("sales_rep.documents.open_failed"), single: true });
```
vkrashenko: useFetch already sends HTTP errors through the global handler - 5xx raises the generic error toast, 403 pushes to NoAccess. So a failed Open shows two toasts now, while Download shows one. Same GenericError group + singleInGroup would collapse them. Might be wrong, but can recheck it ?

client-app/modules/sales-rep/api/graphql/queries/salesRepDocuments/salesRepDocumentsQuery.graphql
```diff
@@ -0,0 +1,34 @@
+query SalesRepDocuments(
+  $first: Int
+  $after: String
+  $keyword: String
+  $sort: String
+  $category: String
+  $pinned: Boolean
```
vkrashenko: dead code after removing useSalesRepPinnedDocument

client-app/modules/sales-rep/pages/documents.vue
```diff
@@ -0,0 +1,571 @@
...
+  categories.value.map((entry) => ({ name: entry.name, label: entry.name, count: entry.count })),
+);
+
+// The "All" tab count is the connection total for the current keyword (root/empty-category documents
+// included). Summing the category buckets would undercount: the backend omits empty-category
+// documents from every bucket. Captured sticky while the All tab is active (no category selected), so
+// switching to a category tab keeps the library total instead of showing a category subtotal.
+const allCount = ref(0);
+watchEffect(() => {
+  if (!category.value) {
```
vkrashenko, line 283: refreshes only while there no category is selected. Is it on purpose ?  searching with a category active updates the category counts but leaves 'All' showing the previous query's number. Possible solution is clearing the category in applyKeyword would fix it without another request

client-app/modules/sales-rep/files.ts
```diff
@@ -0,0 +1,67 @@
...
+  "image/jpeg",
+  "image/gif",
+  "image/webp",
+  "text/plain",
+]);
+
+// Views a protected library file. A plain anchor navigation carries NO bearer token (the theme has
+// no auth cookie), so the backend can't authenticate it — fetch through the authenticated path
+// (auth.plugin attaches the Authorization header), then hand the tab the blob's object URL.
+export async function openAuthorizedFile(fileUrl: string, contentType?: string, fileName?: string): Promise<void> {
```
vkrashenko: I suggest to make both required:
 contentType: string, fileName: string because the view model types them non-optional and all three call sites already pass them (tightening only fileName won't compile, a required param can't follow an optional one). Then line 30 becomes contentType.toLowerCase(), the fileName ?? "" on 32 and 60 become plain fileName, and the undefined in files.test.ts:78 becomes "".

The fallback is what bothers me: an empty download attribute on a blob URL saves the file as "download" with no extension and nothing logged - unreachable today, but only until a new call site forgets the argument.

vkrashenko (dismissed):

ivan-kalachikov (approved):

PR #2440, opened 2026-08-14
feat(VCST-4386): skyflow module

vas11yev1work:
## Description

Adds Skyflow as a payment method: a PCI-compliant tokenized card form (iframe via the `skyflow-js` SDK) at checkout/order-payment/cart, plus an account page to view and delete saved cards.

The whole feature ships as a self-contained module in `client-app/modules/skyflow` — its own GraphQL queries/mutations, composables, CVV validation, and locale files for all 13 supported languages — following `modules/README.md`'s rule that core must be able to work with any module deleted.

### How it plugs into core

- **Extension points, not hardcoded checks.** All three payment surfaces (`checkout/payment.vue`, `account/order-payment.vue`, and the cart-stage `payment.vue`) render Skyflow through `<ExtensionPointList>` against generic `paymentPage` / `orderPaymentPage` / `cartPayment` categories. `cartPayment` is new — added so the cart-stage payment step no longer needs a direct import of the module (previously the only one of the three that did). The module registers itself into all three from `modules/skyflow/index.ts`; core never references Skyflow by name.
- **Reliable feature flag.** Gating uses `useModuleSettings`, which reads from `getPageContext` (the query the app can't boot without), rather than the module-version list populated by a call that's allowed to fail silently on boot. The other 7 modules already use this pattern — Skyflow follows it instead of introducing a second, less reliable source of "is this module installed".
- **Lazy-loaded SDK.** `payment-processing-skyflow.vue` is only pulled in behind `defineAsyncComponent`, so the ~80 KB gzip `skyflow-js` bundle stays out of the eager graph shared by checkout/account routes.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-4386
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.58.0-pr-2440-d131-d1310391.zip


client-app/shared/payment/components/payment.vue
```diff
@@ -37,14 +37,14 @@ import type { IPaymentMethodParameters } from "./types";
 import PaymentProcessingAuthorizeNet from "@/shared/payment/components/payment-processing-authorize-net.vue";
 import PaymentProcessingCyberSource from "@/shared/payment/components/payment-processing-cyber-source.vue";
 
+const props = defineProps<IPaymentMethodParameters>();
+
 // Loaded only when the Skyflow method is the active payment type, so the skyflow-js SDK
 // (~80 KB gzip) stays out of the eager bundle shared across checkout/account routes.
 const PaymentProcessingSkyflow = defineAsyncComponent(
```
maksimzinchuk, line 50: Core importing a module's internal component is the one thing `modules/README.md` rules out: *"Any Module in the /modules folder can be safely removed without affecting Core"*. Three files do it:

- `shared/payment/components/payment.vue:45` (here)
- `pages/checkout/payment.vue:88`
- `pages/account/order-payment.vue:332`

Delete `client-app/modules/skyflow` and none of them compile — and the `'SkyflowPaymentMethod'` string stays hardcoded in core templates on top of that.

For this particular file there is a fair excuse: the TODO just below says the cart-payment extension point doesn't exist yet. That's worth stating explicitly in the PR description though, because the other two call sites do have one.

client-app/pages/checkout/payment.vue
```diff
@@ -85,7 +85,7 @@ import PaymentProcessingDatatrans from "@/shared/payment/components/payment-proc
 // Loaded only when the Skyflow method is the active payment type, so the skyflow-js SDK
 // (~80 KB gzip) stays out of the eager bundle shared across checkout/account routes.
 const PaymentProcessingSkyflow = defineAsyncComponent(
```
maksimzinchuk: This file already renders `ExtensionPointList` for the `paymentPage` category on L51-64, and `modules/loyalty/index.ts:71-77` registers into it exactly this way:

```ts
register("paymentPage", EXTENSION_NAMES.paymentPage.paymentMethods, {
  component: PaymentProcessingLoyalty,
  condition: renderCondition,
});
```

Registering the Skyflow component from `modules/skyflow/index.ts` with a `paymentTypeName === "SkyflowPaymentMethod"` condition would let both the hardcoded `v-else-if` branch on L30-35 and this module import go away. Same for `pages/account/order-payment.vue` and its `orderPaymentPage` extension point.

Worth noting that the extension point sits deliberately last in the `v-else-if` chain — core knows only its own built-in methods and everything else falls through. As long as Skyflow sits above it, it is both a hardcoded core dependency and a reason not to use the socket.

client-app/core/constants/modules.ts
```diff
@@ -51,3 +51,5 @@ export const LOYALTY_MODE_KEY = "Loyalty.Mode";
 export const LOYALTY_CURRENCY_KEY = "Loyalty.Currency";
 
 export const MODULE_ID_MARKETING_EXPERIENCE_API = "VirtoCommerce.MarketingExperienceApi";
+
+export const MODULE_ID_SKYFLOW = "VirtoCommerce.Skyflow";
```
maksimzinchuk: Module-specific constant living in core. Same category as `isCVVinSkyflowRequired`, which stays behind in `config/settings_data.json:37` and `core/types/theme-config.ts:463`. If the module is meant to be removable, these belong to it.

client-app/modules/skyflow/index.ts
```diff
@@ -0,0 +1,60 @@
...
+          },
+        ],
+      },
+    },
+  },
+};
+
+export function init(router: Router): void {
+  const { isAuthenticated } = useUser();
+
```
maksimzinchuk, line 64: `init` doesn't take `i18n` and the module has no `locales/` folder, unlike `loyalty`, `news` and `back-in-stock`, which each ship one and call `loadModuleLocale(i18n, "<module>")`. The keys this module uses are still in the core locales:

- `shared.account.navigation.links.saved_credit_cards` (L27)
- `shared.layout.header.mobile.account_menu.saved_credit_cards` (L40)
- `pages.account.saved_credit_cards.*` (`pages/saved-credit-cards.vue:25,50,51`)

Removing the module would leave them orphaned in core.

The gating itself looks right: `setModules` runs at `app-runner.ts:152`, well before `initSkyflow` at L261, so `hasModule` has data by then. Minor: `hasModule(MODULE_ID_SKYFLOW)` is evaluated twice in a row (L54 and L57) and could be hoisted into a local.

client-app/core/constants/modules.ts
vas11yev1work: `isCVVinSkyflowRequired` stays in `theme-config.ts`/`settings_data.json`, not in `modules/skyflow`. This file has no per-module split — it's one flat interface mirroring the whole theme config, and other modules already own fields in it the same way (`push_messages_enabled`, `quote_statuses`). PushMessages reads its theme flag directly via `themeContext.value.settings.push_messages_enabled`, same pattern Skyflow uses here. Moving just this one field would be inconsistent with the rest of the file and would need a new mechanism (e.g. declaration merging) that doesn't exist elsewhere in the codebase — disproportionate for one boolean.

`MODULE_ID_SKYFLOW` is a different case and I did move it — unlike `LOYALTY_MODULE_ID` (used outside `modules/loyalty` too, in `router/routes/main.ts`), it had zero consumers outside `modules/skyflow`, so it belongs there.

client-app/modules/skyflow/index.ts
```diff
@@ -0,0 +1,82 @@
...
+  },
+};
+
+function renderCondition({ paymentTypeName }: { paymentTypeName: string }) {
+  return paymentTypeName === SKYFLOW_PAYMENT_TYPE_NAME;
+}
+
+export function init(router: Router, i18n: I18n): void {
+  const { isAuthenticated } = useUser();
+  const isSkyflowInstalled = hasModule(MODULE_ID_SKYFLOW);
```
ivan-kalachikov, line 63: ## Skyflow can disappear from checkout, silently

This line switches Skyflow on only if `hasModule(...)` says the module is installed.

**That answer comes from a query that is allowed to fail.**

`useModules()` is filled in exactly one place — `app-runner.ts:152`:

```ts
try {
  const initialStore = await initializeApplication(domain);
  setModules(initialStore?.settings?.modules);
} catch (e) {
  Logger.warn("Failed to verify backend module versions", e);  // boot continues
}
```

That query's real job is checking module **versions** (its result feeds `notifyOutdatedModules`, `app-runner.ts:322`). It is meant to be skippable. The query that must succeed is `getPageContext`.

**So one timeout is enough:** module list stays empty → `hasModule()` returns false → Skyflow never registers.

| Screen | Before this PR | After |
|---|---|---|
| `checkout/payment.vue:44` | always worked | **empty** |
| `order-payment.vue:265` | always worked | **empty** |
| cart `payment.vue` | always worked | still works — only because of the direct import |

There is **no `v-else`** after `<ExtensionPointList>` on either page. The shopper sees the payment method logo and a blank box. No error on screen. Nothing in the log. They just can't pay.

Before this PR none of this was possible — Skyflow was hardwired into the pages.

### Fix: use the source the other modules use

All seven sibling modules (loyalty, quotes, news, customer-reviews, back-in-stock, push-messages, purchase-requests) do this instead:

```ts
const { hasModuleSettings, isEnabled } = useModuleSettings(MODULE_ID_SKYFLOW);
```

`useModuleSettings` reads `themeContext.storeSettings.modules`, which comes from `getPageContext` (`fragments/storeFields.graphql:40`) — the query that cannot fail quietly.

**It is the same module list with the same per-module settings.** The only field the other source adds is `version`.

One line, and this whole failure mode is gone.

*Same pattern already exists elsewhere for marketing-experience-api (`useNavigations.ts:90`, `router/routes/account.ts:85`) — lower stakes there, since nobody is trying to pay.*

client-app/shared/payment/components/payment.vue
```diff
@@ -37,14 +43,14 @@ import type { IPaymentMethodParameters } from "./types";
...
-  () => import("@/shared/payment/components/payment-processing-skyflow.vue"),
+  () => import("@/modules/skyflow/components/payment-processing-skyflow.vue"),
```
ivan-kalachikov, line 51: ## The module can't actually be deleted

`modules/README.md:11` promises: *"Any Module in the `/modules` folder can be safely removed without affecting Core."*

This line breaks that promise. Core imports the module.

**Tested it.** Deleted `client-app/modules/skyflow/`, removed the two `app-runner.ts` lines, ran `yarn build`:

```
error not-to-unresolvable: client-app/shared/payment/components/payment.vue → @/modules/skyflow/components/payment-processing-skyflow.vue
x 3 dependency violations (1 errors, 2 warnings)
```

**One error. This line. Nothing else.** So the rest of the extraction — routes, menu, locales, generated types, barrels — is genuinely clean. Thanks for calling this out honestly in the comment above.

Two things to add.

### 1. Nothing stops this happening again

Full `.dependency-cruiser.cjs` — 222 lines, 11 rules: `no-circular`, `no-orphans`, `no-deprecated-core`, `not-to-deprecated`, `no-non-package-json`, `not-to-unresolvable`, `no-duplicate-dep-types`, `not-to-spec`, `not-to-dev-dep`, `optional-deps-used`, `peer-deps-used`.

**None of them ban core → modules.** That is why CI is green on this import, and why the next one will also pass unnoticed.

```js
{ name: "no-core-to-module", severity: "error",
  from: { path: "^client-app/(shared|core|pages|router)/" },
  to:   { path: "^client-app/modules/" } }
```

Add an explicit allowlist entry for this file until VCST-5757. Then the debt sits somewhere CI watches, not in a comment.

### 2. It can't just reuse `paymentPage`

This component passes a different shape than the checkout pages:

| | checkout / order-payment | cart `payment.vue` |
|---|---|---|
| props | `order`, `paymentTypeName` | `order`, `cart`, `hidePaymentButton`, `disabled`, `payment` |
| events | `@success` / `@fail` | none |

It needs its own category. Six small edits:

1. `shared/common/types/extensionRegistryMap.ts:29` — add `cartPayment: ExtensionEntryType<{...}, ({ paymentTypeName }) => boolean>`
2. `shared/common/constants/initialExtensionRegistry.ts:44` — `cartPayment: {}`
3. `shared/common/constants/extensionPointsNames.ts` — the point name
4. this file — drop the import and the `v-else-if` branch, end the chain with `<ExtensionPointList v-else-if="paymentTypeName" category="cartPayment" :condition-params="{ paymentTypeName }" ... />`
5. `modules/skyflow/index.ts` — a third `register(...)`, reusing the existing `renderCondition` unchanged (it only reads `paymentTypeName`, `index.ts:57-59`)
6. import gone, boundary restored

**Worth doing for its own sake:** the ~80 KB gzip `skyflow-js` SDK leaves the core graph entirely instead of sitting in it lazily, Datatrans from the TODO lands for free, and all three payment screens finally behave the same way when the module is off. Today the cart works while checkout goes blank.

Do it now, or ship as-is **with** the rule and the exception. Shipping with neither is the option to avoid.

client-app/shared/common/constants/extensionPointsNames.ts
```diff
@@ -18,9 +18,11 @@ export const EXTENSION_NAMES = merge({}, INITIAL_EXTENSION_NAMES, {
   },
   paymentPage: {
     paymentMethods: "payment-methods",
+    skyflowPaymentMethod: "skyflow-payment-method",
```
ivan-kalachikov, line 21: ## Core should not know a module's name

`skyflowPaymentMethod` is not the name of an extension **point**. It's the name of **one module's registration slot** inside a point.

The points themselves are already generic and already belong here:

- `paymentPage` / `orderPaymentPage` — any payment module can plug in
- core owns those, correctly, in `extensionRegistryMap.ts`

The registration key just says *who* plugged in. That's the module's identity, so the module should own it.

### Core doesn't read these names

Checked every reference. After this PR, `EXTENSION_NAMES.paymentPage.*` and `.orderPaymentPage.*` are read by exactly two files, both modules:

- `modules/loyalty/index.ts:71,75`
- `modules/skyflow/index.ts:69,73`

**Zero references from core.** The base branch did have one (`pages/checkout/payment.vue:54`, a raw `'payment-methods'` string) — this PR removed it. So this PR turned the block into module-only data and added a module name to it in the same commit.

Compare `sharedList.provenanceNote`: core actually reads that one (`pages/shared-list.vue:8`). These two have no such excuse anymore.

### `paymentMethods` is misleading too

`paymentMethods: "payment-methods"` reads like the shared, canonical payment point. It isn't. It's **loyalty's** private registration key. It only looks generic because loyalty got there first.

### Why two keys exist at all

`register()` keys entries by name, and on a collision it warns and silently drops the second one (`useExtensionRegistry.ts:20-24`). So each module needs a unique key. That's a per-module concern — it doesn't need to be declared in core to work.

### Fix

Move the key next to the module that owns it:

```ts
// modules/skyflow/constants.ts
export const SKYFLOW_PAYMENT_EXTENSION = "skyflow-payment-method";
```

Same for loyalty, whenever it's touched next. Then core's `EXTENSION_NAMES` keeps only the names core actually reads, and adding a payment module stops requiring an edit in core.

ivan-kalachikov (commented):
Went through the extraction against `984972c4` — route and guard, both menu nodes, all 13 locales, generated types, barrels, and a build with the module deleted to see what core still holds on to.

The removal side is clean. Three things below: two I'd settle before merge, one structural.

`yarn build` on head: 0 errors (2 pre-existing `no-orphans` warnings). Affected vitest suites: 48/48. Module locales: 13/13 complete.

client-app/router/routes/account.ts
```diff
@@ -76,11 +75,6 @@ export const accountRoutes: RouteRecordRaw[] = [
       },
     ],
   },
-  {
-    path: "saved-credit-cards",
-    name: "SavedCreditCards",
-    component: SavedCreditCards,
-  },
   {
```
ivan-kalachikov, line 78: Leftover from removing this route: `client-app/core/composables/useNavigations.test.ts` still references it.

- `:72` — a menu fixture with `route: { name: "SavedCreditCards" }`
- `:205` — `it.each(["Profile", "Addresses", "ChangePasswordAccount", "SavedCreditCards"])`

Nothing is broken — the fixture is self-contained and all 22 tests in that file pass. But a core test now asserts against a route name core no longer owns, and the next person to read it will go looking for a route that isn't there.

Swap it for a route that still exists, or an obviously fake name.

*(Commenting here because that test file isn't part of this diff.)*

ivan-kalachikov (dismissed):

ivan-kalachikov (approved):

PR #2441, opened 2026-08-14
feat(VCST-5704): render page anchors and scroll to them on load

basilkot:
Add an anchor setting to the shared section and block schemas so a content manager can give an item a readable link target, and render it as the element id in place of the generated one.

Scroll to the hash on the initial load: Page Builder content is parsed after the route resolves, so the browser has already given up on the anchor by the time it exists in the DOM.

The slug rules mirror generateAnchor in the Page Builder designer, which offers these values in the rich text link dialog — the two sides must derive the same string or the links it produces would resolve to nothing.

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.57.0-pr-2441-a5b6-a5b6b23b.zip

---

> [!NOTE]
> **Medium Risk**
> Hash scrolling and header height CSS variables affect multiple high-traffic routes; behavior is complex but covered by new unit tests.
> 
> **Overview**
> Adds **in-page anchor links** for Page Builder blocks/sections and product template blocks: merchants can set an optional **Anchor** in block/section settings; rendered DOM ids come from `getAnchorId` (authored slug or fallback block id), aligned with the Page Builder designer’s `generateAnchor` rules.
> 
> Introduces **`scrollToAnchor`** and **`useAnchorScroll`** so `#hash` navigation works after async content mounts (static pages, Virto page builder, product page). Scroll waits for the target, respects user wheel/keyboard takeover, manages focus for a11y, and cancels stale waits on route or unmount. **VP page builder** now watches `content` instead of only parsing on mount so client-side navigations re-render and scroll at the right time.
> 
> **Layout:** desktop and mobile headers publish **`--vc-app-header-height`**, and global **`scroll-padding-top`** uses it so anchored sections aren’t hidden under the fixed header. Product **description** block template is flattened so parent `:id` falls through to the visible root.
> 
> Removes redundant `id` props from several static-content block components now that parents set `:id` via `getAnchorId`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a5b6b23b262009cac53b723a88fd669dea844781. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


client-app/shared/static-content/anchors.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  await nextTick();
+
+  const target = await waitForAnchor(id, timeoutMs, () => request === latestRequest);
+  if (!target || request !== latestRequest) {
+    return false;
+  }
+
+  target.scrollIntoView({ behavior: "smooth", block: "start" });
```
ivan-kalachikov, line 68: **The anchor lands under the header.**

`block: "start"` puts the target at y=0, but the top of the viewport is covered — `sticky top-0` on desktop (`vc-header.vue:9`), `fixed` on mobile (`mobile-header.vue:2`). There is no `scroll-margin` / `scroll-padding` anywhere in `client-app`, so the section heading the visitor deep-linked to sits behind the header.

This also hits `#` links inside Rich Text — the browser scrolls those natively, `scrollIntoView` is not involved. So `block: "center"` would not fix it, and on a section taller than the viewport it would push the heading off the top instead.

**Suggestion:** `scroll-padding-top` on the scrollport. It applies to both `scrollIntoView` and native fragment navigation, so one rule covers both paths (and the existing `#main-content` skip-link).

One catch: there is no ready value. `--vc-layout-sidebar-offset-top` (`vc-header.vue:43-53`) is written only by the desktop branch — on mobile it stays at `20px` while the fixed header is ~56px (`h-14` placeholder, `mobile-header.vue:103`).

client-app/pages/matcher/virto-pages/vp-page-builder.vue
```diff
@@ -73,6 +75,8 @@ onBeforeMount(() => {
   }
 });
 
+useAnchorScroll(() => pageBuilderContent.value);
```
ivan-kalachikov, line 82: **This readiness signal can never fire.**

`pageBuilderContent` is assigned only inside `onBeforeMount` (line 70 → `trySetContent`). There is no `watch(() => props.content)`, and `virto-pages.vue:3` renders this component without a `:key`. So after mount the getter's value never changes identity.

Verified by mounting the real component and swapping the prop:

```
after mount with page A:     PAGE-A-BLOCK
after setProps to page B:    PAGE-A-BLOCK   <- not re-parsed
after path change:           scrollToAnchor calls = 0   (gate armed)
after new content prop:      scrollToAnchor calls = 0   (still armed, forever)
```

So it is one of two, and both are bad:

- **the instance survives a path change** → `awaitingContent` stays `true` with no timeout or fallback, the deep link silently never scrolls — *and* the page renders the previous document's blocks;
- **the instance always remounts** → the gate never engages here at all, and this getter is dead code.

Either way the wiring needs a source that actually tracks this page's content — `() => props.content`, plus re-parsing it on change.

client-app/shared/static-content/composables/useAnchorScroll.ts
```diff
@@ -0,0 +1,42 @@
...
+      const movedToAnotherPage = currentPath !== undefined && currentPath !== route.path;
+      currentPath = route.path;
+
+      if (movedToAnotherPage && content) {
+        cancelAnchorScroll();
+        awaitingContent = true;
+        return;
+      }
+
+      scroll();
```
ivan-kalachikov, line 27: **A hash-only change while the gate is armed disarms it permanently.**

`currentPath` is updated on line 20, before the early return. So a hash change that arrives *while* we are still waiting for the new page's content computes `movedToAnotherPage === false` and falls through to `scroll()` here — which runs against the old page's DOM and sets `awaitingContent = false` (line 12). The real content arrival is then ignored by the watcher on line 34.

Sequence, verified against this composable (control case included):

```
/page-a#text2 -> /page-b   scrollToAnchor: []          gate armed, correct
hash -> #text9             scrollToAnchor: ["#text9"]  fires against page-A DOM
page-b content arrives     scrollToAnchor: []          gate disarmed, nothing happens

CONTROL (no hash change):
/page-a#text2 -> /page-c   scrollToAnchor: []
page-c content arrives     scrollToAnchor: ["#text9"]  correct
```

This is the same wrong-page match the gate was added to prevent, just reached from the other side. Keeping `awaitingContent` armed on a hash-only change (and re-reading the hash when the content lands) would close it.

client-app/shared/static-content/anchors.ts
```diff
@@ -0,0 +1,116 @@
...
+let latestRequest = 0;
+
+export function slugifyAnchor(value: string): string {
+  if (!value) {
+    return "";
+  }
+
+  return value
+    .toLowerCase()
+    .replace(/[^\w\s-]+/g, "")
```
ivan-kalachikov, line 38: **Only 1 of the theme's 13 locales slugifies correctly.**

`\w` without the `u` flag is `[A-Za-z0-9_]`, so anything else is deleted, not transliterated:

```
en   "Technical Specifications" -> "technical-specifications"   ok
de   "Über uns"                 -> "ber-uns"
de   "Größe & Maße"             -> "gre-mae"
fr   "Résumé détaillé"          -> "rsum-dtaill"
es   "Información"              -> "informacin"
pl   "Zamówienia"               -> "zamwienia"
sv   "Förfrågan"                -> "frfrgan"
pt   "Informações"              -> "informaes"
ru   "Спецификация"             -> ""    empty
ja   "商品情報"                  -> ""    empty
zh   "产品规格"                  -> ""    empty
```

Empty → `getAnchorId` silently falls back to the generated id (`text2`), which is never shown to the content manager, so there is no recovery path.

Also: `"a - b"` → `"a---b"` (hyphen runs are not collapsed).

---

**The question that decides severity:** does the designer's `generateAnchor` produce the exact same strings? It is not in this repo or in `node_modules`, so I cannot check, and no test can.

- **Same output** → links still resolve; ugly anchors, plus a silent fallback for ru/ja/zh. Cosmetic.
- **Different output** (transliterates `ü` → `ue`, collapses hyphens, keeps Unicode — what most slugifiers do) → **every non-English anchor is a dead link.** Blocker.

Could you paste `generateAnchor` so the table above can be diffed line by line?

client-app/shared/static-content/anchors.ts
```diff
@@ -0,0 +1,116 @@
...
+  await nextTick();
+
+  const target = await waitForAnchor(id, timeoutMs, () => request === latestRequest);
+  if (!target || request !== latestRequest) {
+    return false;
+  }
+
+  target.scrollIntoView({ behavior: "smooth", block: "start" });
+
+  return true;
```
ivan-kalachikov, line 77: **Focus is never moved to the target.**

When the browser follows a `#hash` itself it does two things: scrolls, and moves the sequential focus navigation starting point to the target. `scrollIntoView` only does the first — the viewport moves, focus stays where it was.

Here it is worse than "stays where it was": `skip-to-links.vue:24-28` focuses an empty `<span tabindex="-1">` at the top of the page on **every** `route.path` change. So after an anchor navigation focus is deterministically at the top.

- mouse: fine
- keyboard: viewport on the section, next Tab jumps back to the start of the page
- screen reader: announces the top of the page, never learns the section was reached

One line after the scroll — `target.focus({ preventScroll: true })`, adding `tabindex="-1"` if the target isn't focusable. The precedent is in the same file that causes the problem.

client-app/shared/static-content/anchors.test.ts
```diff
@@ -0,0 +1,199 @@
...
+  function addSection(id: string) {
+    const section = document.createElement("div");
+    section.id = id;
+    document.body.appendChild(section);
+    return section;
+  }
+
+  beforeEach(() => {
+    scrollIntoView = vi.fn();
+    Element.prototype.scrollIntoView = scrollIntoView;
```
ivan-kalachikov, line 60: **Stubbing on the prototype means no test can tell *which* element was scrolled.**

One shared spy for every element, so `toHaveBeenCalledOnce()` proves "a scroll happened", not "the right element scrolled".

That is exactly the bug class this PR is about: generated ids repeat (`text2` exists on the old page and the new one), and the whole readiness gate exists to stop matching the previous page. If `findAnchor` returned the stale element, the call count would still be 1 and the suite would stay green. `"lets a newer navigation supersede a wait that is still polling"` is written for this case and cannot distinguish `#stale` from `#fresh`.

Cheap fix: assert on `scrollIntoView.mock.contexts` — it holds the element.

Two more tests pass for the wrong reason:

- `"survives a malformed percent escape"` assumes `safeDecode` returns something falsy on a broken `%`. It returns the input unchanged (`common/index.ts:186-196`), so the `if (!id)` guard never fires — the test just polls 30 ms for an element named `broken%`. Right result, wrong mechanism.
- `"ignores a programmatic scroll"` dispatches `scroll`, which the code never listens to. It passes even if the entire take-over feature is deleted.

client-app/pages/product.vue
```diff
@@ -275,6 +276,9 @@ const { pushHistoricalEvent } = useHistoricalEvents();
 
 const router = useRouter();
 const route = useRoute();
+
+useAnchorScroll(() => product.value);
```
ivan-kalachikov, line 280: **Question on scope: can an authored anchor ever reach this page?**

Adding the composable here (plus `:id` on product-info's inner blocks) reads as "PDP is supported", but as far as I can trace the `anchor` field never arrives in the product template's data:

- `product.vue:154-155,367` reads the statically imported `@/config/product-default.json` / `product_b2c.json` — bundled files.
- The designer saves the edited template to `default/content/templates/product.json`, which the theme never reads.
- `builder-preview.plugin.ts:82-89`: when `templateKey` is present (product template has one), the incoming template is **not** applied — only `router.push`.

So on the PDP only generated ids resolve, and not for everything:

- work: `product-properties`, `product-description`, `product-assets`
- never get an id: `product-variations` (excluded via `BLOCKS_TO_HIDE`) and all four top-level sections — `product.vue` binds no `:id` on them

Meanwhile `_sections.json` puts **Anchor** on every section and `templates/product.json:6` lists `product-info` / `related-products` / `recommended-products`, so the designer shows the field for sections the storefront cannot honour.

Am I missing a delivery path for the product template? If not, worth stating in the PR description that the PDP is generated-ids-only, so it doesn't read as a promise.

client-app/shared/static-content/anchors.ts
```diff
@@ -0,0 +1,116 @@
...
+    .toLowerCase()
+    .replace(/[^\w\s-]+/g, "")
+    .replace(/\s+/g, "-")
+    .replace(/^-+/, "")
+    .replace(/-+$/, "");
+}
+
+export function getAnchorId(item: Record<string, unknown>): string | undefined {
+  const authored = typeof item.anchor === "string" && item.anchor.trim() ? slugifyAnchor(item.anchor) : "";
+  return authored || (typeof item.id === "string" ? item.id : undefined);
```
ivan-kalachikov, line 46: **Nothing keeps anchor ids unique or namespaced.**

`getAnchorId` is pure per item — no uniqueness check in the schema (plain string field) or at runtime — and `findAnchor` uses `getElementById`, which returns the first match in document order.

Three ways it breaks:

- two sections whose anchors slugify the same (`"Specs!"` and `"specs"`) → duplicate ids, the link hits the first
- an authored slug equal to a neighbour's generated id (`text2`)
- collision with page chrome: `"Main Content"` → `main-content`, already taken by `<main>` in `main-layout.vue:8`

---

Normally I'd say a prefix can't be added here, because the `#...` string is produced by the designer, not by the theme — renaming on one side kills every link, not just the colliding ones.

But if both sides ship together, and since this isn't released yet (no existing anchors to migrate), it's the cheapest moment to fix the contract. Worth considering as one change:

1. **Namespace the id** — a fixed prefix on every rendered anchor id isolates it from chrome ids like `main-content` / `footer`. Only works if `generateAnchor` emits the same prefix.
2. **Uniqueness check in the designer** — a prefix does not help two sections colliding with each other, and the designer is the only place that sees the whole page while editing.
3. **Unicode-aware slug** — `[^\p{L}\p{N}\s_-]` with the `u` flag turns `Über uns` into `über-uns` instead of `ber-uns`, and stops ru/ja/zh collapsing to `""`. Valid HTML5 ids, `getElementById` handles them. Fixes 12 of the theme's 13 locales (see the slugify thread).
4. **Collapse hyphen runs** — `"a - b"` currently gives `a---b`.

The part that pays off long-term: keep the rule in **one** place both sides derive from — a shared function, or at least a written spec with the same test cases on both ends — so parity stops being a verbal agreement no test can check.

client-app/shared/static-content/anchors.ts
basilkot: Here is the designer side, since it decides this — page-builder-designer/src/app/modules/integration/helpers/utils.ts:

```
export function generateAnchor(value: string): string {
  if (!value) { return generateUniqueString(10); }
  return value.toLowerCase()
    .replaceAll(/[^\w\s-]+/g, '')
    .replaceAll(/\n$/g, '')
    .replaceAll(/\s+/g, '-')
    .replaceAll(/^-+|-+$/g, '');
}
```

Same rules: ASCII \w, no transliteration, hyphen runs kept. The extra \n$ step is a no-op — \s+ turns a trailing newline into a dash and -+$ trims it. So it is your first branch: links resolve, the anchors are ugly, and ru/ja/zh fall back to the generated id — identically on both sides. getItemAnchor (anchors.helpers.ts) does the same fallback, and its spec asserts the same values ours does, Спецификация → the id included.
Which is why I would rather not change one side alone: making the theme Unicode-aware while the designer is not is exactly what turns every non-English anchor into a dead link. I made the contract explicit instead — slugifyAnchor now names the counterpart function, its file, and the mirrored specs. A Unicode-aware slug is worth doing, as one change across the two repos rather than in this PR.

client-app/shared/static-content/anchors.test.ts
basilkot: Good catch. c0537fb asserts scrollIntoView.mock.contexts everywhere it matters, so the supersede test now proves #fresh scrolled and #stale did not. The malformed-escape test was wrong for the reason you gave — safeDecode returns the input unchanged — so it is split in two: one that finds an element literally named broken%, one that matches nothing and only has to not throw.
I left "ignores a programmatic scroll" as it is on purpose: it is not there to exercise take-over, it is the regression guard for 0bb83db, where take-over was detected from scroll position and the router's own scroll reset cancelled every deep link. Re-add a scroll listener and it goes red.

client-app/shared/static-content/anchors.ts
basilkot: Fixed in 5738d7b. html { scroll-padding-top: var(--vc-header-height, 0px) } in main.scss, and the variable is now published by both branches — vc-header.vue writes the sticky wrapper's measured height, mobile-header.vue the fixed header's. As you said, scroll-padding covers native fragment navigation from rich text too, so both paths land below the header.

client-app/pages/matcher/virto-pages/vp-page-builder.vue
basilkot: You were right, and it was the first case: the instance is reused, so the previous document stayed on screen and the gate never opened. 23e8139 replaces onBeforeMount with watch(() => props.content, …, { immediate: true }), so the content is re-parsed on navigation and the getter actually changes identity. vp-page-builder.test.ts covers both halves — re-parsing on a prop swap, and the scroll firing only once the new page's content is in.

client-app/shared/static-content/composables/useAnchorScroll.ts
basilkot: Fixed in 5738d7b: the guard is now content && (movedToAnotherPage || awaitingContent), so a hash change arriving while we are still waiting keeps the gate armed, and the hash is re-read when the content lands. Your exact sequence is the test "keeps waiting when the hash changes before the new page's content is in", with the control case next to it.

client-app/shared/static-content/anchors.ts
basilkot: Added in 8e1c174: after the scroll the target gets tabindex="-1" if it is not focusable, then focus({ preventScroll: true }). c0537fb narrows it — since Tab and typing deliberately do not cancel the wait, a late match could otherwise pull focus out of search seconds after load, so the move only happens while focus is still where the page left it. Covered by "moves the keyboard starting point…", "moves focus off the span the skip links park it on" and "leaves focus alone when the visitor has moved it to a control".

client-app/pages/product.vue
basilkot: You are not missing anything, I traced the same path: product.vue reads the bundled product-default.json / product_b2c.json, default/content/templates/product.json is never fetched, and builder-preview.plugin.ts:82-89 only pushes the route when templateKey is set. An authored anchor cannot reach the PDP today.
I am keeping the composable: what it buys here is the hash landing after the product query resolves, for anchors authored inside the description and for the inner block ids. I did not put :id on the four top-level sections — their ids in the bundled template are section1/section3/section4, which the designer's own document does not have to match, so it would read as support without being it. Stated the limitation in the PR description.

client-app/shared/static-content/anchors.ts
basilkot: Agreed on the analysis, and the same constraint as in the slugify thread applies: the #... string comes out of the designer, so the theme cannot prefix or dedupe unilaterally — it would take generateAnchor + getItemAnchor + the picker, and the uniqueness check only means anything where the whole page is in hand, i.e. in the designer.
For this PR: the chrome collision (main-content) is real but narrow, and duplicate slugs resolve to the first match rather than breaking the page. I would rather do the contract change — prefix, uniqueness validation, Unicode-aware slug, hyphen collapsing — as one coordinated change across both repos, and not smuggle half of it into the theme here. Worth a follow-up ticket. What this PR does add: slugifyAnchor now points at the designer function it must match and at the specs on both ends.

client-app/shared/static-content/composables/useAnchorScroll.ts
```diff
@@ -0,0 +1,42 @@
+import { onBeforeUnmount, watch } from "vue";
+import { useRoute } from "vue-router";
+import { cancelAnchorScroll, scrollToAnchor } from "../anchors";
+
+export function useAnchorScroll(content?: () => unknown): void {
+  const route = useRoute();
+
+  let currentPath: string | undefined;
```
vas11yev1work: The manually-tracked currentPath closure variable duplicates what Vue's watch callback already provides natively as its oldValue argument.

```ts
watch(
  () => [route.path, route.hash] as const,
  ([newPath], oldValue) => {
    const [oldPath] = oldValue ?? [];
    const movedToAnotherPage = oldPath !== undefined && oldPath !== newPath;

    if (content && (movedToAnotherPage || awaitingContent)) {
      cancelAnchorScroll();
      awaitingContent = true;
      return;
    }

    scroll();
  },
  { immediate: true },
);
```

vas11yev1work (approved):

PR #2442, opened 2026-08-14
fix: batch rapid-fire cart line-item removals into a single mutation

muller39:
## Description

**Symptom.** Production Application Insights telemetry on a downstream fork (built on this theme) surfaced `DistributedLockService.ExecuteAsync` throwing `LockError: "Service is busy."` when a Redis RedLock on key `Cart:{userId}` can't be acquired within its wait window. This surfaces to the shopper as a visible `ApolloError: Service is busy.` on the Cart page. Over 30 days: ~78 server-side occurrences, ~90% on `RemoveCartItems`. One burst: a single user produced **21 failures in one minute, with 21 distinct Operation Ids** — 21 different concurrent mutations, not 21 retries of the same click. Failed requests do zero work: they just queue behind the lock for the full wait window and throw.

**Root cause.** `RemoveCartItemsCommand` already accepts `lineItemIds: string[]` — a single call can remove several line items at once — but the storefront never used that. The per-row trash icon emits one id per click:

```ts
// client-app/ui-kit/components/molecules/line-items/vc-line-items.vue
function removeSingleItem(itemId: string) {
  emit("remove:items", [itemId]);
}
```

and `removeItems()` fired the mutation directly on every invocation, with no batching. If a shopper removes several line items in the space of a second or two, each click fires its own network call, and they all race the same per-user distributed lock concurrently instead of collapsing into one call. Two other contributing layers were identified by the same RCA — an expensive per-mutation cart recalculation, and the lock's wait-window/hold-time tuning — but both are backend-side (`vc-module-x-api`) and out of scope here; this PR addresses only the storefront fan-out.

**Fix.** Add a `RemoveCartItems` target to the existing `queuedMutationsLink` (already wired into the Apollo link chain and already used for `UpdateShortCartItemQuantity`), instead of a hand-rolled debounce+abort batcher:

```ts
// client-app/core/api/graphql/config/links/queued-mutations/queued-mutations.ts
export const removeCartItemsConfig: IQueueTargetConfig<RemoveCartItemsMutationVariables> = {
  debounceMs: 1000,
  getPartitionKey: (vars) => vars.command?.cartId ?? "",
  mergeQueued: (a, b) => ({
    ...a,
    command: {
      ...a.command,
      lineItemIds: [...new Set([...(a.command?.lineItemIds ?? []), ...(b.command?.lineItemIds ?? [])])],
    },
  }),
};
```

`useCart.ts`'s `removeItems()` and `cart.vue`'s `handleRemoveItems()` go back to exactly what they were before this PR touched them — plain `await _removeItems(...)` with the original `optimisticResponse`. The batching now lives entirely in the Apollo link chain, transparent to the caller.

**Why the link over a hand-rolled batcher, not just "which one is cleaner."** An earlier version of this PR used `useMutationBatcher` (debounce + `AbortController`), reusing a pattern already in this file for quantity-change and select/unselect. Caught in review: that pattern's abort only stops the *client* from waiting on an already-dispatched request — it can't un-send it, and the server keeps processing it. A click arriving after the previous request already went out doesn't cancel that work; it queues a *second* request on top, so two removals can still run concurrently against the same lock — the exact failure mode this PR exists to fix. It also re-sends ids it already sent (`[A]`, then `[A, B]`, then `[A, B, C]`, since the batcher's buffer only clears on success, not on abort), and gets *worse*, not better, as the server slows down: it fires at the pace of clicking regardless of whether the previous request has returned, so a slower server means more requests piled up concurrently, not fewer.

`queuedMutationsLink` never cancels an in-flight request. A new click while one is out just queues; when the in-flight request settles, everything queued goes out together as one merged request. That gives three properties the batcher can't: at most one `RemoveCartItems` in flight at any click speed; no re-sent ids (each flush clears its own queue); and it eases off automatically under load — a slower server means fewer, bigger merged requests, not more concurrent ones.

It also resolves every caller's promise when the batch settles, not just the last one in a burst (each queued call gets its own `observer.next()`/`complete()`), so `await removeItems(...)` and `handleRemoveItems`'s per-click analytics work exactly as they did before batching existed — no fire-and-forget workaround needed. And because nothing gets cancelled, the original `optimisticResponse` survives untouched, with Apollo's normal automatic rollback on a failed mutation — no hand-written cache patch, no missing-rollback trade-off.

`getPartitionKey` is one addition beyond reusing the existing pattern as-is: unlike `UpdateShortCartItemQuantity` (a global singleton always scoped to the user's own cart), `removeItems()` lives in `_useFullCart(cartId?)`, which also serves the SharedCart route (viewing a specific shared cart by id). Without partitioning, the link would key its queue purely by operation name, so two calls for two different carts landing in the same debounce window could merge into a single request carrying only one of the two `cartId`s — silently applying half the removal to the wrong cart. Partitioning by `cartId` keeps the user's own cart on the default queue (empty string, same as omitting the key) while giving any other cart its own independent queue.

## Tests

Added three tests directly against the shipped `removeCartItemsConfig` (exported for this purpose) in `queued-mutations.test.ts`, each driven through a fresh, isolated controller rather than the shared production singleton:
- merges and dedupes `lineItemIds` from multiple calls on the same cart into one request (including a duplicate id sent twice),
- never merges removals from two different `cartId`s into one request,
- routes the user's own cart (no `cartId`) to the same default queue regardless of call order.

`queuedMutationsLink`'s generic debounce/merge/partition/flush mechanics were already covered by the existing 38 tests in this file; these three exist because that generic coverage says nothing about whether *this* target's own merge function and partition key actually do the right thing.

No new test for `useCart.ts`/`cart.vue` — both are back to their pre-PR code, already exercised implicitly wherever cart removal already worked before this change.

## Verification

- `eslint` on changed files — clean.
- `vue-tsc --build --force` — clean.
- `vitest run` — all passing (81 cart-composable tests + 41 in `queued-mutations.test.ts`, including the 3 new ones).
- Manually verified against a running dev instance: added multiple line items, removed them individually down to an empty cart, confirmed instant UI removal via the restored `optimisticResponse`, correct recalculated totals, and correct final cart state after a full page reload. Also confirmed per-click analytics now fire independently for each removal rather than collapsing to the last click in a burst (visible via distinct debug log entries), which the earlier batcher-based version could silently drop.
- Did **not** manage to empirically reproduce two clicks landing inside the same 1s debounce window through browser automation (tool round-trip latency between simulated clicks exceeded 1s, so each fired its own independent, correctly-handled request) — the merge/partition behavior above is verified by the new unit tests instead, not by a live repro of that specific timing window.

## Notes for reviewers

- **Not load-tested against a live distributed lock.** This removes the storefront's own contribution to the concurrency the original RCA measured (and does so more strongly than a debounce-based approach — see above); it does not by itself guarantee the production `LockError` disappears, since two other contributing layers (backend cart-recalculation cost, lock wait-window/hold-time) are explicitly out of scope and unaffected by this PR. That end-to-end validation is being tracked separately by the team that reported the original production incident.
- **Only serializes a mutation against itself.** The distributed lock is per user, not per operation, so `RemoveCartItems` can still race an `AddItem` or a quantity change for the same cart — this PR (and the queued-link approach in general) only guarantees at most one `RemoveCartItems` in flight, not at most one cart mutation of any kind.
- **No `overflowed`-equivalent UX guard.** The previous version disabled further removal input after 10 pending ids; `queuedMutationsLink` has no such cap. The closest available signal is `queuedTotal`/`hasQueuedMutations` from `useQueuedMutations()`, already used for the cross-page "you have pending changes" leave-guard. Not wired into a per-field disable here, matching the fact that the already-shipped `UpdateShortCartItemQuantity` usage doesn't have one either.

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.57.0-pr-2442-29ed-29ed39a5.zip


ivan-kalachikov:
## Use `queuedMutationsLink`, not `useMutationBatcher`

**First, the honest part.**

For normal clicking — a few trash icons within a second — both do exactly the same thing. One request. No difference.

The difference only shows up when things go wrong. Which is the only reason this PR exists.

---

**The problem with the batcher.**

When a new click arrives, the batcher cancels the request that's already out.

But cancelling only stops the browser from listening. The server never finds out. It keeps removing that item.

Then the batcher sends a new request on top.

So now two removals are running at once, fighting for the same cart lock. That's the exact thing this PR is trying to stop.

It gets worse in two ways.

**1. It re-sends items it already sent.**

The batcher only clears its list after a success. A cancel doesn't clear it.

So the requests go out as `[A]`, then `[A, B]`, then `[A, B, C]`.

Three requests where two would do. And each request is a full cart recalculation — the expensive thing your RCA already called out.

**2. It doesn't ease off when the server struggles.**

This is the big one.

The batcher goes at the pace of your clicking. It never checks whether the last request came back — it just cancels it and sends another.

So the slower the server gets, the more requests stack up on it at once.

Server at 3 seconds, clicks every 1.5 → about 2 running at the same time. Server at 10 seconds → about 6.

---

**What the link does instead.**

It doesn't cancel anything. While a request is out, new clicks just wait in line.

When the answer comes back, everything that piled up goes out together as one request.

That gives you three things:

- **Never two removals at once.** At any click speed.
- **No repeats.** It clears its list when it sends. So `[A]`, then `[B, C]`.
- **It can't stack up.** It won't send while it's still waiting. Slower server → fewer requests, bigger merges. It eases off by itself.

Same example — 3 clicks a second and a half apart, server taking 3 seconds:

- Batcher: **3 requests**, all running at once, sending `[A]` `[A,B]` `[A,B,C]`
- Link: **2 requests**, one after the other, sending `[A]` `[B,C]`

**The tradeoff:** in that case the link commits a bit later. It waits for the answer instead of firing on its own schedule. You trade a little speed for never piling up.

---

**Bonus: three of your four caveats go away.**

Nothing gets cancelled. So the optimistic response survives, and Apollo undoes it by itself if the removal fails. No hand-written cache patch. No missing rollback.

And every click that got merged still gets its own answer back. So `await removeItems()` works normally, and the hanging-promise workaround isn't needed.

---

**The whole change:**

```ts
// queued-mutations.ts
const removeCartItemsConfig: IQueueTargetConfig<RemoveCartItemsMutationVariables> = {
  debounceMs: 1000,
  mergeQueued: (a, b) => ({ ...a, command: { ...a.command,
    lineItemIds: [...new Set([...(a.command?.lineItemIds ?? []), ...(b.command?.lineItemIds ?? [])])] } }),
};
// targets: [..., createQueueTarget("RemoveCartItems", removeCartItemsConfig)]
```

Then put `useCart.removeItems` back to how it was before this PR.

And drop `CartItemsFragment`, `updateRemovalCache`, `removeItemsBatchedLoading`, `removeItemsOverflowed`, plus the `async` → sync change in `cart.vue`.

---

**Two things to know.**

There's no `overflowed` equivalent. The closest thing is `queuedTotal` from `useQueuedMutations()`.

And neither approach stops a removal from racing an add or a quantity change. The lock is per user, so both only line up a mutation against itself.

ivan-kalachikov (approved):

PR #2444, opened 2026-08-18
feat(VCST-5733): all customer orders

ivan-kalachikov:
## Description

Gives Sales Reps their customers' **complete order history** inside the hub, instead of sending them to the
buyer-facing `/account/orders`. Backed by the new `salesRepCustomerOrders` / `salesRepCustomerOrder` queries,
so the list is no longer limited to orders the rep personally placed.

Backend PR: https://github.com/VirtoCommerce/vc-module-sales-rep/pull/14

### Routes

| Route | Shows |
|---|---|
| `/company/customer-orders` | Every order of every customer the rep serves, with a **Customer** column and customer filter |
| `/company/my-customers/:organizationId/orders` | The same list scoped to one customer. Breadcrumbs `Account / Sales Rep hub / My customers / {customer} / Orders` |
| `/company/my-customers/:organizationId/orders/:orderId` | One order, **read-only** — Print is the only action |

One component serves both list routes; the customer column and filter appear only where more than one
customer is in scope.

### Behaviour

- **List** — paging (`first` + offset-as-cursor `after`), keyword search committed on Enter or the search
  button, and column sorting on Date and Total, each reversible.
- **Filters panel** — order status (multi-select, with per-status counts) and created-date (presets or custom
  range with validation). The status and customer options come back with the page as facets, so every option
  offered has orders behind it and its count follows the other filters.
- **Opening an order** — an order the **rep placed** opens on the buyer page `/account/orders/:orderId`, where
  the actions they are entitled to (Pay now, Reorder) still work. **Everyone else's** opens read-only in the
  hub. A rep can therefore see and print a colleague's order but never act on it.
- **Dashboard** — the widget is renamed **"My recent orders"** (all 13 locales) to make explicit that it, and
  the statistics beside it, remain scoped to orders the rep placed. Its "All orders" link now opens
  `/company/customer-orders`.

### Notable changes to shared code

- `shared/account/composables/useOrderView.ts` — **new**. The order-derived state (item splits, currency
  grouping, first shipment and payment, their addresses) lifted out of `useUserOrder`, which now delegates to
  it, so the same order renders identically whichever endpoint read it.
- `scripts/graphql-codegen/generator.ts` — the SalesRep entry gained `requiredCommonFragments`; codegen does
  not follow `#import` outside a module's own `apiPath`, and the read-only order page reuses the storefront's
  `fullOrderFields`.

### Known limitations

- The hub order page duplicates most of `pages/account/order-details.vue`'s template. Extracting a shared
  `OrderDetailsView` component is deferred; until then a BOPIS order's pickup-point action is missing from the
  hub page.
- The filters panel carries its own date presets rather than reusing
  `shared/account/components/date-filter-select.vue` — the shared presets are calendar-aligned and these are a
  rolling window, so adopting it is a deliberate behaviour change rather than a refactor.

### Screenshots
image

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5733
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.57.0-pr-2444-d5ba-d5ba9b69.zip


client-app/modules/sales-rep/composables/useSalesRepCustomerOrder.ts
```diff
@@ -0,0 +1,32 @@
...
+    fetchPolicy: HUB_FETCH_POLICY,
+  });
+
+  onError((err) => {
+    Logger.error("[sales-rep] salesRepCustomerOrder failed:", err);
+  });
+
+  const order = computed(() => result.value?.salesRepCustomerOrder as CustomerOrderType | undefined);
+
+  const notFound = computed(() => !loading.value && !order.value);
```
vkrashenko: A failed read arrives here as `notFound`, and since `useSalesRepHubQuery` opts out of the global toast, a 500 or a 403 shows the rep "Order not found" plus a Back button — with no other signal anywhere in the UI.

`useSalesRepCustomer` already draws that line (`enabled && !loading && !failed && !customer`) and `customer-profile.vue` spells out why: a server error must not read as "this customer isn't yours". The list side of this PR does it right too, with the `variant="error"` empty view. Suggest returning `failed` from here as well, making this `!loading && !failed && !order`, and giving the details page its own failure view.

Related, same file's caller: `useSalesRepCustomerOrders` takes only `loading` and `notFound` from `useSalesRepCustomer`, so if the customer lookup fails while the orders read succeeds, the heading quietly falls back to `title_fallback` and nothing is reported.

client-app/modules/sales-rep/components/sales-rep-orders-filters.vue
```diff
@@ -0,0 +1,321 @@
...
+
+function emptyFilter(): SalesRepOrdersFilterDataType {
+  return { statuses: [], customerNames: [], startDate: undefined, endDate: undefined };
+}
+
+function cloneFilter(value: SalesRepOrdersFilterDataType): SalesRepOrdersFilterDataType {
+  return { ...value, statuses: [...value.statuses], customerNames: [...(value.customerNames ?? [])] };
+}
+
+function toDateOnly(date: Date): string {
```
vkrashenko: `toDateOnly` receives a local midnight and formats it in UTC, which is the previous day anywhere east of Greenwich:

```
TZ=Europe/Moscow    -> 2026-09-03
TZ=America/New_York -> 2026-09-04     (local date is 2026-09-04 in both)
```

So in UTC+ zones every preset shifts a day and "Last day" ends up covering [D-2, D-1] — today's orders are unreachable through any preset. Building the string from `getFullYear`/`getMonth`/`getDate` fixes it.

The parsing side is already correct, for what it's worth: `toStartDateFilterValue` reads `YYYY-MM-DD` as local midnight, and `toLocalDate` on the page has the mirror-case comment. It's only the formatting here.

One heads-up if the fix turns into "reuse core's presets": core has the same skew (`getDateFilterRanges` stores `x.toISOString()`, `handleOrdersDateFilterChange` cuts it with `toDateISOString`), so that would carry it over rather than remove it. A small shared `toLocalDateOnly` in `core/utilities/date` would settle both.

client-app/modules/sales-rep/composables/useSalesRepCustomerOrders.ts
```diff
@@ -0,0 +1,131 @@
...
+    Math.max(1, Math.ceil((result.value?.salesRepCustomerOrders?.totalCount ?? 0) / PAGE_SIZE)),
+  );
+
+  watch(pages, (total) => {
+    if (page.value > total) {
+      page.value = total;
+    }
+  });
+
+  watch(orgId, () => {
```
vkrashenko: This reset runs *after* Apollo's own variables watcher — `useQuery` registers `watch(variables, ..., { deep: true, immediate: true })` at the default pre flush, above this line — so an org change fires two requests, the first one with the new org and the previous offset:

```
without flush:"sync" -> org-1/after=20, org-2/after=20, org-2/after=0
with    flush:"sync" -> org-1/after=20, org-2/after=0
```

The `sortRule` watcher on the page already uses `{ flush: "sync" }` for exactly this, and `documents.vue` carries the comment explaining it. One line.

client-app/modules/sales-rep/pages/customer-orders.vue
```diff
@@ -0,0 +1,376 @@
...
+  customerOptions,
+  sortRules,
+  loading,
+  failed,
+  page,
+  pages,
+  keyword,
+  filters,
+  sortRule,
+} = useSalesRepCustomerOrders(() => props.organizationId);
```
vkrashenko, line 181: Both list routes resolve to this same component, and `RouterView` adds no `key` (neither does account-shell's `<router-view>`), so moving between `/company/customer-orders` and `/company/my-customers/:id/orders` reuses the instance: `setup()` doesn't re-run, only the prop changes. Two consequences:

- `keyword`, `filters` and the search input's local draft ride along into the other list — only `page` is reset.
- `enabled` on the customer query flips, and Apollo defers that start to `nextTick`, so `notFound` is true for one tick and the 404 view flashes. Same shape as the one fixed in #2458.

Nothing links the two routes today, so it's latent — but the `watch(orgId)` reset is already written for this case, so it may as well cover the rest of the per-list state, and `notFound` could gate on "the read for this id has started" rather than on `loading` alone. Worth handling now, because the Customer column is the obvious thing to turn into a link later.

client-app/modules/sales-rep/components/sales-rep-orders-filters.vue
```diff
@@ -0,0 +1,321 @@
+<template>
+  <VcPopover placement="bottom-end" class="sales-rep-orders-filters" :offset-options="8" :disabled="disabled" lazy>
```
vkrashenko, line 2: This is the shared trio merged into one file rather than reused: the popover wrapper matches `orders-desktop-filters.vue` prop for prop, the dialog + status list + footer predicates match `orders-filter.vue` (down to `w-[27.5rem]` and `__status-label`), and the date block matches `date-filter-select.vue` including the SCSS and its media queries.

The description explains it with calendar-aligned vs rolling presets, which is true but only covers the presets. The actual blockers, once you open the files, are that `orders-filter.vue` reads the global `facets` singleton and that `date-filter-select.vue` takes its presets from `useUserOrdersFilter` instead of a prop. Giving it an optional `ranges` prop (defaulting to today's presets) would let this module pass its rolling window and delete most of this file — timezone fix included.

If the copy stays, could the description name that seam instead? Otherwise the next reader concludes the shared components can't be reused at all.

client-app/modules/sales-rep/pages/customer-order-details.vue
```diff
@@ -0,0 +1,239 @@
...
+  allItemsAreDigital,
+} = useSalesRepCustomerOrder(() => props.orderId);
+
+const customerName = computed(() => order.value?.organizationName ?? "");
+
+const isCancelled = computed(
+  () => String(order.value?.status).toLowerCase() === String(OrderStatusCode.CANCELLED).toLowerCase(),
+);
+
+const shipmentMethodName = computed<string>(() =>
```
vkrashenko: `useOrderView` is a good split, but it stopped at the item groups: `shipmentMethodName`, `paymentMethodName`, `shipToTitle` and `isCancelled` are equally pure functions of the order, and they're retyped here from `pages/account/order-details.vue`. Moving them into the composable would shrink this page and leave the copied template as the only debt — which is at least documented.

client-app/modules/sales-rep/composables/useSalesRepCustomerOrders.ts
```diff
@@ -0,0 +1,131 @@
...
+import { useSalesRepCustomer } from "./useSalesRepCustomer";
+import { useSalesRepHubQuery } from "./useSalesRepHubQuery";
+import type { SalesRepOrdersFilterDataType, SalesRepRuleType } from "../types";
+import type { MaybeRefOrGetter } from "vue";
+
+export const PAGE_SIZE = 10;
+
+// Both sortable columns reverse freely, so the expression always carries a direction — a bare field name
+// would mean ascending to the backend.
+const SORT_RULES: SalesRepRuleType[] = Object.values(CUSTOMER_ORDERS_SORT_FIELDS).map((name) => ({
```
vkrashenko: Taste, not a bug: three layers to produce `createdDate:desc`. Synthetic rules here, then `useSalesRepColumnSort` writes a bare name for the natural direction, then `withDirection` puts the direction back on. `label` is never read by anyone. For two plain index fields a direct column → `field:dir` map — or an `alwaysEmitDirection` option on the composable — would say the same thing with fewer moving parts.

vkrashenko (commented):
Two things I couldn't settle from this repo (no backend PR link yet):

1. Does `salesRepCustomerOrders` cap `after` / skip? `salesRepActivities` stops returning rows past `MaxSkip = 500` while `totalCount` keeps describing the whole set - a pager sized from `totalCount` then offers pages that come back empty and read as "no orders".
2. Is `salesRepOrders` (the widget) guaranteed to be rep-placed orders only? It doesn't select `customerId`, so the widget can't compute `isOwn`, and its rows link straight to the buyer-facing `OrderDetails` - which breaks in exactly the case `salesRepOrderRoute` is careful about.

client-app/modules/sales-rep/components/sales-rep-orders-filters.vue
yuskithedeveloper: Agreed, and the description will name the seam. One correction to the inventory first.

The customer group is not part of the copied trio — it is the part reuse cannot reach. The shared
control filters `customername` (the buyer who placed the order) through a `VcSelect multiple` over a
plain name list; this one filters `organizationname` (the customer organization) through a checkbox
group with per-option counts from `term_facets`. The field is the load-bearing difference: a rep needs
that customer's orders regardless of which of their buyers placed them, which the shared axis cannot
express — `useSalesRepCustomerOrders` already holds `customerNames` back from `getFilterExpression`
for exactly that reason.

So reuse needs a third change beyond your two: a counted-checkbox group over a different field, which
changes the buyer-facing UI. Not a difference, to be explicit: the two-mode behaviour is not new —
`showCustomerNameFilter` already hides the group outside the organization scope.

Both your blockers are real and `ranges` is the right seam; filing the extraction as follow-up covering
all three rather than touching `date-filter-select` here, since it renders on the buyer-facing page.

The timezone half is fixed independently: `toLocalDateOnly` now lives in `core/utilities/date` as the
inverse of the parsing `toStartDateFilterValue` already does. `getDateFilterRanges` still has the skew
— left as buyer-facing scope, but the helper is there. One note for that fix: the test pairs local
midnight with local 23:59, because a midnight-only case still passes with the bug in UTC− zones.

vkrashenko (approved):

PR #2445, opened 2026-08-19
feat(VCST-5159): discover federated plugins from the platform

ivan-kalachikov:
## Description

The federated loader could only read remotes from a build-time env map, so shipping a plugin meant
rebuilding the storefront. The platform now advertises installed plugins per host app (vc-platform
#3019) and xAPI projects them onto the store (x-api 3.1016.0, x-frontend 3.1005.0), so discovery
becomes a runtime concern: installing a module is enough.

### Discovery

- A dedicated `GetStorePlugins` query (`store.plugins(appId: "vc-frontend")`), issued only by a host
  built as a federation host, gated by `APP_MODULES_FEDERATION_ENABLED`, and skipped when
  `APP_MODULES_FEDERATION_REMOTES` is set — that list wins in the loader anyway. It goes out
  alongside the other boot queries, so it adds no serial round trip; a failure degrades to "no
  plugins" and is budgeted at 2s.
- It carries `SUPPRESS_ERROR_NOTIFICATIONS_CONTEXT`: nothing on screen belongs to it, and on a
  backend that does not know the field the global handler would broadcast a generic error toast to
  every open tab for a feature the visitor cannot see.
- It is deliberately **not** part of `InitializeApplication`. `store.plugins` needs x-api 3.1016.0
  and a single unknown field fails the whole GraphQL document, which would take `settings.modules`
  with it: `apply-gates-link` would then strip every `@gate(module:)` field (branding silently
  reverts to defaults), the promotion-coupons route would 404, and `notifyOutdatedModules` — the one
  diagnostic that names the real cause — would never fire. The document is static, so that would
  happen with `APP_MODULES_FEDERATION_ENABLED` off as well.

### Entry, expose key, permission, stylesheets

- Platform descriptors point at `remoteEntry.js`; the contract gate needs the MF manifest beside it,
  so the last path segment is rewritten to `mf-manifest.json` and cache-busted with the platform's
  own `?v={hash}`. A plugin shipping no manifest stays unverifiable and is skipped. A platform entry
  must be same-origin and http(s) *after* the rewrite — `blob:` shares this origin and has an opaque
  path, which would make the rewrite a no-op and leave MF script-loading the entry as code.
- The expose key comes from the descriptor (`./Module` is the platform default) instead of being
  hardcoded to the scaffold's `./plugin`.
- Permission is evaluated host-side before any fetch: the platform serves one cached list to every
  caller and does not filter per user. No permission check available + declared permission ⇒ skipped.
  This is a UX and latency filter, not an authorization boundary — the bundle is fetched by an
  injected `<script>` with no credentials, so each plugin's data access must be authorized on its own.
- `contentFiles` of type `style` are injected once each, after the plugin has loaded and before its
  `init()` runs: a timeout does not cancel `init()`, so withholding the sheets would leave a live
  page unstyled. A content file of any other kind — or of no declared kind — is logged, not dropped
  in silence. Not wrapped in a cascade layer; see VCST-5760.
- The env map wins when set, so a plugin author's local remote is never replaced by what the backend
  serves.

### Loader hardening

- Remote names are validated (`/^[A-Za-z0-9][A-Za-z0-9._-]*$/`, both discovery paths). MF resolves a
  `loadRemote` id by **prefix**, so with remotes `a` and `a/plugin` the request meant for the second
  matches the first and is served out of *its* bundle — one plugin's code never runs while both are
  reported loaded. Exact-name dedup cannot see that, here or in MF.
- Duplicate remote names: first wins, the rest are skipped loudly. A name is claimed only after the
  descriptor passes validation, so a plugin about to be skipped cannot block a working one.
- A non-`script` `entry.type`, a non-string field anywhere in the projection, and a non-array plugin
  list each cost one plugin instead of throwing out of the batch. A descriptor with neither `id` nor
  remote name is reported by its position, so nothing is dropped without a trace.
- `init()` is called **on** the module, so `{ services, init() { this.services } }` — the natural
  shape of the platform's default `./Module` expose — keeps its receiver. A module exposing no
  `init()` counts as loaded with a warning: its module scope did run.
- Per-phase budgets: manifest 2s, load 3s, init 3s, discovery 2s, under a 12s boot backstop. Three
  of those are **reductions** from `dev` (manifest 3s, load 5s, backstop 20s): the invariant only
  asks that the backstop clear discovery + manifest + 2×load, so the tighter numbers keep the same
  2s of headroom for the unbudgeted loader-chunk fetch while cutting worst-case blank screen from
  20s to 12s. The cost is that a plugin whose `init()` legitimately runs longer than 3s now reports
  `failed`; 5s would need a 16s backstop. The loader chunk's own fetch stays unbudgeted — bounding
  it is what the backstop is for.
- Route guard: for the whole load-and-init phase `addRoute` and `removeRoute` refuse a claim on a
  name the host already owns. Both `addRoute` overloads are covered, dispatched the way vue-router
  dispatches them (by predicate, not argument count), plus every named entry in `children`, since
  vue-router treats those as root adds. `removeRoute` is wrapped in the same window because
  remove-then-add would otherwise leave the name free by the time the add is checked. One guard
  covers the whole phase, and the loader starts only after every host plugin has installed, so a
  host call is never refused in a plugin's place.
- No `{ force: true }` on `registerRemotes`: re-registering a known name is already a no-op in MF
  2.8.0, while `force` calls `removeRemote` first, so a second boot would re-run the plugin's module
  scope and call `init()` twice.
- The manifest **response** is re-checked for origin, per discovery source — same-origin for a
  platform entry, https (or loopback http) for an env one — because `fetch` follows redirects.

### Local dev and scaffolding

- `^/modules/.*/plugins/vc-frontend/` is proxied to `APP_BACKEND_URL` in dev and preview, so the
  discovery path is exercisable without a deployed host. Scoped to the plugin folder, not all of
  `/modules`.
- `create:plugin` now writes `public/plugin.json` (Vite copies it into `dist/`). Without a descriptor
  the platform assumes its own defaults — remote name = the .NET module id, expose `./Module` — and
  the host would `loadRemote` a key the generated plugin does not export.

### Deferred, with tickets

- **VCST-5761** — let a plugin declare its routes in `plugin.json`, so the host can register
  placeholders and stop gating `app.mount()` on this loader.
- **VCST-5762** — extension-registry precedence; needs a facade signature change and a contract rebuild.
- **VCST-5760** — plugin CSS containment through one host-declared cascade layer order. Wrapping only
  the plugin's sheet was tried and reverted: unlayered beats layered, so the plugin would lose to all
  host CSS. Measured regression surface and the rejected alternatives:
  `client-app/modules/federated/specs/2026-08-21-plugin-css-cascade-layers.md`.
- Artifact integrity, the per-plugin kill switch, route authorization and the ingress CSP stay in
  `TODO.md`. The `?v={hash}` cache-buster reaches the manifest URL, not the `remoteEntry.js` it
  points at; the clean fix is a `Cache-Control` on the platform's static file middleware.

### Docs

- `README.md` — the security model no longer advertises a build-time remote list as an enforced
  property. It now states what is enforced (same-origin platform entries, https-or-loopback env
  entries, both re-checked after redirects) and what is not (no origin allowlist for env remotes; a
  same-origin manifest may still declare chunk URLs elsewhere), plus the budgets and the route
  guard's limits.
- `HOWTO.md` — leads with the backend-module path and explains the `plugin.json` the scaffolder
  emits (the scaffold exposes `./plugin`, the platform defaults to `./Module`).
- `TODO.md` — artifact integrity closed as deliberately not done, with the reasons
  (hostile code arrives on the first fetch anyway; MF injects `remoteEntry.js` as a `<script>`, so
  the executed bytes are never ours to hash, and `entry.hash` is a cache-buster, not an SRI digest).
- `specs/2026-07-06-discovery-hosting-decision.md` — the discovery half is marked superseded.

### Generated files

`dev` is merged, and both generated GraphQL type files are regenerated against **vcptcore-dev**:
`CustomerCartStatisticsPeriod.lastCartDate`, which `dev` removed in #2437, is gone, and
`modules/sales-rep/api/graphql/types.ts` is byte-identical to `dev`'s. vcst-qa cannot generate the
sales-rep module — its schema lacks the document queries added in #2439. Three environment deltas
remain: `CartType.loyaltyPoints` is absent on that environment (no document selects it, and the other
hits are untouched module files plus `Product.loyaltyPoints` in the contract); one field description
carries older wording; and vcptcore-dev **adds** `SalesRepDocument{,Category,Connection,Edge}`,
`Query.salesRepDocument{,Categories,s}` and their `*Args` — 86 of the 90 added lines, absent from
`dev`, and reversed again by regenerating on vcst-qa.

The core contract was already stale on `dev` — the `sharedList` extension point and its
wishlist-sharing types were added without a rebuild, and `validate:core-types` runs in no workflow —
so it is rebuilt here. The build's own automation bumps `CORE_VERSION` 1.2.0 → 1.3.0 for the changed
contract; plugins built against `^1.2.0` keep running.

### Verified

- 2447 tests in 159 files; `vue-tsc --build --force` clean; eslint on the touched files: 0 errors,
  23 warnings (10 `sonarjs/no-redundant-optional` on `IPlatformPlugin`, which mirrors codegen's
  `Maybe<T>` on purpose, and a `sonarjs/null-dereference` false positive). `validate:core-types` is
  red on a local checkout, but the only differing line is the property order of `VcButton`'s
  inferred defaults, byte-identical at the merge base — a vue-tsc artefact, not this PR.
  `boot-order.test.ts` executes the real `app-runner` against mocked
  collaborators and pins the order the calls actually happen in: theme context and user before the
  loader, the router installed only after it, the plugin-list query issued before the page context is
  awaited, nothing issued when the flag is off or the env override is set, and the loader started
  after the host's route-mutating plugin.
- Live against vcst-qa: `store.plugins(appId: "vc-frontend")` answers through the new query and
  returns `[]` — no module there ships a storefront plugin yet. The dev proxy was verified end to
  end: a request to `/modules/$(VirtoCommerce.SalesRep)/plugins/vc-frontend/mf-manifest.json` reaches
  the platform (HTTP 404, empty body), while `/modules/$(X)/Content/logo.png` and any other path
  still get Vite's SPA fallback — which is the old failure mode, since that HTML is what the contract
  gate used to try to parse as JSON.
- Not exercised: a real platform-served plugin load. That needs a module on the target environment
  shipping `plugins/vc-frontend/`; no environment has one.

### Known repo-state issues found alongside, not fixed here

`validate:core-types` runs in no PR workflow while a comment claims it does, the breaking-change
detector diffs export names only, `federation.mjs` ships in the versioned tarball but is outside the
change detection, and the production logger noops every level including `error`.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5159
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.58.0-pr-2445-d7c2-d7c2c9c6.zip


client-app/modules/federated/index.ts
```diff
@@ -184,6 +456,125 @@ function resolveRemotes(): IResolvedRemotes {
...
+  // after that cannot be attributed to one plugin.
+  const who = () => (initingPlugin ? `"${initingPlugin}"` : "a plugin");
+
+  // Captured unbound and applied with the router as receiver, so the host gets its own methods
+  // back - restoring a bound copy would leave them permanently replaced.
+  const originalAdd = router.addRoute as (...args: unknown[]) => () => void;
+  const originalRemove = router.removeRoute as (...args: unknown[]) => void;
+
+  const guardedAdd = (...args: unknown[]) => {
+    const record = args.length >= 2 ? args[1] : args[0];
```
goldenmaya: The guard dispatches the overloads by **arity**, but vue-router dispatches by **predicate**, so `addRoute(record, undefined)` walks straight past the check and evicts the host route.

`node_modules/vue-router/dist/vue-router.mjs:1173-1182`:

```js
function addRoute(parentOrRoute, route) {
  let parent, record
  if (isRouteName(parentOrRoute)) { parent = matcher.getRecordMatcher(parentOrRoute); record = route }
  else record = parentOrRoute
  return matcher.addRoute(record, parent)
}
```

So for a two-argument call whose first argument is a *record*, vue-router takes `record = args[0]` while this line takes `args[1]` — `undefined` — and `claimedRouteNames(undefined)` returns `[]`, so nothing is claimed and the call is forwarded.

Verified against a real `vue-router` instance (not a stub), three cases:

* **precondition** — `router.addRoute({ name: "Checkout", path: "/hijacked" }, undefined)` on a plain router does replace the host's `Checkout`, `/checkout` → `/hijacked`.
* **bypass** — the same call from a plugin's `init()` with the guard installed: `result.loaded === ["news"]`, `Checkout` ends up at `/hijacked`, and `Logger.error` is **never called**. Silent takeover, which is the one thing this guard exists to prevent.
* **control** — the one-argument form on the same router is refused correctly, so the guard itself works; only this shape escapes.

Any wrapper that always forwards two arguments hits it too, not just a deliberately crafted call.

Suggested fix — use vue-router's own predicate rather than the argument count:

```ts
const parentGiven = typeof args[0] === "string" || typeof args[0] === "symbol";
const record = parentGiven ? args[1] : args[0];
```

Worth a test with the `(record, undefined)` shape next to the existing two-argument one, since the current two-argument test passes a *name* first and therefore cannot catch this.

client-app/modules/federated/index.ts
```diff
@@ -171,11 +230,224 @@ function resolveRemotes(): IResolvedRemotes {
...
+    }
+    // The platform documents these kinds as lower-case by contract, but a dropped stylesheet is
+    // invisible otherwise: the plugin loads and renders unstyled with nothing to point at.
+    const fileType = asString(file?.type);
+    if (fileType?.toLowerCase() !== PLATFORM_STYLE_FILE_TYPE) {
+      // Only when a kind was actually declared: `kind "undefined"` read like a platform bug rather
+      // than an asset this loader simply has no use for.
+      if (fileType) {
+        Logger.info(`[MF] Plugin "${String(plugin?.id)}": ignoring content file of kind "${fileType}"`);
+      }
```
goldenmaya: A `contentFiles` entry with **no declared `type`** is dropped here with zero diagnostics — the silent path the PR description says round 2 closed.

`type` is nullable in the generated type (`contentFiles: Array<{ type?: string, path?: string, hash?: string }>`), so `fileType` is `undefined`, `undefined !== "style"` takes the skip branch, and the `if (fileType)` guard then suppresses the only log.

This is the opposite of what `isLoadableEntryType` does two functions down, for the same optionality:

> An absent type is accepted, because the platform declares the field optional.

Verified, with a control:

* `contentFiles: [{ type: null, path: "/modules/X/plugins/vc-frontend/style.css" }]` → plugin reports `loaded`, `document.head.querySelectorAll("link[data-mf-plugin-style]")` is **0**, and no `Logger.info` call contains `"ignoring content file"`.
* `type: "style"` → 1 `<link>`, as expected.

So the platform returning `{ type: null }` gives you a plugin that loads, renders unstyled, and leaves nothing to point at — exactly the failure mode the comment above says is invisible otherwise.

Either treat an absent `type` as a stylesheet (consistent with the entry-type rule), or keep skipping it but log it. Whichever way, I'd move the log out of the `if (fileType)` so the "no kind declared" case is never the silent one.

client-app/modules/federated/index.ts
```diff
@@ -243,44 +649,66 @@ export async function initFederatedModules(options?: IFederatedLoaderOptions): P
...
+    // the "never rejects" contract.
     compatible.forEach((remote) => result.failed.push(remote.name));
     Logger.error("[MF] registerRemotes failed", error);
-    reportOutcome(result);
+    reportOutcome(result, versions);
     return result;
   }
 
+  // One guard for the whole phase — see installRouteGuard on why it must not be per plugin.
+  const releaseRouteGuard = installRouteGuard();
```
goldenmaya, line 710: The guard is caller-agnostic and it wraps the **shared** host router, while host boot code can still be mutating routes. Net damage today is nil, but the reason is luck rather than design — so flagging it as fragility, not as a break.

`app-runner` does not `await federatedModulesReady` until line 316. Between the `startFederatedModules(...)` call at 280 and that await there are two `await import(...)` points, and the first is followed by `app.use(builderPreviewPlugin, { router })` — whose `install()` calls `modifyRoutes()` synchronously (`builder-preview.plugin.ts:337`), which does `removeRoute("Matcher")` → `addRoute(...)` → `addRoute(matcher)` (`:313-321`).

What I measured, running that exact sequence from *host* code inside an open window on a real router:

```
DAMAGE: /designer-preview -> StaticPage | /some/page -> Matcher
        routes=Checkout,StaticPage,Matcher | refusals=2
```

Both host calls are refused, but because the `removeRoute` is refused too, `Matcher` never leaves, and `/designer-preview` still wins on path score. So builder-preview keeps working — the visible outcome is two misleading log lines blaming "a plugin" for something the host did.

Window size, measured with fake timers:

* opens right after `registerRemotes`, closes when `Promise.allSettled` settles;
* **~2900 ms** for a remote whose `loadRemote` hangs;
* load and init budgets **stack** — 2.5 s load + hanging init keeps it open past 3050 ms and closes it by 6100 ms, so worst case ≈ 6 s;
* for a healthy plugin the whole loader took **1 ms** and a 1 ms poller never observed the guard at all.

One correction to the PR description while I'm here: the backstop cannot hold this window open. I tested the claim directly and it is false — every leg inside `allSettled` is budgeted (2 + 3 + 3 = 8 s), so the guard is always released well before `BOOT_BACKSTOP_MS` (12 s) fires. A test asserting the router was still wrapped after the backstop fails.

Cheapest fix is ordering, not logic: move the two preview-plugin blocks above the `startFederatedModules(...)` call, so no host `app.use(..., { router })` can interleave with the phase. The discovery query still overlaps boot, since it starts at line 161 regardless.

client-app/modules/federated/index.ts
```diff
@@ -171,11 +230,224 @@ function resolveRemotes(): IResolvedRemotes {
...
+  for (const plugin of Array.isArray(plugins) ? plugins : []) {
+    const name = asString(plugin?.remote?.name) || asString(plugin?.id) || "";
+    const path = asString(plugin?.entry?.path);
+    if (!SAFE_REMOTE_NAME.test(name)) {
+      const label = asString(plugin?.id) || name;
+      Logger.error(`[MF] Skipping plugin "${label}": remote name "${name}" must match ${String(SAFE_REMOTE_NAME)}`);
+      // Reported under whatever identifies it, so the outcome summary never drops a plugin silently.
+      if (label) {
+        resolved.invalidNames.push(label);
+      }
```
goldenmaya: When a descriptor has **neither** `remote.name` **nor** `id`, `name` is `""`, `label` is `""`, and `if (label)` skips the push — so the plugin never appears in `skipped`. That contradicts the contract stated a few lines up ("never silently dropped") and repeated in the README.

Verified:

```ts
await initFederatedModules({
  plugins: [{ entry: { type: "script", path: "/modules/X/plugins/vc-frontend/remoteEntry.js" } }],
});
// => { loaded: [], failed: [], skipped: [] }
```

The `Logger.error` above is the only trace, and per the README's own note `Logger` is a no-op at **every** level in production — including `error` — so in prod this plugin vanishes without any signal at all.

A positional placeholder would close it while staying honest about not knowing the identity, e.g. `plugin #${index}` (or the raw `String(plugin?.id)` you already use in the duplicate-name branch below).

client-app/app-runner.ts
```diff
@@ -154,6 +155,14 @@ export default async () => {
     Logger.warn("Failed to verify backend module versions", e);
   }
 
+  // Issued here so its round trip overlaps the boot queries instead of following them. The flag is
+  // still what decides whether it happens at all. The no-op handler only marks a rejection as
+  // observed until startFederatedModules attaches the one that logs and degrades to "no plugins".
+  const storePluginsPromise = isMfFlagEnabled(import.meta.env.APP_MODULES_FEDERATION_ENABLED)
+    ? getStorePlugins(domain)
+    : undefined;
+  void storePluginsPromise?.catch(() => {});
```
goldenmaya: The gate is the flag only, so the query is also issued when `APP_MODULES_FEDERATION_REMOTES` is set — and `resolveRemotes` then returns the env map unconditionally and throws the result away.

That is exactly the plugin-author loop `HOWTO.md` step 4 prescribes: a wasted round trip on every boot, and against a backend older than x-api 3.1016.0 also a per-boot `[MF] Could not read the platform's plugin list` for a developer who never asked for platform discovery. (`withDiscoveryBudget` runs before `resolveRemotes`, so the error path fires regardless of the override winning.)

```ts
const storePluginsPromise =
  isMfFlagEnabled(import.meta.env.APP_MODULES_FEDERATION_ENABLED) &&
  !import.meta.env.APP_MODULES_FEDERATION_REMOTES
    ? getStorePlugins(domain)
    : undefined;
```

Both vars are build-time inlined, so this stays a compile-time decision and costs nothing.

Unrelated, and a compliment rather than a note: I checked whether `void storePluginsPromise?.catch(() => {})` is really load-bearing, and it is — `Promise.race` inside `withDiscoveryBudget` does subscribe to the inner promise, but only once the loader gets there, so this is what covers the gap in between. The comment is accurate.

client-app/core/api/graphql/types.ts
```diff
@@ -991,17 +991,19 @@ export type CustomerCartStatisticsComparison = {
...
 export type CustomerCartStatisticsPeriod = {
-  /** 'total' divided by 'count'. */
+  /** Average cart value in the range (amount, formatted amount and currency). */
   average: MoneyType;
-  /** Number of distinct carts contributing to 'total'; a cart whose lines are all parked reports quantities but does not count. */
+  /** Number of carts in the range. */
   count: Scalars['Int']['output'];
-  /** Summed quantity of the line items selected for checkout. */
+  /** Date of the most recent cart in the range. */
+  lastCartDate?: Maybe<Scalars['DateTime']['output']>;
```
goldenmaya: **Merge blocker.** This re-adds a schema member `dev` deliberately removed, and the merge will not surface it in any way.

Traced end to end:

| Check | Result |
|---|---|
| `git show origin/dev:…/types.ts \| grep -c lastCartDate` | `0` |
| same on this branch | `1` |
| who removed it in `dev` | `ab9f0a5ea` (#2437): `-  lastCartDate?: Maybe<Scalars['DateTime']['output']>;` |
| descriptions too | same commit: `-/** Average cart value in the range … */` → `+/** 'total' divided by 'count'. */` — this branch restores the pre-#2437 wording for ~14 fields |
| `git merge-base origin/dev HEAD` | **equals `origin/dev`** |
| `lastCartDate` used anywhere in `client-app` | nothing outside `types.ts` |

Because `dev` is an ancestor of this branch, the merge is a fast-forward: **no conflict, no prompt, this file simply wins.** `vue-tsc` stays green because nothing references the field. So the regression lands entirely unobserved, and the first symptom is someone writing a selection on `lastCartDate` months later, passing type-check, and getting `Cannot query field` at runtime against the deployed schema.

The PR description already parks this — I'm marking it inline so it can't be merged by accident. Rebase + `yarn generate:graphql-types` against the environment `dev` was generated from; not a hand edit.

client-app/modules/sales-rep/api/graphql/types.ts
```diff
@@ -84,17 +84,19 @@ export type CustomerCartStatisticsComparison = {
...
 export type CustomerCartStatisticsPeriod = {
-  /** 'total' divided by 'count'. */
+  /** Average cart value in the range (amount, formatted amount and currency). */
   average: MoneyType;
-  /** Number of distinct carts contributing to 'total'; a cart whose lines are all parked reports quantities but does not count. */
+  /** Number of carts in the range. */
   count: Scalars['Int']['output'];
-  /** Summed quantity of the line items selected for checkout. */
+  /** Date of the most recent cart in the range. */
+  lastCartDate?: Maybe<Scalars['DateTime']['output']>;
```
goldenmaya: Same revert as in `core/api/graphql/types.ts` — part of the merge blocker, listed separately because this file needs regenerating too.

md5 of the whole file:

```
origin/dev            c823da8ba673281e5347d19fa733ddcf
ab9f0a5ea (#2437)     c823da8ba673281e5347d19fa733ddcf   <- unchanged since
this branch           e28b9503b90500bdaa3264aa288f5a63
```

So `dev` and #2437 agree, and this branch diverges from both. One small correction to the PR description, which says this file "matches neither parent": since `dev` is an ancestor of the branch it is the only parent, and the accurate statement is simply that this branch reverts `dev`'s version.

Same fix — `yarn generate:graphql-types` after the rebase, so both files come from one generator run.

client-app/modules/federated/boot-order.test.ts
```diff
@@ -0,0 +1,235 @@
...
+        order.push("getStorePlugins");
+        return Promise.resolve([]);
+      });
+
+      await runBoot();
+
+      // Ahead of the loader by a lot: awaiting it there instead would add a serial round trip to
+      // boot, which is the whole reason the call site sits where it does.
+      expect(order.indexOf("getStorePlugins")).toBeGreaterThan(-1);
+      expect(order.indexOf("getStorePlugins")).toBeLessThan(order.indexOf("startFederatedModules"));
```
goldenmaya: This case does not test what its name claims. The name says "before it awaits the page context", but the `getPageContext` mock never writes a marker into `order`, so the only comparison is against `startFederatedModules` — which is ~120 lines later in `app-runner` and passes no matter where the query is issued.

Proved by mutation. I moved the `getStorePlugins` call site from line 161 down to immediately before `startFederatedModules(...)` — i.e. the exact loss of overlap this test is named for — and re-ran the file:

```
MUTATION APPLIED: call site moved to just before startFederatedModules
Test Files  1 passed (1)      Tests  4 passed (4)
```

Given how carefully the rest of the suite is built to fail on moved call sites (the `LOADER_REACHED` rejection trick is genuinely nice), this one is worth closing the same way — have the `getPageContext` mock push a marker and assert `getStorePlugins` precedes it:

```ts
getPageContext: vi.fn(async () => {
  order.push("getPageContext");
  return { store: { storeId: "s" }, user: { id: "u1" }, whiteLabelingSettings: {}, slugInfo: undefined };
}),
```

then `expect(order.indexOf("getStorePlugins")).toBeLessThan(order.indexOf("getPageContext"))`. That mutation turns it red.

client-app/modules/federated/boot-order.test.ts
```diff
@@ -0,0 +1,235 @@
...
+}));
+vi.mock("@/modules/back-in-stock", () => ({ init: vi.fn() }));
+vi.mock("@/modules/customer-reviews", () => ({ init: vi.fn() }));
+vi.mock("@/modules/google-analytics", () => ({ init: vi.fn() }));
+vi.mock("@/modules/loyalty", () => ({ init: vi.fn() }));
+vi.mock("@/modules/news", () => ({ init: vi.fn() }));
+vi.mock("@/modules/purchase-requests", () => ({ initialize: vi.fn() }));
+vi.mock("@/modules/push-messages", () => ({ init: vi.fn() }));
+vi.mock("@/modules/quotes", () => ({ init: vi.fn() }));
+vi.mock("@/modules/sales-rep", () => ({ init: vi.fn() }));
```
goldenmaya, line 122: Not a defect — a maintenance cost worth deciding on consciously, given this repo is forked and clients pull updates.

This block mocks the whole `app-runner` import graph, so every module a fork adds to `app-runner` becomes a required entry here, and the failure mode when one is missing points nowhere near the cause. I removed a single unrelated mock (`@/modules/news`) to see it:

```
× puts the theme context and the user in place before the loader…
  → expected [Function] to throw error including 'loader reached' but got 'Theme context is missing.'
Tests  4 failed (4)
```

All four cases fail with `Theme context is missing.` — a client debugging that will not guess "you added a module import to app-runner". The comment at the top of the block already anticipates part of this (the "always a factory, never a bare `vi.mock`" note is a good catch), so it's clearly been thought about.

The invariants themselves are real and I'd keep the test. Options, in the order I'd consider them: assert the loader-boundary order inside `bootstrap`/`index` where the graph is small; or keep this file but fail loudly on an unmocked import (a catch-all `vi.mock` factory that throws a message naming the path). At minimum, a line in the file header saying "adding a module to app-runner means adding a mock here" would save the next person the hour.

client-app/modules/federated/index.ts
```diff
@@ -317,7 +749,8 @@ export async function initFederatedModules(options?: IFederatedLoaderOptions): P
       }
     }),
   );
+  releaseRouteGuard();
```
goldenmaya: Nit, and I checked before raising it: I walked every statement between `installRouteGuard()` and here looking for a reachable throw and found none — the whole `compatible.map` body is inside its own `try/catch`, `Promise.allSettled` cannot reject by contract, and even a throwing `Logger.error` in the catch would only reject an already-settled member.

So this is unreachable today, not a bug. But "the host always gets its router back" is the kind of invariant that is cheaper to hold by construction than to re-derive on the next edit:

```ts
try {
  await Promise.allSettled(/* … */);
} finally {
  releaseRouteGuard();
}
```

Same behaviour, one less thing a future change can quietly break.

client-app/modules/federated/index.ts
```diff
@@ -1,33 +1,61 @@
...
+
+/** One plugin as xAPI projects it (`store.plugins`); structural so the loader stays off the core GraphQL layer. */
+export interface IPlatformPlugin {
+  id: string;
+  version?: string | null;
+  permission?: string | null;
+  entry?: { type?: string | null; path?: string | null; hash?: string | null } | null;
+  contentFiles?: readonly ({ type?: string | null; path?: string | null; hash?: string | null } | null)[] | null;
+  remote?: { name?: string | null; exposed?: string | null } | null;
 }
```
goldenmaya, line 59: Minor: this interface accounts for 10 of the 12 new lint warnings this file picks up. Measured on the touched files — `origin/dev` version of `index.ts` reports **1 warning**, this version reports **13**, all `sonarjs/no-redundant-optional` on lines 54-58 plus `function-return-type` (188) and `null-dereference` (272). Zero errors either way, so nothing is failing.

The last two are false positives (`url.pathname` is always a string) and not worth contorting the code for. The 10 are avoidable by expressing nullability the way the generated types do, since that is what this structural type is mirroring:

```ts
version?: Maybe<string>;
permission?: Maybe<string>;
```

Up to you whether it's worth it — I'd rather the file stay readable than chase warnings — but flagging it since the module otherwise keeps a clean lint surface.

goldenmaya (commented):
Reviewed on the branch, with everything below verified by execution rather than by reading: throwaway specs driving a **real `vue-router`** instance, mutation tests against the PR's own test files, fake-timer measurements of the budgets, and reading MF 2.8.0 in `node_modules` for each runtime claim. Baseline: `yarn test:unit` on `modules/federated` **151/151**, `yarn validate:types` **exit 0**, `eslint` on the touched files **0 errors / 25 warnings**. All scratch files removed; the branch is untouched.

The engineering standard here is high — fail-closed on every path, per-phase budgets, total isolation, and documentation that states its own limits instead of overselling them. Findings are inline. Summary:

**Blocker**
* Generated types revert `dev`'s codegen from #2437. `merge-base(dev, HEAD) == dev`, so the merge is a fast-forward: no conflict, no prompt, `vue-tsc` green, regression lands unobserved. Already parked in the description; marked inline so it can't slip through.

**Should fix (each with a reproducing case)**
* `index.ts:525` — the route guard is bypassed by `addRoute(record, undefined)`; a real router shows the host's `Checkout` taken over to `/hijacked` with `Logger.error` never called. vue-router dispatches its overloads by predicate, not arity.
* `index.ts:294` — a `contentFiles` entry with `type: null` is dropped with no log and no stylesheet, the opposite of what `isLoadableEntryType` does for the same optionality.

**Worth addressing**
* `index.ts:711` — the guard wraps the shared router while host boot code can still run. Measured damage is nil (builder-preview survives), cost is two misleading log lines and fragility-by-luck. Cheapest fix is ordering in `app-runner`.
* `app-runner.ts:161` — the discovery query is issued even when the env override wins, so the documented plugin-author loop pays a wasted round trip plus a per-boot error against older x-api.
* `index.ts:391` — a descriptor with neither `id` nor `remote.name` disappears from the outcome entirely.
* `boot-order.test.ts:220` — the overlap case passes after the exact mutation it is named to catch.
* Plus two nits (`releaseRouteGuard` outside `finally`, lint warnings on `IPlatformPlugin`) and one maintenance note on the mock graph.

**Two claims in the description I checked and found not to hold**
* *"if the 12s backstop fires, the router stays wrapped"* — it cannot. Every leg inside `allSettled` is budgeted (2 + 3 + 3 = 8 s), so the guard is always released first; a test asserting otherwise fails. Measured window: ~2900 ms for a hanging `loadRemote`, ≈6 s worst case when load and init budgets stack, and **1 ms with the guard never observable** for a healthy plugin.
* The blank-screen worst case is the backstop, **12 s**, not the 8 s of loader budgets — `startFederatedModules` awaits `Promise.race([work, backstop])`.

**Claims I verified as accurate** (listing these because they were load-bearing and I did not want to take them on trust)
* `isPureRemoteEntry` really is `!remote.entry.includes(".json")` — `runtime-core/dist/utils/tool.js:18`.
* MF really resolves a `loadRemote` id by **prefix**, first match wins — `runtime-core/dist/utils/manifest.js:8`. The `a` / `a/plugin` hazard is real and `SAFE_REMOTE_NAME` closes it on both discovery paths.
* Dropping `{ force: true }` is correct: `remote/index.js:354-374` dedupes by exact name, and the non-`force` branch is genuinely **empty** — the `warn` sits inside the `force` branch, which does call `removeRemote` first. `registerRemotes` also registers one at a time, exactly as the adjacent comment now says.
* Named `children` are root adds in vue-router (`vue-router.mjs:624-625`), so the recursive walk is right.
* The round-3 cross-origin fix works: the HOWTO step-4 override (`http://localhost:3001` against a host on `:3000`) loads. Protocol-relative platform entries, off-origin manifest redirects on the platform path, and the `$(ModuleId)` token surviving the path rewrite plus `?v=` all behave as documented.
* `AbortSignal.timeout` is not a new compatibility risk — `config/links/timeout.ts:7` already uses it, and `AbortSignal.any`, under the current browserslist.

**One thing to decide rather than fix here:** `app.mount()` is gated on this loader, so up to 12 s of that is blank-screen time for the whole storefront on any store that turns MF on. VCST-5761 is the right answer; the call is whether discovery ships before it or with it. Same for the untested leg the description is candid about — no platform-served plugin has actually been loaded end to end yet.

client-app/modules/federated/index.ts
ivan-kalachikov: Fixed — the guard now dispatches by `isRouteName` like vue-router does, with a `(record, undefined)` test.

client-app/modules/federated/index.ts
ivan-kalachikov: Fixed — an undeclared kind is logged now too.

client-app/modules/federated/index.ts
ivan-kalachikov: Fixed by ordering — the loader starts after the preview plugins install. Your backstop correction is right, and it goes further: the loader-chunk fetch is unbudgeted, so the window can outlive the cap. Documented.

client-app/modules/federated/index.ts
ivan-kalachikov: Fixed — reported as `plugin #N`.

client-app/app-runner.ts
ivan-kalachikov: Fixed — the query is skipped when the override is set.

client-app/modules/federated/boot-order.test.ts
ivan-kalachikov: Fixed — it asserts against `getPageContext` now and fails on that mutation.

client-app/modules/federated/boot-order.test.ts
ivan-kalachikov: Added a maintenance note in the file header.

client-app/modules/federated/index.ts
ivan-kalachikov: Taken — released in a `finally`.

client-app/core/api/graphql/types.ts
ivan-kalachikov: Fixed — regenerated against vcptcore-dev (vcst-qa fails on the sales-rep module: no document queries yet). `lastCartDate` is gone. Two env deltas remain: `CartType.loyaltyPoints` is absent there (no document selects it) and one description is older wording.

client-app/modules/sales-rep/api/graphql/types.ts
ivan-kalachikov: Fixed — this file is now byte-identical to dev.

client-app/modules/federated/index.ts
ivan-kalachikov: Measurement reproduces exactly: 13 warnings, 10 `no-redundant-optional` on 54-58, plus 188 and 272 (the 13th is the pre-existing `todo-tag` at 618). Zero errors.

Leaving it, though — and the suggested form would not do what it looks like. Codegen here runs with `maybeValue: "T"` (`scripts/graphql-codegen/generator.ts:95`), so `Maybe<T> = T`, not `T | null`. `version?: Maybe<string>` would therefore mean `version?: string` and drop the `| null` this file actually relies on: every read of these fields goes through `asString(plugin?.version)` / `asString(plugin?.entry?.path)` precisely because xAPI does send null.

A local `type Nullable<T> = T | null` keeps the type honest and silences 9 of the 10, but the nested `exposed?` at 60 survives it, it needs a prettier reflow, and it adds an alias whose only job is dodging a syntactic rule. So taking your own preference: the file stays readable.

ivan-kalachikov: Counts confirmed (1 vs 13, 0 errors). Left as is: this repo generates with `maybeValue: "T"`, so `Maybe<string>` is just `string` — the suggestion silences the rule by dropping `| null`, which the backend can actually send. Keeping the honest type over 10 warnings the CI never runs.

goldenmaya:
**Review of #2445 — one consolidated comment**

Reviewed at `60252959a` against merge-base `17ad3e6f6`. Gates I ran: `eslint` on all 14 changed code files (0 errors / 26 warnings), `validate:types` clean, `build-only` clean, `vitest client-app/modules/federated` 8 files / 157 tests green, SonarCloud 0 open issues. `validate:core-types` is red on my checkout, but the only differing line is the property order of `VcButton`'s inferred defaults, and that block is byte-identical at the merge base — a toolchain artefact, not this PR.

I also closed part of the PR's own "not exercised" gap by proxying `APP_BACKEND_URL` through a local mock that answers `GetStorePlugins` and serves an `mf-manifest.json`. That confirmed end to end: the query is issued at boot; `remoteEntry.js` → `mf-manifest.json?v={hash}` rewriting works through the new `vite.config.ts:47` proxy rule; nine adversarial descriptors each hit the guard they should, all but the valid ones before any network; and on a backend that rejects `store.plugins` the app boots with both log lines and **no** error toast. The security guards hold up under probing — `blob:` is caught by the `^https?:$` check after the rewrite (`index.ts:366`), `//evil.example/…` by `isSameOrigin`. Findings below, most significant first.

**1. A non-string `permission` silently switches the permission gate off — `index.ts:660-663`, `index.ts:256-258`; test at `platform-discovery.test.ts:390-406`**
`asString` maps a non-string to `undefined`, and `!permission` then reads as "no permission required". Verified live with a control in the same run: a descriptor with `permission: 7` passed the filter as an **anonymous visitor** and its manifest was fetched (`?v=nonstr`), while a sibling with a string permission the visitor lacks was correctly skipped before any network (`[MF] Skipping "stringPerm": requires permission …`). The `it.each` at `:390-406` looks like the test for this, but all four rows assert only `expect(result.loaded).toContain("healthy")` — nothing about the bad plugin — and the title promises a skip that never happens; the real behaviour is graceful degradation. Mutating `asString` to a passthrough reddens only 3 of the 4 rows, so the `remote.name` row is vacuous even against a real mutation. Suggest renaming to "…costs one plugin, not the batch" and asserting `result.loaded`/`result.skipped` in full, so the `permission` row's actual effect is on the record. `Logger.info` is a no-op in production, so today this leaves no signal at all.

**2. The "stylesheets before `init()`" ordering has no test — `index.ts:733-737`**
Moving `injectStyles(remote.styles)` to after `await runInit(...)` in both branches leaves the whole suite green (0 failing tests across `index.test.ts`, `platform-discovery.test.ts`, `bootstrap.test.ts`). The two style tests don't discriminate: `platform-discovery.test.ts:150` uses an `init` that resolves, and `:242` fails at `loadRemote` so no `<link>` is expected either way. This is the property the four-line comment at `:733-736` defends and that `HOWTO.md:182-185` promises plugin authors ("a plugin whose `init()` fails or overruns is still styled"). One test closes it: a `style` content file plus an `init` that throws or overruns `loadTimeoutMs`, asserting the `<link>` is present.

**3. `IPlatformPlugin` is a hand-written mirror in which only `id` is load-bearing — `index.ts:52-59`**
`tsc` probe: renaming `entry` → `entryPoint` in `getStorePlugins.graphql` compiles clean; so does dropping `remote`, or renaming `entry.path`. Only removing `id` errors. Every other field is `?: X | null`, so absence type-checks. Runtime cost of each drift is silent: `permission` dropped ⇒ the gate is off for every plugin; `entry.path` renamed ⇒ every plugin skipped as "declared no entry path"; `remote` dropped ⇒ the expose key falls back to `./Module`, so every scaffold-built plugin (`./plugin`) fails to load — all with a green build. The layer-independence rationale at `:51` is fair, but the comment reads as though the type is checked; either derive it from `NonNullable<GetStorePluginsQuery["store"]>["plugins"]` (keeping the `asString` runtime guards), or add an exactness assertion in the test file, or say in the comment that the guarantee covers `id` only.

**4. An undeclared `contentFiles[].type` drops the stylesheet, while the same omission on `entry.type` is accepted — `index.ts:295` vs `index.ts:339-346`**
`isLoadableEntryType` treats an absent `entry.type` as `script` because the platform declares the field optional; `collectStyles` requires `type === "style"` exactly, so a sheet with a `path` and no `type` is dropped, the plugin is still reported `loaded`, and its page renders unstyled — the outcome the comment at `:292-294` says the branch exists to prevent. `Logger.info` is the only signal and is a production no-op. Suggest treating an undeclared type the way `entry.type` is treated, e.g. `fileType ? fileType.toLowerCase() !== PLATFORM_STYLE_FILE_TYPE : !filePath.toLowerCase().endsWith(".css")`. Relatedly, `type` is never trimmed although `permission` is (`:661`), so `" script "` skips the whole plugin.

**5. The plugin-facing budget was narrowed from 5s to 3s, and the PR body doesn't say so — `index.ts:97-98`, `bootstrap.ts:38`**
Base: `manifest 3_000 / load 5_000 / backstop 20_000`. Head: `2_000 / 3_000 / 12_000` plus the new `DISCOVERY_TIMEOUT_MS = 2_000`. The invariant only needs backstop > discovery + manifest + 2×load, so 2+2+5+5 = 14 → a 16s backstop would keep the old 5s allowance and still improve on 20s. A plugin whose `init()` measured 4s in production now reports `failed`. The budgets section lists the new numbers without mentioning that two of them are reductions.

**6. The "CONTRACT GATE" JSDoc is now stranded above the wrong function — `index.ts:447-460`, `index.ts:577`**
Two JSDoc blocks sit stacked: the contract-gate block now documents `claimedRouteNames`, and `isCompatible` at `:577` is undocumented. At the merge base that block sat on `isCompatible`; the new functions were inserted between them.

**7. The `errorLoadRemote` comment is wrong about the mechanism — `index.ts:727-729`**
In `@module-federation/runtime-core/dist/remote/index.js` the catch does `if (!failOver) { … throw error }` and only then `return failOver`: a hook returning `null`/`undefined` re-throws, and only a **truthy** failover is resolved. The `plugin === null` check is still worth keeping, but its justification is the declared `loadRemote<T>(): Promise<T | null>` signature, not the behaviour described.

**8. The cascade-precedence comment states the direction backwards — `index.ts:314-315`**
"Wrapping only the plugin's sheet would put it **above** everything unlayered" — unlayered beats layered, so it would put the plugin **below** all host CSS, which is exactly what `specs/2026-08-21-plugin-css-cascade-layers.md:127-129` says in this same diff. The spec's premise checks out (host CSS is effectively unlayered), so it's just the one word — but this comment is what the VCST-5760 implementer will read.

**9. `APP_MODULES_FEDERATION_REMOTES="{}"` loses every deployed plugin in complete silence — `index.ts:438-443`, `app-runner.ts:76-77`, `index.ts:675-680`**
The "N platform plugin(s) are ignored" warning can't fire in the app: when the override is set, `ASK_PLATFORM_FOR_PLUGINS` is false, so the query never goes out and `plugins` is always empty. With `{}` there are then no remotes and no invalid names, so `reportOutcome` isn't called either — not one line, which is the precise failure the comment at `:439` names. Invalid JSON does log. Logging the override unconditionally (without the count) would restore it.

**10. The boot-sequence diagram describes the ordering this PR deliberately inverted — `README.md:200-202`**
It still reads `const ready = startFederatedModules(); // fire early, don't await yet` / `… other boot work …`, whereas at head the call is `app-runner.ts:318` with the `await` at `:324` and nothing between — and `app-runner.ts:314-317` explains why. The rest of the diagram was updated.

**11. A non-string `entry.path` is reported as an absent one — `index.ts:395-398`**
Observed live: `entry.path = 12345` produces `Skipping plugin "badShapePlugin": the platform declared no entry path`. The path is declared, it's just not a string, so the message sends the plugin author looking in the wrong place.

**12. The manifest is fetched twice per plugin per boot — `index.ts:600` plus the MF runtime's own load**
A clean single boot shows two requests for `mf-manifest.json?v={hash}`: the contract gate's, then the runtime's. Given the missing `Cache-Control` on the platform's static middleware that TODO.md already notes, that's a second real round trip inside the load budget.

**13. The one deferred item without a ticket is the one that makes all the others invisible — `TODO.md:213`, `bootstrap.ts:105`, `index.ts:618`**
`logger.prod.ts` noops all four levels, so the ~20 diagnostic exits in the loader produce nothing in production; the structured `IFederatedLoadResult` is the only other signal and `bootstrap.ts:105` discards it (nothing outside `index.ts` consumes the type). VCST-5760/5761/5762 all carry ids; this one is a markdown bullet, yet `index.ts:618` and `bootstrap.ts:35` describe it as "a tracked stage-2 follow-up".

**14. `core/api/graphql/types.ts` — the environment delta is disclosed as a removal only**
86 of the 90 added lines are `SalesRepDocument{,Category,Connection,Edge}`, `Query.salesRepDocument{,Categories,s}` and their `*Args` types, absent from `dev` and arriving from generating on vcptcore-dev. The body names the `CartType.loyaltyPoints` removal (inert today — nothing selects it; the other hits are untouched module files and `Product.loyaltyPoints` in the contract) and the sales-rep module file, but not this addition. Worth naming as the third delta, since regenerating on vcst-qa reverses all ~87 lines and every fork regenerates this file.

**15. Comment prose exceeds the logic it guards in a few places — `bootstrap.ts:14-36` and `:38-45`, `index.ts:165-171` and `:694-698`**
Measured across 207 non-test `.ts` files of ≥60 lines: median comment ratio 0%, `index.ts` 33%, `bootstrap.ts` 42%, and every file ranked above them is a types or constants file. `bootstrap.ts:14-36` is 22 comment lines on `BOOT_BACKSTOP_MS = 12_000`. The reasoning is genuinely good — the backstop arithmetic, the MF prefix rule and the no-`force` rationale would just serve readers better in `README.md` with a one-line pointer at each site.

**16. Discovery policy sits in the hottest fork-merge file — `app-runner.ts:76-77`, `:166`, `:170`; `bootstrap.ts:58-60`**
`ASK_PLATFORM_FOR_PLUGINS` re-derives "the env override wins" from `APP_MODULES_FEDERATION_REMOTES`, a rule `resolveRemotes` (`index.ts:433-445`) owns, and `isMfFlagEnabled` is evaluated here and again at `bootstrap.ts:78`. Change the precedence in the loader and app-runner still won't ask. Since `bootstrap.ts` is already statically imported at `app-runner.ts:33`, a `startPluginDiscovery(domain)` export holding the flag check, the env check and the `void … .catch` would leave app-runner two lines. Related: a no-arg `startFederatedModules()` — the shape a fork might land on when resolving this merge — disables discovery with no log at all, since `withDiscoveryBudget(undefined)` returns silently and `index.ts:675-680` skips `reportOutcome`; a `Logger.warn` in the `!fetchPlugins` branch costs nothing.

**17. Docs: the `vendor` layer is missing from both order statements, and the hand-written-CSS channel has no theming story — `TODO.md:77`, `specs/2026-08-21-plugin-css-cascade-layers.md:141`, `HOWTO.md:182-185`**
`spec:44` includes `vendor` and `spec:50` explains why (it ships as its own unlayered `<link>`, so it would outrank everything), but neither the TODO checklist order nor the v4 order carries it. And the new `contentFiles` stylesheet bullet gives plain CSS no theming guidance, while the Tailwind path just below it does — `var(--color-*)` works there too (`config.plugin.ts:7-37`), so one sentence would stop an author hardcoding a hex and breaking dark mode and six of seven presets.

**18. Minor lint/claim points**
"eslint clean on the touched files" holds for errors only: the diff adds 26 warnings, including ten `sonarjs/no-redundant-optional` in the new `IPlatformPlugin` (`index.ts:54-58`), `sonarjs/function-return-type` (`:188`), a `sonarjs/null-dereference` false positive (`:272`), and `import/order` at `boot-order.test.ts:1`. Also `boot-order.test.ts:92` mocks `@/i18n` without an `install`, so all six tests emit a Vue plugin warning — cosmetic; the harness does reach the real `startFederatedModules`, so the ordering claims are genuinely pinned.

Nothing here is a blocker. Things I suspected and had to withdraw: a host module's own `addRoute`/`removeRoute` cannot land inside the guard window (all 20 call sites in `client-app` checked; the guard is only installed once at least one compatible plugin exists); a cyclic plugin `children` throws in ~2ms rather than hanging boot; the discovery promise genuinely can't produce an unhandled rejection (probe: 0 with the `void … .catch` at `app-runner.ts:170`, 1 without, and the loader still sees the failure either way); `getStorePlugins` matches the `getStore` wrapper exactly and its placement matches the code-placement table; the vite proxy rule overlaps no route or other proxy key and `preview.proxy` shares the config; the 1.2.0 → 1.3.0 bump is additive; and the `InitializeApplication` rationale holds — I reproduced it against a live backend, where one unknown field returns `errors` with no `data` at all. The load-bearing third-party claims I checked in `node_modules` (prefix resolution of a `loadRemote` id, the `.json` manifest heuristic, no-`force` re-registration being a silent no-op, and `isRootAdd = !originalRecord` with children recursed first) are all correct.

ivan-kalachikov:
Re-verified each finding against the code before touching anything: 17 of 18 reproduce as written, two have overcounted sub-claims. Head is now `856fa6830`.

**Fixed**

1. Non-string `permission` fails closed. So do `entry.path` and `entry.type` — one `nonStringFieldReason` helper. This goes a step past the finding: I made a non-string `entry.type` a skip too, because otherwise `collectStyles`' new rule ("a drifted kind is not a missing one") contradicted `isLoadableEntryType` two screens down. Say the word if you'd rather keep it tolerant.
2. Test added: style file + a throwing `init()`. Reproduced the gap first — moving `injectStyles` after `runInit` left all 157 green.
3. Comment only: it now says only `id` is load-bearing. My tsc probe agrees — renaming `entry` or dropping `remote` compiles clean, only dropping `id` errors. Not deriving from codegen; that reverses the layer-independence decision.
4. Undeclared `contentFiles[].type` falls back to the extension. Entry types trimmed.
6. CONTRACT GATE block back on `isCompatible`.
7. Comment now cites the `Promise<T | null>` signature. Your reading of `if (!failOver) { throw error }` is right.
8. Now "below all host CSS".
9. Notice is no longer count-gated: `warn` when the override resolves to no remotes, `info` otherwise. Replaced the test — it was asserting a branch app-runner makes unreachable.
10. Diagram updated, plus a line on why the loader starts after the host plugins.
11. Separate message for a declared-but-non-string path.
13. Reworded to "open follow-up, not yet ticketed" in all three places; Jira id still to be filed.
14. Third delta named in the body.
16. Cheap half only: `withDiscoveryBudget` warns when no plugin-list source is passed.
17. `vendor` in both order statements; HOWTO now points plain-CSS authors at `var(--color-*)`.
18. Body says "0 errors, 22 warnings"; `import/order` autofixed; `install` added to the `@/i18n` mock, which kills the six Vue warnings.

Also reworked the drifted-field `it.each`: every row now records its actual outcome (`skipped`/`loaded`, and under which name) instead of only asserting the healthy plugin survived.

**Not fixing**

5. Budgets stay 2s/3s/12s — the tighter numbers are a deliberate call. Your 16s arithmetic is right if we ever want 5s back. The body now states they are reductions and that a >3s `init()` reports `failed`.
12. Already tracked, and better than a comment: TODO.md's "seed the validated manifest through the runtime's `fetch` loader hook" removes the second round trip and closes the TOCTOU window with it.
15. Ratios reproduce (index.ts 30.6%, bootstrap.ts 39.8%, median 0.7% over 244 files) but "every file ranked above them is a types or constants file" does not — `useProductSortings.ts` (32.9%) and `builder-preview/utils.ts` (30.8%) sit above `index.ts`. Keeping the reasoning at the guard sites; trimmed only `bootstrap.ts`'s 22 lines on one constant.
16 (the refactor). `startPluginDiscovery(domain)` forces a rewrite of `boot-order.test.ts`, which is this PR's main new evidence. Follow-up.
18 (the warnings). The 10 `sonarjs/no-redundant-optional` are `IPlatformPlugin` mirroring codegen's `Maybe<T>` on purpose; `null-dereference` is the known repo-wide false positive. One correction: 26 is the total on changed files — 2 are "file ignored" notices and the 8 `void-use` in app-runner predate this PR, so the net new is ~15.

On `validate:core-types`: same diagnosis as yours. The only diff is `tabindex`'s position in `VcButton`'s inferred defaults, so I reverted the regeneration rather than commit churn that flips back on the next machine.

goldenmaya (dismissed):
**Approving at `856fa6830`.** Re-reviewed as a delta against my round-1 head `60252959a`, and re-verified each closed finding rather than reading the diff for it.

Gates at this head: `eslint` on the changed files 24 problems / **0 errors** (22 rule warnings + 2 "file ignored" notices), `validate:types` clean, `build-only` clean, `vitest client-app/modules/federated` 8 files / **161** tests green, SonarCloud 0 open issues. `validate:core-types` behaves exactly as in round 1 — the contract diff is byte-identical, so the `VcButton` ordering artefact is unchanged and still not attributable to this PR.

**How I checked the fixes.** Each new test was mutation-tested in a scratch copy: moving `injectStyles` after `runInit` fails the new "styles a plugin whose `init()` throws" test; removing the `.css` extension fallback fails its test; removing the `withDiscoveryBudget` warn fails its test; restoring the count-gate on the override notice fails its test. For the permission fix I also re-proved it the way the finding was proved — live against a mock platform list: `[MF] Skipping plugin "nonStringPerm": the declared permission is not a string`, with **zero** manifest fetches where round 1 produced two, and the string-permission control still skipped before any network. `boot-order.test.ts` now runs 6 tests with 0 Vue plugin warnings, down from 6. The doc fixes (`vendor` in both layer orders, the `var(--color-*)` line in HOWTO, the JSDoc back on `isCompatible`, the corrected cascade direction, the boot diagram) all check out, and the extra "started only after every host plugin has installed" bullet in the README goes past what was asked.

**Three corrections to my own review — you were right on all three.**

1. *"The diff adds 26 warnings"* was wrong. Running `eslint` at the merge base shows 7 `sonarjs/void-use` and 2 `sonarjs/todo-tag` already present, so 26 was the total on changed files and net new was ~17 then, ~15 now. One refinement to your version: 7 of the 8 `void-use` predate the PR — the eighth is the `void storePluginsPromise?.catch(...)` this PR adds.
2. *"Every file ranked above them is a types or constants file"* was false, and it was the clause that made the finding persuasive. My own recomputation puts `shared/catalog/composables/useProductSortings.ts` (38.2%) and `core/utilities/search/facets.ts` (34.8%) above `modules/federated/index.ts` (33.2%). The median is method-dependent too — I measure 5.3%, you 0.7%. The ratio observation stands; the qualifier should not have shipped.
3. Finding 12 should never have been raised. `client-app/modules/federated/TODO.md:100-108` has tracked the double manifest fetch **and** the TOCTOU window since before this PR — it is present at the merge base — with the `fetch`-loader-hook fix. I read that file for its claims and never used it as the deferral register, so I re-raised a settled item and undersold it at the same time.

**Three non-blocking residuals, none worth holding the merge for.**

- `index.ts:456` — `nonStringFieldReason` does not cover `remote.exposed`: `asString(plugin?.remote?.exposed) || PLATFORM_EXPOSE_KEY` still turns a drifted value into `./Module` silently, which is the scaffold-plugin (`./plugin`) load failure from finding 3. The principle the new helper's JSDoc states — undefined and drifted must not look alike — applies here too. Nearby, `index.ts:293`'s `asString(file?.path)` drops a non-string path with no log at all.
- `index.ts:379-395` — the new distinct message for a declared-but-non-string `entry.path` has no test: deleting that branch of the helper fails nothing, because the outcome stays `skipped` and no assertion touches the message. The `permission` and `entry.type` branches are each pinned by a test; this one is the odd branch out.
- `index.ts:294-300` — `contentFiles[].type: ""` on a `.css` path is treated as declared-and-wrong rather than absent, so the sheet is dropped instead of falling back to the extension.

For the record, `bootstrap.test.ts:147` weakens from `.not.toHaveBeenCalled()` to `.not.toHaveBeenCalledWith(stringContaining("boot backstop"))`. That is the right call given the new no-source warning, and it still asserts what the test is named for — noting it only so the weakening is on the record.

**One condition I would attach to the approval rather than leave implicit.** A real platform-served plugin that actually loads and runs `init()` has still never executed — your own "not exercised" note, and I confirmed it: my stand reaches `loadRemote` and stops there. So the route guard, the stylesheet ordering in a live page, the `init()` receiver and the shared-dependency gate rest on unit tests alone. That is not a merge blocker, because `APP_MODULES_FEDERATION_ENABLED` is build-time and off by default — with the flag off I see no `[MF]` output and the loader chunk is never imported — but it is a reason not to turn the flag on in any environment before the pilot in `TODO.md` #1. What ships unconditionally is `core/api/graphql/types.ts`, `contract/index.d.ts` and the `CORE_VERSION` bump; I checked those: the `CartType.loyaltyPoints` removal is inert (nothing selects it), and 1.2.0 → 1.3.0 is additive so plugins pinned to `^1.2.0` still resolve.

Good round. The fixes went past the letter of several findings, and the reworked drifted-field `it.each` — every row now recording what that field's drift actually costs — is a better test than the one I asked for.

ivan-kalachikov:
All three residuals fixed at `8a1641226`, and the probe I used to re-verify them turned up two more of the same class.

**Your three**

- `remote.exposed` is in `nonStringFieldReason` now. Measured before fixing: `exposed: 7` gave `loaded=["sales-rep"]` with `loadRemote("sales-rep/Module")` and **zero** log lines. `remote.name` deliberately stays out — falling back to the id is the platform's own default and is tested.
- `contentFiles[].path` non-string logged nothing at any level; it now names the file it dropped.
- Blank `contentFiles[].type` falls back to the extension. Whitespace-only did the same thing and isn't in your list — `"  "` trims to `""` but `rawType != null`, so it took the strict branch too. Both covered.

**Two more the probe found**

- A non-string `entry.hash` or `contentFiles[].hash` produced a URL with **no `?v=`** and no log. Given the platform sets no `Cache-Control` — the TODO you cited — that cache-buster is the only freshness signal there is, so losing it silently is worse than it looks. `withCacheBuster` now takes `unknown`, owns the `asString`, and warns when it drops one.
- `response.url` absent or `""` skips the post-redirect origin re-check entirely (`if (servedFrom && …)`). A real browser `fetch` always sets it, so this needs a service worker or a synthesized `Response` — narrow enough that I did not want to fail closed unilaterally, since an environment that legitimately omits it would lose every plugin. Flagging rather than changing; say if you want `!servedFrom ||`.

**On the untested branch.** Confirmed exactly as you had it: deleting the `entry.path` branch left 161/161 green. Fixed by pinning all four branches to their messages instead of their outcome — each one now dies alone (`entry.path` 6 tests, `entry.type` 7, `permission` 7, `remote.exposed` 1). Same for the new code: dropping the no-usable-path log fails 1, reverting the blank-kind rule fails 2, dropping the cache-buster warn fails 1.

169 tests in the module (was 161), 2447 across `client-app`, `vue-tsc --build --force` clean, eslint 0 errors / 23 warnings. Body updated with the new counts.

Noted on the flag: nothing gets turned on anywhere before the pilot in TODO.md #1.

client-app/modules/federated/index.ts
```diff
@@ -171,11 +235,389 @@ function resolveRemotes(): IResolvedRemotes {
...
+      resolved.invalidNames.push(asString(plugin?.id) || name);
+      continue;
+    }
+    seen.add(name);
+    resolved.remotes.push({
+      name,
+      entry: withCacheBuster(manifestUrl, plugin?.entry?.hash),
+      exposed: asString(plugin?.remote?.exposed) || PLATFORM_EXPOSE_KEY,
+      permission: asString(plugin?.permission),
+      styles: collectStyles(plugin),
```
vas11yev1work, line 474: `resolvePlatformRemotes` calls `collectStyles(plugin)` for every plugin the platform returns — parsing/validating every content-file URL — before the permission filter (later, in `initFederatedModules`) throws out plugins the user isn't allowed to see. So permission-gated plugins (e.g. admin-only modules) pay the URL-validation cost even though they'll be filtered out anyway. Not a correctness bug, just wasted work.

Suggestion: pass `hasPermission` down into `resolveRemotes`/`resolvePlatformRemotes`, and skip `collectStyles` for any plugin whose permission fails the check — before doing any URL work for it, not after.

ivan-kalachikov: **Not doing it.** I implemented it to check, and it changes four things the suite does not catch — 169/169 still green with the change in.

Moving the gate above the URL work means a denied plugin leaves the loop before `seen.add(name)` and never gets validated. Measured across ten scenarios:

- **Who owns a remote name starts depending on the viewer.** Two descriptors both named `x`, the first admin-only: today the first claims it and the second is reported "already taken" — same for everyone. After, an admin loads the first and everyone else loads the second. The collision is a deployment error either way, but `remote.name` is the identity MF resolves `loadRemote` against, so the winner has to be viewer-independent.
- **A broken descriptor reports "requires permission" instead of the real cause** — no entry path, off-origin entry, unsupported entry type — to anyone lacking that permission.
- The name-collision warning stops reaching those users as well.
- `skipped` loses its version label (`x@3.1.0` → `x`), since `versions` is built from the resolved list.

There is also a report where the same name lands in both `loaded` and `skipped`.

All fixable, but well past a small change, and it buys ~0.6 ms of CPU — no network, the manifest fetch already sits behind the gate.

vas11yev1work (approved):

PR #2448, opened 2026-08-24
feat(VCST-5750): add SSO support to the login on behalf page

ivan-kalachikov:
## Description

The login on behalf page (`/account/impersonate/:userId`) only offered username/password verification, so an operator of a store that authenticates through identity providers had no way to start an impersonation session.

Identity providers now render on that page, with the same layout the sign-in page uses:

- **Password + providers** — the verification form stays in place, the providers are shown next to it behind the `Or` divider.
- **Providers only** — the form is hidden and the providers are rendered in the content column.
- The providers return to the impersonation URL (`returnUrl`), where the already existing authenticated flow picks the session up and starts the impersonation automatically. No backend change is involved: the `impersonate` grant is requested with the operator's token, exactly as for an operator who was already signed in.

Parts shared by the sign-in and the impersonate pages:

- `useIdentityProviders()` — the enabled-providers logic.
- `identity-providers.vue` — the provider buttons.
- `sign-in-divider.vue` — the `Or` divider, which each page composes with the providers.
- `useIdentityProvider()` takes the return URL as a required input; the provider components pass it down as a prop.
- `useReturnUrl()` resolves that URL — the requested page, then the store default — and the `openReturnUrl` handler reads it from there too.

The heading of the verification step lives on the page rather than in `impersonate-form.vue`, so it is also shown when the store has no password authentication — the same split the sign-in page already has.

`openReturnUrl` navigates through `toSameOriginPath`, the guard `useImpersonate` already applies. A `returnUrl` of the same host but a different origin (other port or scheme) used to be followed and now falls back to the home page. Nothing in the app produces such a value — the router guard and the callback both pass paths.

`getReturnUrlValue` returns `null` for a `returnUrl` that does not parse, the same as for one that points to another host: the value falls through to the next key and then to the default. It threw a `TypeError` before, on the sign-in page render, the router guard, the push-message page and the `openReturnUrl` handler.

The provider components are memoized with `useMemoize`. A component whose chunk fails to load is no longer dropped from the cache: a provider this theme does not ship never loads, and a chunk that fails to fetch is handled by the `vite:preloadError` recovery.

Tests cover the providers logic, the return URL resolution, the provider prop plumbing and both pages.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5750
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.57.0-pr-2448-243a-243a0c42.zip


client-app/core/utilities/common/index.ts
```diff
@@ -3,8 +3,10 @@ import type { RouteLocationRaw, RouteLocationNormalizedLoaded, RouteLocationNorm
 
 const RETURN_URL_KEYS = ["returnUrl", "ReturnUrl"] as const;
 
-export function getReturnUrlValue(): string | null {
-  const { searchParams, origin, hostname } = new URL(location.href);
+export function getReturnUrlValue(url?: string): string | null {
```
vas11yev1work, line 15: `getReturnUrlValue` can throw an uncaught `TypeError` if the `returnUrl` query value fails `new URL()` parsing (e.g. `?returnUrl=http://[not-valid-ipv6`). Every other "invalid returnUrl" case (wrong host) is treated as "not found" and falls back to a default — this one isn't, and it's now read inside a Vue computed on `sign-in.vue` (`returnUrl = computed(() => getReturnUrl(route.fullPath))`), so a malformed link breaks the sign-in page render for any anonymous visitor.

client-app/shared/sign-in/components/identity-providers.vue
```diff
@@ -9,19 +14,37 @@ import { defineAsyncComponent } from "vue";
...
+        // Keeping the failed component would hold the button back until a page reload.
+        providerComponents.delete(name);
+        fail();
+      },
+    });
+
+    providerComponents.set(name, component);
+  }
+
+  return component;
```
vas11yev1work: `loadProviderComponent`'s manual `Map` (get/set/delete) reimplements a pattern already used in this codebase via `useMemoize` from `@vueuse/core` (see `useModuleSettings.ts` and `useFiles.ts`), which also ships a built-in `.delete(key)` for exactly the "evict on load failure" case handled here by hand.

```typescript
const loadProviderComponent = useMemoize(
  (providerName: string): Component =>
    defineAsyncComponent<Component>({
      loader: () => import(`./${providerName.toLowerCase()}-provider.vue`),
      errorComponent: { render: () => null },
      onError(error, _retry, fail) {
        Logger.error(`Failed to load ${providerName} provider component`, error);
        loadProviderComponent.delete(providerName);
        fail();
      },
    }),
  { getKey: (providerName) => providerName.toLowerCase() },
);
```

client-app/broadcast.ts
```diff
@@ -154,7 +154,7 @@ export function setupBroadcastGlobalListeners() {
     });
   });
   on(openReturnUrl, () => {
-    location.href = getReturnUrlValue() ?? themeContext.value.settings.default_return_url ?? "/";
+    location.href = toSameOriginPath(getReturnUrl());
```
vas11yev1work, line 157: `getReturnUrl` already runs its result through `toSameOriginPath`

client-app/core/utilities/common/index.ts
ivan-kalachikov: **Fixed.**
- `parseUrl` helper returns `null` instead of throwing — applied to both parse sites
- a malformed value now falls through like a wrong host already did
- same fix covers 3 call sites that threw on `dev` too: `router/index.ts:61`, `push-message.vue:21`, `broadcast.ts:157`
- +5 tests; `new URL(location.href)` left unguarded on purpose

client-app/shared/sign-in/components/identity-providers.vue
ivan-kalachikov: **Applied, without the `.delete()`.**
- `.delete()` inside `onError` loops: `useMemoize` keeps its cache in `shallowReactive(new Map())` and the component is resolved from the template, so evicting a key retriggers the render, builds a fresh async component and fails again — measured 25 creations / 50 renders before an artificial cap, against 1/1 for the plain `Map`
- Dropped the eviction instead: it bought nothing. A provider this theme does not ship never loads, and a chunk that fails to fetch is already covered globally by the `vite:preloadError` recovery in `chunk-load-recovery.ts`
- Without it the memoized version settles at 1 creation / 2 renders — the extra pass is `cache.set()` running inside the render

client-app/broadcast.ts
ivan-kalachikov: **Correct, but keeping it.**
- `toSameOriginPath` is idempotent, so the outer call is a runtime no-op — agreed
- the suppression is textual: `pattern-not-regex: '(getReturnUrlValue|toSameOriginPath)\('` — `.semgrep/vc-security.yml:136`
- `location.href = getReturnUrl()` matches neither name → alert re-fires, plus the 2 CodeQL alerts currently `fixed` on this line

vas11yev1work (approved):

PR #2458, opened 2026-08-27
feat(VCST-5337): sales rep hub Google Analytics reports

yuskithedeveloper:
## Description

Adds the storefront half of **Customer insights: GA** — the Sales Rep Hub surfaces that show what a rep's customers actually do on the storefront, plus the tagging that makes those reports possible in the first place.

Two halves. **Write:** every GA4 event now carries the customer identity as user-scoped custom dimensions, so a GA report can be narrowed to one contact or organization. **Read:** a full **`/company/activities`** feed (My activity, or one customer's), a **My activity** dashboard widget, and a consolidated **Customer activity** widget on the customer profile.

Requires the backend PRs:
- https://github.com/VirtoCommerce/vc-module-google-ecommerce-analytics/pull/61 — reads GA4 through the Data API (`IAnalyticsService`)
- https://github.com/VirtoCommerce/vc-module-sales-rep/pull/15 — the `salesRepActivities` / `salesRepCustomerActivitySummary` / `salesRepCustomerInsights` queries

## Behaviour

- **Customer identity on every GA4 event** (`modules/google-analytics/user-properties.ts`): `contact_id`, `organization_id`, `organization_name`, `session_kind` and `is_sales_rep` go out with `gtag('set', 'user_properties', …)` **before the first `config`**, so the initial `page_view` is tagged too — GA never backfills an event that went out untagged. Every property is sent every time, `undefined` included: gtag *merges*, so an omitted key would carry the previous user's value into the next session. Re-applied when the user changes in place (another tab signs in), and pushed to `dataLayer` as well when a GTM container is configured, since GTM cannot read gtag's internal state. `is_sales_rep` reads raw permission membership, not `checkPermissions` — an administrator is granted everything and reporting them as a sales rep would be worse than reporting nothing. The five names must be registered by hand in GA4 Admin; the table is in [the module README](client-app/modules/google-analytics/README.md).

- **Activity feed** (`pages/activities.vue` → `/company/activities`): reps only, `?organizationId=` narrows it to one customer and the heading follows (**My activity** / **{Customer} activity**, the customer name linking back to their profile). Category tabs with counts — All, Orders, Customers, Searches, Product views, Logins — over a period filter, paged 20 at a time. Rows read as sentences: `Order placed #12345` with total and status, `New customer assigned {org}`, `Searched for “coffee” ×3`, `Viewed {product} ×3`, `3 sign-ins`. Zero-count tabs stay: a rep must see that a category exists and is quiet.

- **Tracked rows never claim more precision than they have**: analytics rows are hour buckets, so they render as *"27 Aug 2026, during the hour of 14:00"* (and `~2 hours ago` in the compact widget), `×N` is the count inside that hour, and a caveat names the approximation. Product rows link by **id** (`/product/{id}` always resolves, the tracked SEO segment alone does not) and fall back to plain text with the GA-recorded name when a code does not resolve in the store's catalog; search terms link to the catalog search page.

- **Top | Recent on the ranked categories**: the Searches and Product views tabs offer a mode chip that swaps the time-ordered feed for a ranked count list from `salesRepCustomerInsights`. Each Top list runs only while its tab shows it — no speculative fetching of a GA-backed query.

- **Customer activity widget** (`components/customer-activity.vue`): one widget instead of three stacked ones — **Summary** (Created on, Last web login, Visits, Last search term, Last viewed product) with **Searches** and **Product views** as sub-views behind chips, each with its own Top|Recent toggle. Sub-views mount on first visit and are `v-show`n after, so opening a customer profile fires **one** GA-backed query rather than three, and returning to a visited view refires nothing.

- **Two requests per view, not one** (`composables/useSalesRepActivities.ts`): counting every category is the expensive half — it reads the tracked ones from Google — so the badge counts run in their own request, which does not change as the rep switches tabs or turns pages, while the rows request selects no badges. A database-backed tab (Orders, Customers) therefore never waits on Google at all. The **selected** tab's badge still comes from the rows themselves, so the badge a rep is looking at can never disagree with the list under it.

- **Opens on This month, and All time sits last**: with no bounds the tracked categories are read from GA4's earliest supported date (2015) — a decade scanned to render one page. The period chips read as a progression (This month, This year, All time) rather than putting the widest option in front of the two it contains.

- **An hourglass marks the tracked figures** (`components/tracked-metric-hint.vue`): on the tabs that carry them, on All (which merges them in) and on the tracked rows of the customer summary, with a tooltip saying the one thing a reader can act on — the figure can take up to 48 hours to appear. Created on carries no mark; it is a platform fact.

- **Missing analytics is a state, never an error**: unconfigured stores, unauthorized organizations and empty ranges all arrive as nulls and zeros by contract, and each surface renders its own not-configured / empty / failed view instead of a blank card or a toast.

- **Reuse**: the shared `SalesRepRuleChips` gained an optional per-rule `tracked` mark and an `allLast` placement for the baseline tab — existing consumers are unchanged, and the component gained its first test file.

- **i18n**: new `sales_rep.activity.*` and `sales_rep.customer_insights.*` keys in all 13 locales.

## Screenshots

image
image
image

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5337
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.57.0-pr-2458-5838-5838ca87.zip


client-app/modules/sales-rep/components/sales-rep-rule-chips.vue
```diff
@@ -1,28 +1,21 @@
 <template>
   <div class="sales-rep-rule-chips">
-    <!-- Baseline tab: active when no rule is chosen; clicking it clears the filter. -->
+    <!-- The baseline tab (no rule name) renders in the same loop as the rules, so it can sit at
+         either end: clicking it clears the filter, and it is active while nothing is selected. -->
     <button
+      v-for="tab in tabs"
+      :key="tab.name ?? ''"
```
vas11yev1work, line 7: `:key="tab.name ?? ''"` can collide with the baseline tab's key if the backend ever sends `rule.name === ""` — `selectableFilterRules` only filters out `"all"`, not empty strings, so such a rule would get the same key (`""`) as the baseline tab (`name: undefined`).

Maybe something like:
```vue
:key="tab.name !== undefined ? `rule:${tab.name}` : 'baseline'"
```

Prefix keys by kind so the two key spaces can never overlap: `rule:${tab.name}` for rules, `'baseline'` for the baseline tab — no backend-supplied `rule.name` can ever collide with it, unlike a bare sentinel string.

client-app/modules/sales-rep/composables/useSalesRepBrowseHistory.ts
```diff
@@ -0,0 +1,71 @@
...
+  sort?: string | Ref<string | undefined> | (() => string | undefined);
+  periodFrom?: string | Ref<string | undefined> | (() => string | undefined);
+  periodTo?: string | Ref<string | undefined> | (() => string | undefined);
+  take?: number | Ref<number | undefined> | (() => number | undefined);
+  // Apollo's `enabled`: lets a surface that shows the list on demand skip the query until then.
+  enabled?: boolean | Ref<boolean>;
+};
+
+// Owns the browsedProducts half of the salesRepCustomerInsights op (VCST-5337).
+export function useSalesRepBrowseHistory(options: UseSalesRepBrowseHistoryOptionsType) {
```
vas11yev1work, line 25: `useSalesRepBrowseHistory.ts` and `useSalesRepSearchHistory.ts` are near-byte-identical: same options type, same `variables` construction (organizationId/sort/periodFrom/periodTo/take), same `useSalesRepHubQuery` call with `fetchPolicy`/`enabled`, same `onError`/`notConfigured`/`dataAsOf` logic — differing only in the GraphQL document, the error-log label, the payload field read, and the per-row mapping.

Fix: extract a shared `useSalesRepInsightsList(document, pickRows, logLabel, options)` factory that owns the common query/notConfigured/dataAsOf plumbing; each composable becomes a thin wrapper passing its own document and row mapper. Any future fix to the shared fetch-policy/notConfigured logic (e.g. the VCST-5586-style tweak) then lands once instead of drifting between two copies.

```typescript
export function useSalesRepInsightsList<TRow>(
  document: TypedDocumentNode<...>,
  pickRows: (payload) => TRow[],
  logLabel: string,
  options: UseSalesRepInsightsListOptionsType,
) {
  const variables = computed(() => ({ ... }));
  const { result, loading, error, onError } = useSalesRepHubQuery(document, variables, {
    fetchPolicy: HUB_FETCH_POLICY,
    enabled: options.enabled ?? true,
  });
  onError((err) => Logger.error(`[sales-rep] salesRepCustomerInsights ${logLabel} failed:`, err));
  const payload = computed(() => result.value?.salesRepCustomerInsights);
  const notConfigured = computed(() => Boolean(result.value) && !payload.value);
  const dataAsOf = computed(() => payload.value?.dataAsOf as string | undefined);
  const items = computed(() => pickRows(payload.value));
  return { items, notConfigured, dataAsOf, loading, error };
}
```

client-app/modules/google-analytics/index.ts
```diff
@@ -45,6 +47,17 @@ export async function init({ extendEvents, extendConfig, extendSet }: InitOption
...
+  const usesDataLayerVariables = Boolean(gtmContainerId);
+
+  // Before the `config` call below, not after: properties set afterwards miss the initial page_view, and
+  // GA never backfills an event that went out untagged.
+  applyUserProperties(usesDataLayerVariables);
+
+  // Login, organization switch and impersonation all end in a full navigation, so this looks redundant —
+  // but `setUser` also replaces the user in place when another tab signs in (userReloadEvent -> fetchUser),
+  // and a one-shot read would keep tagging every later event with the previous identity.
+  watch(userPropertiesKey, () => applyUserProperties(usesDataLayerVariables));
```
vas11yev1work, line 59: `watch(userPropertiesKey, ...)` uses a getter as the watch source: Vue calls `userPropertiesKey()` (which calls `buildUserProperties()` + `JSON.stringify`) to track the dependency, then the callback calls `applyUserProperties()`, which calls `buildUserProperties()` again — so the properties object is built twice per identity-change event (login/logout/org switch/impersonation).

Fix: wrap `buildUserProperties` in a `computed`, and pass its cached value into the callback instead of rebuilding inside `applyUserProperties`:

```ts
const userProperties = computed(buildUserProperties);
function applyUserProperties(properties: UserPropertiesType, pushToDataLayer: boolean): void {
  window.gtag("set", "user_properties", properties);
  // ...
}
applyUserProperties(userProperties.value, usesDataLayerVariables);
watch(userProperties, (properties) => applyUserProperties(properties, usesDataLayerVariables));
```

`computed` caches the built object and Vue diffs by reference, so the build only happens once per change, and the stringify-based key can go away entirely. Only wrinkle: `userPropertiesKey` is unit-tested directly (`user-properties.test.ts:137-163`), so those tests would need to move to testing `buildUserProperties`/the computed instead.

PR #2459, opened 2026-08-28
feat(VCST-5472): expose selection scope and headAttrs on VcTable #header slot

goldenmaya:
## Problem

  `#header` replaces the built-in header row entirely — selection cell included — but the
  slot got no scope. A custom header therefore could not:

  - render the leading selection cell, so every column shifted by one while body rows kept
    theirs, silently and with no error;
  - drive select-all, or read whether the page is fully / partially selected;
  - reproduce `vc-table__head--sticky`, the sole carrier of the header's sticky positioning,
    so `sticky-header` and `max-height` had no effect on a custom header.

  The last one was live in our own docs: `SlotsApiCustomHeader` hardcoded
  `class="vc-table__head"` and its sticky controls did nothing.

  ## What this adds

  `#header` now receives a scope:

  | Field | |
  |---|---|
  | `showSelectionColumn` | whether a selection cell must be rendered |
  | `selectionMode` | `"single"` \| `"multiple"` \| `undefined` |
  | `isAllSelected` / `isSomeSelected` | select-all checkbox state |
  | `canSelectAll` | `false` when no row on the page is selectable |
  | `toggleSelectAll` | no-op outside `multiple` |
  | `selectionColumnAttrs` | `v-bind` onto a `<th>` — matching width + sticky behaviour |
  | `headAttrs` | `v-bind` onto the `<thead>` — carries the sticky modifier |

  Plus a DEV-only console warning when a custom header renders one cell fewer than the body,
  which is the misalignment case above.

  Both attrs bags expose `class` as a **string**, not an array. A bare `v-bind="attrs"`
  compiles to `normalizeProps(guardReactiveProps(attrs))`; `guardReactiveProps` does not clone
  a plain object, so `normalizeProps` rewrites `class` in place on the computed's cached
  value. With a sibling attribute Vue routes through `mergeProps` and clones instead — which
  is why the built-in call sites never showed it while the documented usage was broken.

  ## Breaking change — needs a decision before merge

  `--vc-table-selected-bg` → `--vc-table-selected-bg-color`.

  The old name shipped in 2.53.0–2.56.0. A fork overriding it in `_custom.scss` gets no merge
  conflict, no build error and no lint failure — the selected-row highlight just reverts to
  the default in all 7 presets and both themes. `ui-kit/DEPRECATION.md` requires keeping the
  old name working through a deprecation period for renamed props/slots/events; CSS custom
  properties are not literally named there.

  Options: ship an alias (`var(--vc-table-selected-bg, …)` as the fallback tail), or accept
  the rename and call it out in the changelog.

  ## Verification

  - `yarn test:unit` 2101/2101; the table file 108 → 156 tests.
  - `yarn validate:types` clean; `yarn lint` 0 errors, warning set unchanged vs `dev`.
  - New tests mutation-checked: deleting the `maxHeight` disjunct, or the built-in `sr-only`
    label, fails exactly and only its own test.
  - Storybook, both new/updated stories: sticky header pins on scroll on a consumer-authored
    `<thead>`; 27 hit-test points across the header band show no body bleed-through in light
    or dark; `single` mode renders the `sr-only` label and keeps 3 `th` / 3 `td`.

  ## Notes

  - Additive against `dev`: the slot had no scope there, so all three new types are new-on-new
    and no fork can depend on them. The only non-additive change is the token rename above.
  - `SlotsApiCustomHeader` was swept to `v-bind="headAttrs"` so the file stops teaching the
    hardcoded class; rendering is byte-identical with that story's own args.
  - `SELECTION_COLUMN_WIDTH` stays the plain `3rem` it is on `dev` — an earlier revision
    turned it into a public `--vc-table-selection-cell-width` token; dropped as unrequested
    API.

  **Known, not fixed here:** VcTable never prunes `selection` when `selectionMode` changes,
  so `multiple` → `single` leaves every key selected and renders several checked radios.
  Pre-existing on `dev`; fixing it means emitting `update:selection` in response to a prop
  change, which is a public-behaviour change out of scope for this PR. Worked around in the
  story; worth its own ticket.

### Jira-link:

https://virtocommerce.atlassian.net/browse/VCST-5472
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.58.0-pr-2459-2f36-2f363489.zip

---

> [!NOTE]
> **Medium Risk**
> Touches public VcTable slot API, selection key semantics, and CSS custom properties that forks may override; behavior changes are mostly additive with alias compatibility for the old token.
> 
> **Overview**
> **VcTable `#header` now exposes a scoped API** so custom headers can mirror the built-in row: selection column attrs, select-all state/`toggleSelectAll`, and `headAttrs` for sticky header behavior. Stories and docs were updated to bind `headAttrs` on `<thead>` instead of hardcoding classes.
> 
> **DEV-only guards** warn when a custom header is missing `<thead>`, drops the selection cell while selection is on, or is one header cell short of the body (including after column changes or mobile→desktop flip). **Selection behavior** is tightened: id-less rows use `__row_<index>` keys, `toggleSelectAll` is gated to multiple mode and skips no-op commits, and single-mode header cells get an accessible label.
> 
> **Theming:** selected-row background moves to `--vc-table-selected-bg-color` with `--vc-table-selected-bg` kept in the fallback chain; dark theme only overrides `--selected-bg-alpha`. **DEPRECATION.md** documents the CSS-variable deprecation pattern and checklist.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f3634894b81d52eb0f9160f040529b5c236175f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


client-app/ui-kit/components/organisms/table/vc-table.vue
```diff
@@ -1026,7 +1109,7 @@ watch(
 
   --radius: var(--vc-table-radius, var(--vc-radius, 0.5rem));
   // Selected-row highlight; dark override bumps the alpha for dark surfaces.
-  --vc-table-selected-bg: rgb(from var(--color-primary-500) r g b / 0.08);
+  --vc-table-selected-bg-color: rgb(from var(--color-primary-500) r g b / 0.08);
```
ivan-kalachikov: `--vc-table-selected-bg` shipped in 2.53.0–2.56.0, so a fork overriding the old name loses the selected-row highlight silently — no conflict, no build error, no lint failure.

That said, the token is ~6 weeks old and row selection itself landed in the same commit that introduced it, so the population that could be affected is tiny. Accepting the rename looks like the right call — just call it out in the release notes.

If you'd rather keep it seamless, the alias you suggested does work, as long as it's chained at the **declaration** site (here and in `dark/organisms/vc-table.scss:3`) rather than at the two read sites — same shape as the `--radius` line just above:

```scss
--vc-table-selected-bg-color: var(--vc-table-selected-bg, rgb(from var(--color-primary-500) r g b / 0.08));
```

client-app/ui-kit/components/organisms/table/vc-table.vue
```diff
@@ -637,6 +645,45 @@ function syncRetryListener() {
...
+const bodyRendersSelectionCell = computed<boolean>(() =>
+  props.loading ? !slots["desktop-skeleton"] : hasColumnSlots.value,
+);
+
+function warnOnMisalignedCustomHeader(): void {
+  if (headerAlignmentWarned || !showSelectionColumn.value || !bodyRendersSelectionCell.value) {
+    return;
+  }
+
+  const rows = desktopTableRef.value?.querySelectorAll<HTMLTableRowElement>("thead tr");
```
ivan-kalachikov: nit: `thead tr` is a descendant selector, so a table a consumer renders inside a body cell contributes its own header rows here — a wider nested `<thead>` then wins the `Math.max` below and silences a genuinely short outer header.

Verified with a probe: misaligned outer header + a nested 3-cell table → 0 warnings; the identical setup without the nested table → 1.

`":scope > thead > tr"` (or `desktopTableRef.value?.tHead?.rows`) scopes it to this table.

client-app/ui-kit/components/organisms/table/vc-table.vue
```diff
@@ -879,6 +938,30 @@ function toggleSelectAll(): void {
...
+  toggleSelectAll,
+  selectionColumnAttrs: selectionColumnAttrs.value,
+  headAttrs: headAttrs.value,
+}));
+
+// Selection can be switched on after mount. Watching it instead of hooking `onUpdated` keeps
+// a correctly written header from re-querying the DOM on every render. `hasColumnSlots` is
+// deliberately not a source: reading it here would invoke the default slot outside render.
+watch(
+  [showSelectionColumn, () => props.loading],
```
ivan-kalachikov: nit: these two sources miss a `VcTableColumn` that mounts later behind a `v-if` — `stateColspan` changes, a previously aligned custom header is now short by one, and nothing re-runs the check. Same on a mobile → desktop flip, since the mount-time check bails while the desktop `<table>` doesn't exist.

Agreed on keeping `hasColumnSlots` out of the sources; `() => childColumns.value.size` would cover the column case without invoking the default slot.

client-app/ui-kit/components/organisms/table/vc-table.vue
goldenmaya: Verified the range: the old name is in 2.53.0–2.56.0, so I kept the alias.

Not the exact shape you suggested, though — declaring the public name on the block is what makes the *new* name unoverridable: `.vc-table` beats an inherited `:root`, so a fork's `:root { --vc-table-selected-bg-color: … }` would be dead and only the deprecated name configurable. Followed your `--radius` analogy instead, with a private target:

```scss
--selected-bg-alpha: 0.08;
--selected-bg-color: var(
  --vc-table-selected-bg-color,
  var(--vc-table-selected-bg, rgb(from var(--color-primary-500) r g b / var(--selected-bg-alpha)))
);
```

The alpha is factored out so `dark/organisms/vc-table.scss` overrides one line instead of repeating the chain. Checked in Chrome against the built CSS: 8%/16% unchanged, a `:root` override of either name applies in both themes, new name wins when both are set.

client-app/ui-kit/components/organisms/table/vc-table.vue
goldenmaya: Took `":scope > thead > tr"`, with a test that nests a VcTable inside a column slot: with the old selector the inner table's 4 header cells win the `Math.max` and the outer misalignment goes unreported.

client-app/ui-kit/components/organisms/table/vc-table.vue
goldenmaya: Both taken. Went past `size` to id lists, because a count misses a rename at constant count: a child column `name` plus `columns: [{ id: "name" }]` merges to 1 column; renaming the prop column to `other` makes it 2 while both counts stay 1. Tested.

Mobile → desktop confirmed: with a `#mobile-item` slot the desktop branch doesn't exist at mount, so the check bails on the missing ref without spending the once-per-instance latch, and afterwards no source changes. Added `isMobile` as a source, with a test — it drives the flip through `mobileBreakpoint`, since the suite's `useBreakpoints` mock isn't reactive.

client-app/ui-kit/components/organisms/table/vc-table.vue
```diff
@@ -863,22 +919,40 @@ function toggleRow(item: T, index: number): void {
...
     const keys = props.selection.map(String).filter((key) => !pageKeySet.has(key));
     commitSelection(keys, { action: "deselect-all" });
   } else {
-    const keys = props.selection.map(String);
-    for (const key of selectableKeysOnPage.value) {
-      if (!keys.includes(key)) {
-        keys.push(key);
-      }
+    // Dedup: two items can share a key, and `toggleRow` splices a single occurrence.
+    const added = [...new Set(selectableKeysOnPage.value)].filter((key) => !selectionSet.value.has(key));
```
vas11yev1work, line 966: The `[...new Set(...)]` dedupe in `toggleSelectAll` only patches the select-all path — `isRowSelected`, `toggleRow`, and `selectedCountOnPage` still key off `getItemKey`, so two rows that collide on the index-fallback key (one has `id`, the sibling doesn't) still cross-select on a plain checkbox click, not just on select-all. Worth fixing at the source (`getItemKey`) rather than patching each caller.

client-app/ui-kit/components/organisms/table/vc-table.vue
```diff
@@ -637,6 +645,42 @@ function syncRetryListener() {
...
+function warnOnMisalignedCustomHeader(): void {
+  // Only the default skeleton and the VcTableColumn rows inject selection cells.
+  const bodyHasSelectionCell = props.loading ? !slots["desktop-skeleton"] : hasColumnSlots.value;
+
+  if (headerAlignmentWarned || !showSelectionColumn.value || !bodyHasSelectionCell) {
+    return;
+  }
+
+  // Direct descendants only: column slots render consumer content, a nested table included.
+  const rows = desktopTableRef.value?.querySelectorAll<HTMLTableRowElement>(":scope > thead > tr");
```
vas11yev1work: `warnOnMisalignedCustomHeader` only matches `:scope > thead > tr`. A `#header` slot that returns bare `<tr>`s without a `<thead>` wrapper (easy to do by accident — Vue doesn't foster-parent them the way the HTML parser would) makes the selector match nothing, so the check silently no-ops instead of warning. Worth broadening the query or warning separately when no `<thead>` is found at all.

```typescript
const thead = desktopTableRef.value?.querySelector(":scope > thead");
if (!thead) {
  // eslint-disable-next-line no-console
  console.warn("VcTable: the custom `#header` slot must render a <thead> element.");
  return;
}
const rows = thead.querySelectorAll<HTMLTableRowElement>(":scope > tr");
```

vas11yev1work (approved):

PR #143, opened 2022-05-18
ST-1351: Control view mode using local storage

ilyawzrd:
Description:

--

QA-test: ST-1354

Demo-test: ST-1355

Download artifact URL:

PR #144, opened 2022-05-18
Feat/templates builder

basilkot:
Description:
extract app extensions to separated file to give template builder ability to add own plugins

--

QA-test:

Demo-test:

Download artifact URL:

PR #2338, opened 2026-06-18
feat(VCST-5292): frontend security CI

AndrewEhlo:
## VCST-5292 — Frontend security CI (Phase 1)

Adds security scanning tooling to the repo. **CI tooling only — no application code is changed.** Scanners run **non-blocking** to establish a baseline before gating is enabled later.

### What's added
- **Dependabot** — npm ecosystem (weekly, minor/patch grouped) to keep dependencies current.
- **Semgrep** (`.semgrep/vc-security.yml` + `security.yml` workflow) — 16 project-specific, class-based SAST rules, run alongside the `p/javascript`, `p/typescript`, `p/owasp-top-ten` registry packs.
- **OSV-Scanner** (`security.yml`) — SCA over `yarn.lock`.
- **CodeQL** (`codeql.yml`) — GitHub-native JS/TS analysis, security-extended queries.
- **ESLint** — `eslint-plugin-no-unsanitized` + `no-unsanitized/method`, `no-unsanitized/property`, `vue/no-v-html`, all at `warn`.

Findings upload as SARIF to the **Security** tab. All third-party actions are SHA-pinned (passes `pin-check`).

### Notes
- ESLint rules are `warn` (not `error`) and the scanners are non-blocking on purpose, so existing inherited findings don't break the build. No local impact: `yarn build`/`yarn dev`/pre-commit are unaffected.
- **Dependabot activates only after merge to the default branch**, not from this PR.

### Follow-ups (out of scope here)
- Enable secret scanning + push protection (repo Settings).
- Remediate flagged findings, then flip ESLint rules to `error` and make scanners blocking.
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5292
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.52.0-pr-2338-ff0c-ff0c14b3.zip

---

> [!NOTE]
> **Medium Risk**
> Touches impersonation post-login navigation (auth-sensitive) and the push notification service worker; CI changes are non-blocking but broaden the security surface reviewers must triage.
> 
> **Overview**
> **VCST-5292** adds automated security scanning in **non-blocking** mode (SARIF to the Security tab) and a set of targeted frontend fixes that align with the new Semgrep rules.
> 
[...]

PR #2460, opened 2026-08-31
fix(VCST-5771): harden three untrusted-input sinks in client-app

muller39:
## Description

Closes [VCST-5771](https://virtocommerce.atlassian.net/browse/VCST-5771). CodeQL and Semgrep flagged 5 alerts across 3 places where untrusted input reaches a navigation or query sink. One commit per sink so any single flow can be reverted independently.

- **Return-URL redirect** (`core/utilities/common/index.ts` — `getReturnUrlValue`): only accept genuine same-origin relative paths before assigning `location.href`, instead of a hostname comparison scanners don't recognize as a sanitizer. Normalizes backslashes and ASCII tab/newline first, mirroring the browser's own URL parsing, so a value like `/\t/evil.com` can't collapse into a protocol-relative `//evil.com` after the check already said "safe". Closes #143 (open redirect) and #144 (XSS). This also hardens the OIDC external sign-in redirect in `useIdentityProvider.ts`, which shares the same source function.
- **Order navigation** (`useOrderNavigation.ts`): add `"noopener"` to the `_blank` `window.open` call, closing #139 (missing noopener / reverse tabnabbing). The `router.resolve()`-based `location.href` branch (#140) is left as-is and should be dismissed in code scanning — that URL comes from an internal route name/id, not user input.
- **Organization search** (`useUserOrganizations.ts`): the search phrase escaped `"` but not `\`, so a trailing backslash broke out of the quoted filter-syntax string sent to the Contact/Member query. Closes #142. Extracted the escaping into `escapeFilterSyntaxValue()` in `core/utilities/search/facets.ts` and reused it from the two facet-filter call sites that already implemented the identical rule — three independent copies of the same escaping logic is how this bug happened in the first place.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5771

[VCST-5771]: https://virtocommerce.atlassian.net/browse/VCST-5771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.58.0-pr-2460-d04a-d04ac964.zip

### DIFF
--- a/client-app/core/utilities/common/index.test.ts
+++ b/client-app/core/utilities/common/index.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, afterEach, vi } from "vitest";
 import {
   getReturnUrlValue,
+  isSafeRelativeUrl,
   extractHostname,
   truncate,
   appendSuffixToFilename,
@@ -28,55 +29,66 @@ describe("getReturnUrlValue", () => {
     });
   });
 
-  it("should return the value of returnUrl parameter", () => {
-    // Mock location.href
+  it.each([
+    { case: "the returnUrl parameter", href: "http://example.com?returnUrl=/home", expected: "/home" },
+    {
+      case: "the ReturnUrl parameter (case-insensitive)",
+      href: "http://example.com?ReturnUrl=/dashboard",
+      expected: "/dashboard",
+    },
+    { case: "no returnUrl parameter present", href: "http://example.com", expected: null },
+    {
+      case: "returnUrl pointing to a different hostname",
+      href: "http://example.com?returnUrl=http://malicious.com/home",
+      expected: null,
+    },
+    {
+      case: "returnUrl as an absolute URL on the same hostname",
+      href: "http://example.com?returnUrl=http://example.com/home",
+      expected: null,
+    },
+    {
+      case: "returnUrl as a javascript: URL",
+      href: "http://example.com?returnUrl=javascript:alert(1)",
+      expected: null,
+    },
+    {
+      case: "returnUrl as a protocol-relative URL",
+      href: "http://example.com?returnUrl=//malicious.com",
+      expected: null,
+    },
+    {
+      case: "returnUrl with a tab hiding a protocol-relative bypass (browsers strip tabs before parsing)",
+      href: "http://example.com?returnUrl=/%09/malicious.com",
+      expected: null,
+    },
+  ])("should handle $case", ({ href, expected }) => {
     Object.defineProperty(window, "location", {
       configurable: true,
-      value: {
-        href: "http://example.com?returnUrl=/home",
-      },
+      value: { href },
     });
 
-    const result = getReturnUrlValue();
-    expect(result).toBe("/home");
+    expect(getReturnUrlValue()).toBe(expected);
   });
+});
 
-  it("should return the value of ReturnUrl parameter (case-insensitive)", () => {
-    // Mock location.href
-    Object.defineProperty(window, "location", {
-      configurable: true,
-      value: {
-        href: "http://example.com?ReturnUrl=/dashboard",
-      },
-    });
-
-    const result = getReturnUrlValue();
-    expect(result).toBe("/dashboard");
-  });
-
-  it("should return null when returnUrl parameter is not present", () => {
-    // Mock location.href
-    Object.defineProperty(window, "location", {
-      configurable: true,
-      value: {
-        href: "http://example.com",
-      },
-    });
-
-    const result = getReturnUrlValue();
-    expect(result).toBeNull();
-  });
-
-  it("should return null when returnUrl points to a different hostname", () => {
-    Object.defineProperty(window, "location", {
-      configurable: true,
-      value: {
-        href: "http://example.com?returnUrl=http://malicious.com/home",
-      },
-    });
-
-    const result = getReturnUrlValue();
-    expect(result).toBeNull();
+describe("isSafeRelativeUrl", () => {
+  it.each([
+    { case: "a relative path", url: "/account/orders", expected: true },
+    { case: "a relative path with query and hash", url: "/account?tab=orders#top", expected: true },
+    { case: "a protocol-relative URL", url: "//evil.com", expected: false },
+    { case: "an absolute https URL", url: "https://evil.com", expected: false },
+    { case: "a javascript: URL", url: "javascript:alert(1)", expected: false },
+    { case: "a backslash-then-slash bypass", url: "/\\evil.com", expected: false },
+    { case: "a slash-then-backslash bypass", url: "\\/evil.com", expected: false },
+    { case: "a double-backslash bypass", url: "\\\\evil.com", expected: false },
+    { case: "a tab-hidden protocol-relative bypass", url: "/\t/evil.com", expected: false },
+    { case: "a newline-hidden protocol-relative bypass", url: "/\n/evil.com", expected: false },
+    { case: "a carriage-return-hidden protocol-relative bypass", url: "/\r/evil.com", expected: false },
+    { case: "a leading tab hiding a protocol-relative bypass", url: "\t//evil.com", expected: false },
+    { case: "a path with no leading slash", url: "account/orders", expected: false },
+  ])("should return $expected for $case ($url)", ({ url, expected }) => {
+    expect(isSafeRelativeUrl(url)).toBe(expected);
   });
 });
 
--- a/client-app/core/utilities/common/index.ts
+++ b/client-app/core/utilities/common/index.ts
@@ -5,7 +5,7 @@ const RETURN_URL_KEYS = ["returnUrl", "ReturnUrl"] as const;
 
 export function isSafeRelativeUrl(url: string): boolean {
   // Mirrors the browser's tab/newline stripping + backslash normalization, or a hidden "//" could slip past.
-  const normalized = url.replaceAll(/[\t\n\r]/g, "").replaceAll(/\\/g, "/");
+  const normalized = url.replaceAll(/[\t\n\r]/g, "").replaceAll("\\", "/");
   return normalized.startsWith("/") && !normalized.startsWith("//");
 }
 
--- a/client-app/core/utilities/common/index.ts
+++ b/client-app/core/utilities/common/index.ts
@@ -5,7 +5,7 @@ const RETURN_URL_KEYS = ["returnUrl", "ReturnUrl"] as const;
 
 export function isSafeRelativeUrl(url: string): boolean {
   // Mirrors the browser's tab/newline stripping + backslash normalization, or a hidden "//" could slip past.
-  const normalized = url.replace(/[\t\n\r]/g, "").replace(/\\/g, "/");
+  const normalized = url.replaceAll(/[\t\n\r]/g, "").replaceAll(/\\/g, "/");
   return normalized.startsWith("/") && !normalized.startsWith("//");
 }
 
--- a/client-app/core/utilities/common/index.ts
+++ b/client-app/core/utilities/common/index.ts
@@ -3,17 +3,20 @@ import type { RouteLocationRaw, RouteLocationNormalizedLoaded, RouteLocationNorm
 
 const RETURN_URL_KEYS = ["returnUrl", "ReturnUrl"] as const;
 
+export function isSafeRelativeUrl(url: string): boolean {
+  // Mirrors the browser's tab/newline stripping + backslash normalization, or a hidden "//" could slip past.
+  const normalized = url.replace(/[\t\n\r]/g, "").replace(/\\/g, "/");
+  return normalized.startsWith("/") && !normalized.startsWith("//");
+}
+
 export function getReturnUrlValue(): string | null {
-  const { searchParams, origin, hostname } = new URL(location.href);
+  const { searchParams } = new URL(location.href);
 
   // Try each return URL key until we find one
   for (const key of RETURN_URL_KEYS) {
     const returnUrl = searchParams.get(key);
-    if (returnUrl) {
-      const returnUrlObj = new URL(returnUrl, origin);
-      if (returnUrlObj.hostname === hostname) {
-        return returnUrl;
-      }
+    if (returnUrl && isSafeRelativeUrl(returnUrl)) {
+      return returnUrl;
     }
   }
 
--- a/client-app/core/utilities/search/facets.test.ts
+++ b/client-app/core/utilities/search/facets.test.ts
@@ -10,11 +10,32 @@ import {
   getFilterExpressionForAvailableIn,
   getFilterExpressionFromFacets,
   generateFilterExpressionFromFilters,
+  escapeFilterSyntaxValue,
   termFacetToCommonFacet,
   rangeFacetToCommonFacet,
 } from "@/core/utilities";
 import type { RangeFacet, TermFacet, SearchProductFilterResult } from "@/core/api/graphql/types";
 import type { FacetItemType } from "@/core/types";
+
+describe("escapeFilterSyntaxValue", () => {
+  it("returns a plain value unchanged", () => {
+    expect(escapeFilterSyntaxValue("acme")).toBe("acme");
+  });
+
+  it("escapes double quotes", () => {
+    expect(escapeFilterSyntaxValue('acme "corp"')).toBe('acme \\"corp\\"');
+  });
+
+  it("escapes backslashes", () => {
+    expect(escapeFilterSyntaxValue("acme\\corp")).toBe("acme\\\\corp");
+  });
+
+  it("escapes a backslash immediately before a quote without under- or double-escaping", () => {
+    // Regression test: escaping quotes before backslashes would turn `\"` into `\\"`,
+    // leaving the quote unescaped and breaking out of the surrounding quoted string.
+    expect(escapeFilterSyntaxValue('acme\\"corp')).toBe('acme\\\\\\"corp');
+  });
+});
 
 describe("getFilterExpressionFromFacetRange", () => {
   it.each`
--- a/client-app/core/utilities/search/facets.ts
+++ b/client-app/core/utilities/search/facets.ts
@@ -23,7 +23,7 @@ import type { MaybeRef } from "vue";
  * {@link https://github.com/VirtoCommerce/vc-module-experience-api/blob/master/docs/filter-syntax.md#escaping-special-characters}
  */
 export function escapeFilterSyntaxValue(value: string): string {
-  return value.replaceAll(/\\/g, "\\\\").replaceAll(/"/g, String.raw`\"`);
+  return value.replaceAll("\\", "\\\\").replaceAll('"', String.raw`\"`);
 }
 
 /**
--- a/client-app/core/utilities/search/facets.ts
+++ b/client-app/core/utilities/search/facets.ts
@@ -17,6 +17,16 @@ import type { MaybeRef } from "vue";
  */
 
 /**
+ * Escapes backslashes and double quotes per the filter syntax's escaping rules, for embedding
+ * a value inside a filter-syntax expression (a facet filter clause or a quoted search phrase).
+ * Backslashes must be escaped first, otherwise the backslash added for the quote gets re-escaped.
+ * {@link https://github.com/VirtoCommerce/vc-module-experience-api/blob/master/docs/filter-syntax.md#escaping-special-characters}
+ */
+export function escapeFilterSyntaxValue(value: string): string {
+  return value.replace(/\\/g, "\\\\").replace(/"/g, '\\"');
+}
+
+/**
  * Generates a filter expression for category subtree filtering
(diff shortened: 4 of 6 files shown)


client-app/core/utilities/common/index.ts
```diff
@@ -3,17 +3,20 @@ import type { RouteLocationRaw, RouteLocationNormalizedLoaded, RouteLocationNorm
 
 const RETURN_URL_KEYS = ["returnUrl", "ReturnUrl"] as const;
 
+export function isSafeRelativeUrl(url: string): boolean {
+  // Mirrors the browser's tab/newline stripping + backslash normalization, or a hidden "//" could slip past.
+  const normalized = url.replaceAll(/[\t\n\r]/g, "").replaceAll("\\", "/");
+  return normalized.startsWith("/") && !normalized.startsWith("//");
+}
+
 export function getReturnUrlValue(): string | null {
```
ivan-kalachikov: **Blocker — this conflicts with `dev`, and the natural conflict resolution regresses VCST-5750.**

#2448 landed on `dev` and rewrote this function:

```ts
// dev
export function getReturnUrlValue(url?: string): string | null
```

`core/composables/useReturnUrl.ts:9` calls it *with* that argument, so taking this side of the conflict drops SSO login-on-behalf.

The bigger one: `isSafeRelativeUrl` **rejects** absolute same-origin URLs where `dev` **normalizes** them.

| `?returnUrl=` | dev | this PR |
| --- | --- | --- |
| `/account/orders` | `/account/orders` | `/account/orders` |
| `http://mystore.com/account/orders?page=2#list` | `/account/orders?page=2#list` | `null` → `default_return_url` |
| `//evil.com` | `/` | `null` |

Row 2 is pinned by `core/composables/useReturnUrl.test.ts:61` ("reduces a requested page of the current origin to a path"), so that test goes red on rebase.

client-app/core/utilities/common/index.ts
```diff
@@ -3,17 +3,20 @@ import type { RouteLocationRaw, RouteLocationNormalizedLoaded, RouteLocationNorm
 
 const RETURN_URL_KEYS = ["returnUrl", "ReturnUrl"] as const;
 
+export function isSafeRelativeUrl(url: string): boolean {
```
ivan-kalachikov: `toSameOriginPath()` already covers this, and it predates the PR — `core/utilities/same-origin/index.ts:7`, added in #2338, exported from the same `@/core/utilities` barrel.

Two things it does differently:

- delegates to `new URL()`, so the tab/backslash bypasses fall out of real WHATWG parsing instead of hand-rolled `replaceAll`
- returns a sanitized path rather than a boolean, so the caller gets something safe to *use*, not just a verdict

On `dev` the sink named in the ticket already goes through it:

```ts
// dev broadcast.ts:157
location.href = toSameOriginPath(getReturnUrl());
```

That also covers the `?? default_return_url` half of the expression, which this PR still assigns to `location.href` unchecked.

Since the commit series argues that three copies of the escaping rule is what caused the org-search bug, a second same-origin helper sitting next to the existing one seems worth avoiding.

client-app/core/utilities/common/index.ts
```diff
@@ -3,17 +3,20 @@ import type { RouteLocationRaw, RouteLocationNormalizedLoaded, RouteLocationNorm
...
   // Try each return URL key until we find one
   for (const key of RETURN_URL_KEYS) {
     const returnUrl = searchParams.get(key);
-    if (returnUrl) {
-      const returnUrlObj = new URL(returnUrl, origin);
-      if (returnUrlObj.hostname === hostname) {
-        return returnUrl;
-      }
+    if (returnUrl && isSafeRelativeUrl(returnUrl)) {
+      return returnUrl;
```
ivan-kalachikov: The guard checks the normalized string, but the raw one is returned:

```
input       /foo%09bar%5Cbaz
decoded     /foo<TAB>bar\baz
checked     /foobar/baz        ✅ passes
returned    /foo<TAB>bar\baz   ← raw
```

`location.href` re-normalizes, but `router.push({ path })` and `next()` route on the literal string. Returning `normalized` (or `toSameOriginPath`'s path) keeps what was checked and what is handed out identical.

One caller down, related: `push-message.vue:25` does `router.push({ path: returnUrl })`, which percent-encodes `?` and `#` into the path — and line 78 of the new suite blesses `/account?tab=orders#top` as safe. Pre-existing, but the tests now guarantee that input reaches it.

client-app/shared/account/composables/useOrderNavigation.ts
```diff
@@ -11,7 +11,7 @@ export function useOrderNavigation() {
     const orderRoute = router.resolve({ name: "OrderDetails", params: { orderId: order.id } });
 
     if (browserTarget.value === BrowserTargetType.BLANK) {
-      globalThis.open(orderRoute.href, "_blank")?.focus();
+      globalThis.open(orderRoute.href, "_blank", "noopener")?.focus();
```
ivan-kalachikov: `noopener` in `windowFeatures` makes `open()` return `null` per spec, so `?.focus()` can never run. The focus this line was written for is silently gone and the `?.` now only hides dead code — worth dropping it explicitly if it isn't needed.

Also worth knowing: a non-empty features string nudges some engines (Safari) toward a popup instead of a tab.

Separately — alert #140 on line 16 is still `open` in code scanning. The PR body says it should be dismissed, but nobody has.

ivan-kalachikov: `orderRoute.href` comes from `router.resolve()` on an internal route — the same reasoning the PR body uses to argue #140 one line down should be dismissed. Meanwhile the `window.open` that actually hands a third party a live `window.opener` is untouched:

```ts
// shared/layout/components/header/_internal/mobile-menu/mobile-menu-link.vue:73
if (isExternalLink.value) {
  window.open(props.link.route as string, "_blank")?.focus();
}
```

CMS-authored external URL, no `noopener` — the real instance of #139's class. Outside this ticket's alert list, so a follow-up is fine.

Heads-up if you do extend it: `quotes.vue:205` uses `window.open(...)!.focus()`, which would throw once `open()` starts returning `null`.

client-app/core/utilities/search/facets.ts
```diff
@@ -16,6 +16,16 @@ import type { MaybeRef } from "vue";
...
  */
 
+/**
+ * Escapes backslashes and double quotes per the filter syntax's escaping rules, for embedding
+ * a value inside a filter-syntax expression (a facet filter clause or a quoted search phrase).
+ * Backslashes must be escaped first, otherwise the backslash added for the quote gets re-escaped.
+ * {@link https://github.com/VirtoCommerce/vc-module-experience-api/blob/master/docs/filter-syntax.md#escaping-special-characters}
+ */
+export function escapeFilterSyntaxValue(value: string): string {
+  return value.replaceAll("\\", "\\\\").replaceAll('"', String.raw`\"`);
```
ivan-kalachikov: Good extraction — it just stops short of three of the five interpolations in this file:

```
:79   getFilterExpressionForAvailableIn   ❌ branch ids (reach it from localStorage, category.vue:520)
:88   getFilterExpressionForBrand         ❌ brandName
:99   getFilterExpressionFromFacets       ✅ fixed here
:113  "${facet.paramName}"                ❌ key raw, value escaped — same line
:127  generateFilterExpressionFromFilters ✅ fixed here
:131 / :144  "${filter.name}"             ❌ key raw
```

`:88` with a brand named `Acme "Pro"`:

```
"BRAND":"Acme "Pro""
             ↑ closes the literal early
```

Same break as the org-search bug. The value of the helper is that the rule lives in one place, so the remaining raw copies will re-diverge.

client-app/core/utilities/common/index.test.ts
```diff
@@ -28,55 +29,66 @@ describe("getReturnUrlValue", () => {
...
+    { case: "a javascript: URL", url: "javascript:alert(1)", expected: false },
+    { case: "a backslash-then-slash bypass", url: "/\\evil.com", expected: false },
+    { case: "a slash-then-backslash bypass", url: "\\/evil.com", expected: false },
+    { case: "a double-backslash bypass", url: "\\\\evil.com", expected: false },
+    { case: "a tab-hidden protocol-relative bypass", url: "/\t/evil.com", expected: false },
+    { case: "a newline-hidden protocol-relative bypass", url: "/\n/evil.com", expected: false },
+    { case: "a carriage-return-hidden protocol-relative bypass", url: "/\r/evil.com", expected: false },
+    { case: "a leading tab hiding a protocol-relative bypass", url: "\t//evil.com", expected: false },
+    { case: "a path with no leading slash", url: "account/orders", expected: false },
+  ])("should return $expected for $case ($url)", ({ url, expected }) => {
```
ivan-kalachikov: `$url` interpolates raw control characters into the generated test names — tab on `:85`, LF on `:86`, CR on `:87`. The newline splits the case across lines in the console reporter and puts a literal control character into the JUnit XML. `$case` already describes each row, so dropping `($url)` or escaping the value is enough.

client-app/core/utilities/search/facets.test.ts
```diff
@@ -10,12 +10,33 @@ import {
...
+  it("escapes double quotes", () => {
+    expect(escapeFilterSyntaxValue('acme "corp"')).toBe('acme \\"corp\\"');
+  });
+
+  it("escapes backslashes", () => {
+    expect(escapeFilterSyntaxValue("acme\\corp")).toBe("acme\\\\corp");
+  });
+
+  it("escapes a backslash immediately before a quote without under- or double-escaping", () => {
+    // Regression test: escaping quotes before backslashes would turn `\"` into `\\"`,
```
ivan-kalachikov: This narrates an implementation that was never committed. The test name already states the contract and `facets.ts:22` carries the same rationale — "why not the other way round" reads better in the commit message than in the file.

client-app/core/utilities/common/index.ts
ivan-kalachikov: **Blocker — this conflicts with `dev`, and the natural conflict resolution regresses VCST-5750.**

#2448 landed on `dev` and rewrote this function:

```ts
// dev
export function getReturnUrlValue(url?: string): string | null
```

`core/composables/useReturnUrl.ts:9` calls it *with* that argument, so taking this side of the conflict drops SSO login-on-behalf.

The bigger one: `isSafeRelativeUrl` **rejects** absolute same-origin URLs where `dev` **normalizes** them. Comparing the end of the chain, `useReturnUrl().getReturnUrl()`:

| `?returnUrl=` | on `dev` | after this PR |
| --- | --- | --- |
| `/account/orders` | `/account/orders` | `/account/orders` |
| `http://mystore.com/account/orders?page=2#list` | `/account/orders?page=2#list` | falls back to `default_return_url` |
| `//evil.com` | falls back | falls back |

Row 2 is pinned by `core/composables/useReturnUrl.test.ts:61` ("reduces a requested page of the current origin to a path"), so that test goes red on rebase.

client-app/core/utilities/common/index.ts
ivan-kalachikov: `toSameOriginPath()` already covers this, and it predates the PR — `core/utilities/same-origin/index.ts:7`, added in #2338, exported from the same `@/core/utilities` barrel.

Two things it does differently:

- delegates to `new URL()`, so the tab/backslash bypasses are caught by the origin check that falls out of real WHATWG parsing, instead of hand-rolled `replaceAll`
- returns a sanitized path rather than a boolean, so the caller gets something safe to *use*, not just a verdict

On `dev` the sink named in the ticket already goes through it:

```ts
// dev broadcast.ts:157
location.href = toSameOriginPath(getReturnUrl());
```

That also covers the `?? default_return_url` half of the expression, which this PR still assigns to `location.href` unchecked.

Worth rebasing for on its own: the copy on this branch returns `//evil.com` for a same-origin URL whose *path* starts with `//` (e.g. `https://mystore.com//evil.com/p`) — #2448 added the `path.startsWith("//")` guard that closes it.

Switching also removes a mismatch this version has: the guard checks the *normalized* string but the raw one is returned, so `?returnUrl=/foo%09bar%5Cbaz` validates as `/foobar/baz` and is handed out with the tab and backslash intact. `location.href` re-normalizes; `router.push({ path })` and `next()` route on the literal string.

Unrelated to which helper wins, one caller down: `push-message.vue:25` does `router.push({ path: returnUrl })`, which percent-encodes `?` and `#` into the path — and `index.test.ts:78` now blesses `/account?tab=orders#top` as safe. Pre-existing, but the tests now guarantee that input reaches it.

Since the commit series argues that three copies of the escaping rule is what caused the org-search bug, a second same-origin helper sitting next to the existing one seems worth avoiding.

client-app/shared/account/composables/useOrderNavigation.ts
ivan-kalachikov: `noopener` in `windowFeatures` makes `open()` return `null`, so `?.focus()` can never run. The focus this line was written for is silently gone, and the `?.` now only hides dead code — worth dropping it explicitly if it isn't needed.

Verified rather than assumed — [window open steps](https://html.spec.whatwg.org/multipage/nav-history-apis.html#window-open-steps) step 18, plus a run in Chrome behind a real user gesture:

| call | returns |
| --- | --- |
| `open(url, "_blank")` | `Window` |
| `open(url, "_blank", "noopener")` | `null` |

Both still open as a tab, so the features string doesn't turn this into a popup.

Separately: alert #140 on line 16 is still `open` in code scanning. The PR body says it should be dismissed, but nobody has.

ivan-kalachikov: `orderRoute.href` comes from `router.resolve()` on an internal route — the same reasoning the PR body uses to argue that #140, one line down, should be dismissed. Meanwhile the `window.open` that actually hands a third party a live `window.opener` is untouched:

```ts
// shared/layout/components/header/_internal/mobile-menu/mobile-menu-link.vue:73
if (isExternalLink.value) {
  window.open(props.link.route as string, "_blank")?.focus();
}
```

CMS-authored external URL, no `noopener`. Confirmed in Chrome: a page opened that way reports `window.opener` as present, so the reverse-tabnabbing handle is real. That's #139's class, on a line the alert didn't point at — outside this ticket's list, so a follow-up is fine.

Heads-up if you do extend it: `quotes.vue:205` uses `window.open(...)!.focus()`, which would throw once `open()` starts returning `null`.

client-app/core/utilities/search/facets.ts
ivan-kalachikov: Good extraction — two call sites in this same file still interpolate raw:

- `:88` `getFilterExpressionForBrand` — `` `"BRAND":"${unref(brandName)}"` ``
- `:79` `getFilterExpressionForAvailableIn` — branch ids, which reach it from localStorage (`category.vue:520`)

`:88` with a brand named `Acme "Pro"`:

```
"BRAND":"Acme "Pro""
             ↑ closes the literal early
```

Same break as the org-search bug. The point of the helper is that the rule lives in one place, so the leftovers will re-diverge.

client-app/core/utilities/common/index.ts
ivan-kalachikov, line 12: This conflicts with `dev`: #2448 changed the signature to `getReturnUrlValue(url?: string)`, and `core/composables/useReturnUrl.ts:9` calls it with that argument.

The behavioural half matters more — `isSafeRelativeUrl` rejects absolute same-origin URLs, where `dev` reduces them to a path:

| `?returnUrl=` | on `dev` | here |
| --- | --- | --- |
| `/account/orders` | `/account/orders` | `/account/orders` |
| `https://mystore.com/account/orders?page=2#list` | `/account/orders?page=2#list` | falls back to `default_return_url` |
| `//evil.com` | falls back | falls back |

`core/composables/useReturnUrl.test.ts:61` pins row 2, so it fails once this is rebased. Worth deciding which rule is intended before resolving the conflict.

client-app/core/utilities/common/index.ts
ivan-kalachikov, line 6: `toSameOriginPath()` already covers this — `core/utilities/same-origin/index.ts:7`, exported from the same `@/core/utilities` barrel — and CodeQL accepts it as a sanitizer, which is the deciding difference:

| ref | guard at the sink | #143 | #144 |
| --- | --- | --- | --- |
| `refs/heads/dev` | `toSameOriginPath` | fixed | fixed |
| `refs/pull/2460/merge` | `isSafeRelativeUrl` | open | open |

Both alerts close on their own on `dev`, while this branch still reports them — which is what made the 31 Aug dismissal necessary. Two things follow from that:

- #144 was a **true** positive. The old hostname check accepted `javascript://mystore.com/%0a<payload>` — a non-special scheme still parses a host, so `hostname` matched — and that executes when assigned to `location.href`. Dismissing it as a false positive suppresses the alert for good, including if the guard is weakened later. Reopening both and letting the rebase close them is safer.
- The tab and backslash bypasses given as the rationale were already rejected before this change: `new URL()` performs that normalization itself, so the hostname comparison caught `//evil.com`, `/\evil.com` and `/\t/evil.com`. The actual hole was the scheme-with-host case above.

`toSameOriginPath` also covers the `?? default_return_url` half of `broadcast.ts:157`, which this branch still assigns unchecked.

ivan-kalachikov (commented):
Reviewed — detail in the inline notes. Short version:

**Rebase onto `dev` first.** #2448 landed after this branch: it changed `getReturnUrlValue` to take a `url?` argument, and routes these sinks through the existing `toSameOriginPath()`. Two of the four notes fold into that one change, and without it `core/composables/useReturnUrl.test.ts:61` fails — `isSafeRelativeUrl` rejects the absolute same-origin return URLs that `dev` reduces to a path.

Worth knowing before choosing between the two guards: CodeQL clears #143 and #144 on `dev`, but still reports both on this branch's merge ref. `toSameOriginPath` is accepted as a sanitizer; `isSafeRelativeUrl` is not. That also makes #144 a true positive rather than a false one, so reopening both and letting the rebase close them is safer than leaving them dismissed.

Outstanding either way: #140 needs the dismissal the description mentions, and `?.focus()` after `window.open(…, "noopener")` is dead — `open()` returns `null`.

The escaping fix is correct, ordering included, and the `acme\"corp` test pins exactly the right case.

client-app/shared/account/composables/useOrderNavigation.ts
ivan-kalachikov, line 14: `open()` returns `null` once `noopener` is set ([window open steps](https://html.spec.whatwg.org/multipage/nav-history-apis.html#window-open-steps), step 18), so `?.focus()` cannot run. Either drop it, or raise the tab another way.

#139's rule text asks for `"noopener,noreferrer"`; only `noopener` is passed here.

#140 on line 16 is still `open` in code scanning — it needs the dismissal the description mentions.

ivan-kalachikov: `orderRoute.href` comes from `router.resolve()` on an internal route — the same reasoning the PR body uses to argue that #140, one line down, should be dismissed. Meanwhile the `window.open` that actually hands a third party a live `window.opener` is untouched:

```ts
// shared/layout/components/header/_internal/mobile-menu/mobile-menu-link.vue:73
if (isExternalLink.value) {
  window.open(props.link.route as string, "_blank")?.focus();
}
```

CMS-authored external URL, no `noopener`. Confirmed in Chrome: a page opened that way reports `window.opener` as present, so the reverse-tabnabbing handle is real. That's #139's class, on a line the alert didn't point at — outside this ticket's list, so a follow-up is fine.

Heads-up if you do extend it: `quotes.vue:205` uses `window.open(...)!.focus()`, which would throw once `open()` starts returning `null`.

client-app/core/utilities/search/facets.ts
ivan-kalachikov, line 26: Optional, not a change request: two other interpolations in this file still build the same quoted expression by hand —

- `:88` `getFilterExpressionForBrand` — `` `"BRAND":"${unref(brandName)}"` ``
- `:79` `getFilterExpressionForAvailableIn` — branch ids, which arrive from localStorage (`category.vue:520`)

Outside #142's scope so fine to leave, but the helper's value is that the rule lives in one place.

muller39:
Thanks for the detailed review — addressed all four points:
- Rebased onto `dev` (24af04957): dropped `isSafeRelativeUrl`, adopted `toSameOriginPath`. Reopened #143/#144 — now show `fixed` via dev's CodeQL-recognized guard.
- Dismissed #140 (value comes from `router.resolve()`, not user input).
- Dropped the dead `?.focus()`, added `noreferrer` (5f7187847).
- Also applied the optional `facets.ts` suggestion — `getFilterExpressionForBrand`/`AvailableIn` now reuse `escapeFilterSyntaxValue` (d04ac9645).

ivan-kalachikov (approved):

PR #147, opened 2022-05-19
fix: make vue version fixed

trueboroda:
Description:
to avoid warning messages like 'do not use built-in elements name as  the component id'

--

QA-test:

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.1.0-pr-147-676c66b0.zip

PR #222, opened 2022-07-05
ST-2417: Change menu item for compare products, fix badge position

ilyawzrd:
Description:
https://virtocommerce.atlassian.net/browse/ST-2417

--

QA-test: ST-2419

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.5.0-pr-222-dc677713.zip

PR #221, opened 2022-07-05
ST-2132: Limit error size to 2 lines on bulk order

ilyawzrd:
Description:

--

QA-test: ST-2134

Demo-test:

Download artifact URL: https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-1.4.0-pr-221-ea3edee6.zip

PR #2461, opened 2026-08-31
fix(VCST-5772): clear scanner noise — 3 false positives and 8 dev-script findings

muller39:
## Description
- Fixes the `vc-postmessage-no-origin-check` Semgrep rule: the deep-expression operator never reached into a nested `if`-condition, so the real origin guards in `builder-preview.plugin.ts` and `fcm-service-worker-v1.12.js` slipped past it (false positives [#217](https://github.com/VirtoCommerce/vc-frontend/security/code-scanning/217), [#221](https://github.com/VirtoCommerce/vc-frontend/security/code-scanning/221), [#222](https://github.com/VirtoCommerce/vc-frontend/security/code-scanning/222)).
- Closes the 8 CodeQL findings in `scripts/` (build-only dev tooling, never reaches a browser):
  - TLS bypasses ([#145](https://github.com/VirtoCommerce/vc-frontend/security/code-scanning/145), [#146](https://github.com/VirtoCommerce/vc-frontend/security/code-scanning/146)) are already gated to localhost/dev-only — annotated with CodeQL suppression comments.
  - Log injection ([#148](https://github.com/VirtoCommerce/vc-frontend/security/code-scanning/148)-[#152](https://github.com/VirtoCommerce/vc-frontend/security/code-scanning/152)) — response/error data is now stripped of control characters before it reaches `console.log`/`console.error`.
  - Write-to-file ([#147](https://github.com/VirtoCommerce/vc-frontend/security/code-scanning/147)) — the resolved locale file path is validated to stay inside the locale folder before the translated content is written back.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5772
Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.58.0-pr-2461-ea8a-ea8a8f4d.zip

### DIFF
--- a/.semgrep/vc-security.yml
+++ b/.semgrep/vc-security.yml
@@ -42,7 +42,7 @@ rules:
           - pattern: sessionStorage.setItem($KEY, $TOK)
       - metavariable-regex:
           metavariable: $KEY
-          regex: '(?i).*(auth|token|session|jwt|credential).*'
+          regex: "(?i).*(auth|token|session|jwt|credential).*"
     message: >
       Auth/refresh tokens in local/session storage are XSS-exfiltratable. Store the
       refresh token in an HttpOnly cookie and keep the access token in memory only.
@@ -198,12 +198,44 @@ rules:
           - pattern: '$X.addEventListener("message", function ($E) { ... })'
           - pattern: '$X.addEventListener("message", async ($E) => { ... })'
           - pattern: '$X.addEventListener("message", async function ($E) { ... })'
-      # An $E.origin check anywhere in the handler body clears it — including async
-      # handlers, which Semgrep treats as structurally distinct from sync ones.
+      # An $E.origin check anywhere in the handler body clears it — including
+      # async handlers, which Semgrep treats as structurally distinct from sync
+      # ones. Two exclusion shapes are kept side by side (a match is cleared if
+      # it satisfies either): the original bare deep-expression check below
+      # covers a guard written as a plain expression/ternary anywhere in the
+      # body, while the pattern-not-inside further down additionally covers a
+      # guard written inside an `if` condition — the deep-expression operator
+      # placed directly on the body never reaches into a nested if-condition,
+      # which is why a real `if (event.origin !== trusted) return;` guard used
+      # to slip past the bare check alone.
       - pattern-not: '$X.addEventListener("message", ($E) => { <... $E.origin ...> })'
       - pattern-not: '$X.addEventListener("message", function ($E) { <... $E.origin ...> })'
       - pattern-not: '$X.addEventListener("message", async ($E) => { <... $E.origin ...> })'
       - pattern-not: '$X.addEventListener("message", async function ($E) { <... $E.origin ...> })'
+      - pattern-not-inside: |
+          $X.addEventListener("message", ($E) => {
+            ...
+            if (<... $E.origin ...>) { ... }
+            ...
+          })
+      - pattern-not-inside: |
+          $X.addEventListener("message", function ($E) {
+            ...
+            if (<... $E.origin ...>) { ... }
+            ...
+          })
+      - pattern-not-inside: |
+          $X.addEventListener("message", async ($E) => {
+            ...
+            if (<... $E.origin ...>) { ... }
+            ...
+          })
+      - pattern-not-inside: |
+          $X.addEventListener("message", async function ($E) {
+            ...
+            if (<... $E.origin ...>) { ... }
+            ...
+          })
     message: >
       'message' handler must validate event.origin against a trusted allowlist
       before acting on the payload.
--- a/scripts/fix-missing-locales.ts
+++ b/scripts/fix-missing-locales.ts
@@ -21,13 +21,27 @@ function delay(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
+// localeFolder/targetFile both trace back to CLI-supplied glob patterns (see
+// check-locales-missing-keys.ts), so before writing the translated result back
+// out, pin the resolved destination to stay inside the resolved locale folder.
+function resolveLocaleJsonPath(localeFolder: string, filePath: string): string {
+  const resolvedFolder = path.resolve(localeFolder);
+  const resolvedFilePath = path.resolve(filePath);
+  // eslint-disable-next-line sonarjs/null-dereference -- false positive: path.resolve always returns a string
+  if (!resolvedFilePath.startsWith(resolvedFolder + path.sep) || path.extname(resolvedFilePath) !== ".json") {
+    throw new Error(`Refusing to write outside the locale folder: "${filePath}"`);
+  }
+  return resolvedFilePath;
+}
+
 async function processTargetFile(targetFilePath: string, keysToFix: MissingKeyType[]): Promise<void> {
   const { originFile, localeFolder } = keysToFix[0];
   const originFilePath = path.join(localeFolder, originFile);
+  const resolvedTargetFilePath = resolveLocaleJsonPath(localeFolder, targetFilePath);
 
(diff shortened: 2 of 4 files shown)


scripts/fix-missing-locales.ts
```diff
@@ -21,13 +21,27 @@ function delay(ms: number): Promise<void> {
...
+// out, pin the resolved destination to stay inside the resolved locale folder.
+function resolveLocaleJsonPath(localeFolder: string, filePath: string): string {
+  const resolvedFolder = path.resolve(localeFolder);
+  const resolvedFilePath = path.resolve(filePath);
+  // eslint-disable-next-line sonarjs/null-dereference -- false positive: path.resolve always returns a string
+  if (!resolvedFilePath.startsWith(resolvedFolder + path.sep) || path.extname(resolvedFilePath) !== ".json") {
+    throw new Error(`Refusing to write outside the locale folder: "${filePath}"`);
+  }
+  return resolvedFilePath;
+}
```
ivan-kalachikov, line 35: **This doesn't close #147.**

#147 is `js/http-to-file-access` ("Network data written to file") at `scripts/generate-backend-packages-json.js:462`. The tainted value is the content, not the path:

```
HTTP response -> fetchJson -> handleRemoteBackend -> packagesJson
  -> JSON.stringify(packagesJson)        <- tainted
  -> writeFileSync(join(__dirname, "..", "backend-packages.json"), ...)
                   ^ constant path, never in the flow
```

Path validation can't clear that alert. Fetching `packages.json` and writing it to disk is what the script does by design, so #147 is better dismissed with that reason than fixed.

Separately, this guard can never fire: `localeFolder` comes from `glob.sync(..., { absolute: true })`, `targetFile` from `readdirSync(localeFolder)`, and the caller passes `path.join(localeFolder, targetFile)`. The path is always inside the folder, so the `throw` is unreachable.

scripts/generate-backend-packages-json.js
```diff
@@ -165,6 +174,7 @@ async function httpRequest(url, options = {}, retries = 5, retryInterval = 5000)
   // Only disable certificate verification for localhost/local development environments
   // This prevents MITM attacks on public endpoints like GitHub
   if (isHttps && isLocalHost(urlObj.hostname)) {
+    // codeql[js/disabling-certificate-validation]: gated to a local/loopback hostname (see isLocalHost above)
```
ivan-kalachikov, line 177: **This comment probably won't close #145/#146.**

Syntax and placement are correct, but CodeQL doesn't drop a suppressed result — it keeps it and tags `suppressions[].kind = IN_SOURCE`. GitHub's SARIF ingestion only reads `ruleId`, `rule`, `level`, `message.text`, `locations`, `partialFingerprints`, `codeFlows`, `relatedLocations` — [`suppressions` is not on that list](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning). Hence the separate [`advanced-security/dismiss-alerts`](https://github.com/advanced-security/dismiss-alerts) action, which this repo doesn't run.

Dismissing both in the Security tab is the reliable path. The comments are still worth keeping as in-code rationale.

The PR CodeQL check doesn't cover this either way: [incremental analysis](https://github.blog/changelog/2025-05-28-incremental-security-analysis-makes-codeql-up-to-20-faster-in-pull-requests/) only reports alerts on changed lines, and 168 / 462 / `generator.ts:40` aren't in the diff. `workflow_dispatch` on `codeql.yml` runs a full scan.

Nit: "gated to a local/loopback hostname" promises more than `isLocalHost` delivers — it uses unanchored `startsWith("10.")`, `startsWith("192.168.")` and so on, so `10.foo.example.com` passes too. "gated to localhost/private-range hostnames" would be accurate. The function itself is fine as is.

scripts/graphql-codegen/generator.ts
```diff
@@ -35,6 +35,7 @@ if (
   backendUrl.startsWith("https://localhost") &&
   process.env.LOCAL_DEV_ALLOW_INSECURE_TLS === "true"
 ) {
+  // codeql[js/disabling-certificate-validation]: gated above to dev + https://localhost + an explicit opt-in env var, never reachable in CI/prod
```
ivan-kalachikov, line 38: Same as `generate-backend-packages-json.js:177` — syntax is right, but the comment alone won't close #145. Needs a dismissal in the Security tab.

.semgrep/vc-security.yml
```diff
@@ -198,12 +198,44 @@ rules:
...
+            ...
+            if (<... $E.origin ...>) { ... }
+            ...
+          })
+      - pattern-not-inside: |
+          $X.addEventListener("message", async function ($E) {
+            ...
+            if (<... $E.origin ...>) { ... }
+            ...
+          })
```
ivan-kalachikov, line 238: The rule behaves correctly now.

- **old version:** fires on every handler shape regardless of guards, including the three real guards behind #217/#221/#222
- **new version:** clears those three, still fires on unguarded arrow / async-arrow / function-expression handlers
- braced, unbraced and helper-based guards (`if (!isTrusted(event.origin))`) all clear

One caveat for later: `pattern-not-inside` only needs `$E.origin` to appear in the condition, so `if (event.origin) { /* no return */ }` clears too. Reasonable trade-off for killing the false positives.

ivan-kalachikov (commented):
Semgrep rule is good. On the CodeQL side 5 of 8 alerts close; #145, #146 and #147 look like they'll stay open. Details inline.

Those three need a manual dismissal in the Security tab to satisfy the ticket's "closed or dismissed with a reason".

scripts/fix-missing-locales.ts
ivan-kalachikov: Worth being precise about the dismissal reason. `backend-packages.json` isn't a throwaway file — it's committed to the repo, imported by `client-app/core/utilities/checkModulesVersions.ts`, and CI passes it as `customPackagesJsonUrl` in `theme-ci.yml:263`, so its contents drive what the e2e rig installs (`PlatformImage`, `ServiceUri`, module versions).

What makes it safe is the review gate: the script is run by hand and the resulting diff is committed and reviewed before anything consumes it. That's the reason worth recording, rather than "build-only tooling".

Reason category: **Won't fix**, not False positive — network data really is written to a file, so the query isn't wrong about the facts.

scripts/generate-backend-packages-json.js
ivan-kalachikov: On the dismissal reason for #145/#146: **Won't fix** rather than False positive — certificate validation really is being disabled, so "false positive" would be inaccurate in the audit trail. Both are severity `high`, which makes the recorded reason worth getting right.

#145 is clean to dismiss: three gates plus an explicit opt-in env var. #146 is the weaker one — with `isLocalHost` matching by unanchored prefix, a dismissal recorded as "gated to localhost" claims more than the code delivers. Anchoring those checks (or parsing the address properly) first would make the dismissal accurate. Separate change — that line isn't in this diff.

muller39:
3 CodeQL alerts still need a manual dismissal in the Security tab (Won't fix) — code is done, in-code comments alone don't close alerts:

1. [#145](https://github.com/VirtoCommerce/vc-frontend/security/code-scanning/145) — TLS is only disabled in dev mode + localhost + an explicit opt-in flag
2. [#146](https://github.com/VirtoCommerce/vc-frontend/security/code-scanning/146) — same, now honestly scoped to localhost/private IPs via anchored parsing
3. [#147](https://github.com/VirtoCommerce/vc-frontend/security/code-scanning/147) — file is committed and human-reviewed before merge, so the write is safe

Once these are dismissed, all 11 alerts from the ticket are closed or dismissed with a reason.

ivan-kalachikov (approved):

ivan-kalachikov (approved):

PR #2463, opened 2026-08-31
feat(VCST-5652): button icon sizing per design spec

goldenmaya:
### Jira-link:

https://virtocommerce.atlassian.net/browse/VCST-5652
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.58.0-pr-2463-90db-90dbedc7.zip

---

> [!NOTE]
> **Medium Risk**
> Shared **`VcIcon`** sizing behavior changed globally (CSS variable model), and **`VcButton`** default icon dimensions differ from the previous line-height-based defaults, which may shift UI across the app.
> 
> **Overview**
> **VcIcon** now drives dimensions through a single **`--size`** custom property (presets set it in CSS; numeric/arbitrary sizes set it inline) instead of separate inline `width`/`height` or Tailwind `size-*` utilities on presets. Tests were updated to match.
> 
> **VcButton** defines per-size **`--icon-size`** values and uses them as the default for **`--vc-icon-size`**, so prepend/append/icon-only glyphs follow the design spec while **`iconSize`** still overrides. Loading spinners scale with **`--vc-icon-size`**. **`iconVariant`** is forwarded to all built-in **`VcIcon`** instances. Layout tweaks keep labels vertically centered when icons differ from the text line (flex on prepend/append, margin adjustment for icons in the label slot).
> 
> Storybook adds **`iconVariant`** controls plus **IconVariant** and **PrependIconSmall** examples.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 90dbedc7156e1e15425b610a280beb463ac2f7f5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


yuskithedeveloper (dismissed):
Findings (please, comment or fix before merging):

---

### 1. Stale guard in the preset test

[`vc-icon.test.ts:32`](https://github.com/VirtoCommerce/vc-frontend/blob/9dd08556682cac499965a8136d561877dd52793f/client-app/ui-kit/components/atoms/icon/vc-icon.test.ts#L32)

```js
expect(wrapper.attributes("style") || "").not.toContain("width");
```

This line existed to prove *presets emit no inline dimension*. Dimensions now travel as `--size`, so the assertion is vacuous — a regression that inlined `--size: sm` for a preset would sail straight through it. The other three assertions in the file were updated correctly; this one was missed:

```diff
-    expect(wrapper.attributes("style") || "").not.toContain("width");
+    expect(wrapper.attributes("style") || "").not.toContain("--size");
```

### 2. The loading spinner no longer matches the icon it replaces

[`vc-button.vue:287-289`](https://github.com/VirtoCommerce/vc-frontend/blob/9dd08556682cac499965a8136d561877dd52793f/client-app/ui-kit/components/molecules/button/vc-button.vue#L287)

`__loader-icon` is still `size-[--line-height]`. That used to equal the icon size by construction; now on an `md` icon button the glyph is `1.5rem` and the spinner that replaces it is `1.25rem` (`lg`: 1.75 vs 1.5). Does the spec cover the loader? If it should track the icon, it wants `size-[--icon-size]`.

### 3. `inline-flex` is inert (nit)

[`vc-button.vue:460`](https://github.com/VirtoCommerce/vc-frontend/blob/9dd08556682cac499965a8136d561877dd52793f/client-app/ui-kit/components/molecules/button/vc-button.vue#L460) and [`:472`](https://github.com/VirtoCommerce/vc-frontend/blob/9dd08556682cac499965a8136d561877dd52793f/client-app/ui-kit/components/molecules/button/vc-button.vue#L472)

`__prepend`/`__append` are grid items of `__content` (`grid grid-flow-col`), so `display: inline-flex` is blockified to `flex` — writing `flex` says what actually happens. Also worth noting the fix is a no-op for the *default* icons: with the new tokens every prepend icon is >= `--line-height`, so it only bites for a custom `icon-size` or slot content. Fine, but it means none of the Storybook stories demonstrate it.

goldenmaya:
@yuskithedeveloper Good catches — all three addressed.

1. Guard is now `not.toContain("--size")`. Verified it fails on an inlined `--size` for a preset.
2. Went with `size-[--vc-icon-size]` rather than `--icon-size`: that's the button's fully resolved chain, so the spinner also follows a custom `iconSize` / `--vc-button-icon-size` (with `--icon-size`, a `2.5rem` icon still got a 24px spinner). Icon-only `md` while loading is now 44×44 with a 24px spinner.
3. Changed to `flex`, and added a `PrependIconSmall` story (`iconSize: "0.75rem"`) so the centring is actually demonstrated.

vas11yev1work (approved):

PR #2464, opened 2026-09-01
feat(VCST-5732): sales rep task management

vkrashenko:
## Description

Adds **task management** to the Sales Rep hub: a rep can see, create, edit, complete and delete their own tasks, from a new Calendar page and a dashboard widget. Server side is [vc-module-sales-rep#16](https://github.com/VirtoCommerce/vc-module-sales-rep/pull/16).

- **Calendar page** (`/company/calendar`) — month grid with per-day markers, the selected day's list, and `Upcoming` / `Overdue` / `Completed` tabs with counts.
- **Dashboard widget** — the same mini calendar and day list, plus a standing overdue total, linking through to the page.
- **Create / edit modal** — title, due date, priority, type and notes, with delete behind a confirmation.
- 5 queries and 4 mutations against the sales-rep X-API. The tab and sort vocabularies are **fetched, not hard-coded**, so a project that re-registers the server-side resolver gets the new tabs here without a frontend change.

**Shared ui-kit change** — `VcCalendar` gained a `day` slot, per-day screen-reader descriptions (`dayDescriptions`) and an optional controlled `month` with an `update:month` event. Reka's cell slot is only taken over when one of the first two is in use, so an undecorated calendar renders exactly as before and existing usages are unaffected.

**Worth knowing**
- The whole feature is gated on the backend `VirtoCommerce.TaskManagement` module being installed — with it absent the nav item, the route and the widget stay hidden rather than showing an empty page.
- Status is **derived, never stored**: from `isActive`, `completed` and `dueDate` against the viewer's local day boundary, so the tab a task falls into and the pill it renders always agree.
- Translations added for all 13 locales.

### Screenshots
image
image
image

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5732
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.58.0-pr-2464-fd52-fd52185e.zip


yuskithedeveloper:
## Review

Nice work on the hard parts — the local-calendar-vs-UTC discipline in `tasks.ts` (one `dayStart` sent to the server as `today` *and* used for client status derivation, so a tab and the pill beside it can't disagree), the day-vs-tab decision where a status tab *replaces* the day window instead of intersecting it (with the empty-state copy following it), the `onTaskSaved` refetch arithmetic that skips surfaces Apollo will restart anyway, and `VcCalendar`'s `dayDescriptions` correctly working around reka's explicit `aria-label` excluding slot content from the accessible name. The header `grid-template-columns` change keeps the nav buttons on the same visual columns, so no existing `VcDatePicker` consumer regresses.

### 🔴 Blocking

**1. The regenerated `api/graphql/types.ts` does not compile — CI's Build step will fail.**

```
$ yarn validate:types   →  exit 2,  25× error TS2304
client-app/modules/sales-rep/api/graphql/types.ts(1167,14): Cannot find name 'FacetTypes'.
client-app/modules/sales-rep/api/graphql/types.ts(540,10):  Cannot find name 'CustomerOrderType'.
… 23 more, all in this one file
```

The regeneration deleted 17 type/enum definitions that the same file still references. Verified on both sides — every one is defined on `dev`, none on HEAD: `InventoryInfo`, `ImageType`, `OutlineType`, `SeoInfo`, `Facet`, `FacetTypes`, `FacetRangeType`, `FilterFacet`, `TermFacet`, `CustomerOrderType`, `OrderAddressType`, `OrderPaymentMethodType`, `DynamicPropertyValueType`, `TierPriceType`, `DescriptionType`, `VariationType`, `VideoConnection`. `dev` has zero type errors, so all 25 come from this branch.

This isn't only a red local check: `theme-ci.yml` runs `yarn build`, and `build = yarn validate && yarn build-only` with `validate = validate:types && …`. The green Sonar gate doesn't cover it — Sonar is an earlier, separate step.

Runtime is unaffected: no app code imports the dropped names, and no operation document was lost. It's purely a broken generated artifact.

Related symptom in the same file, suggesting it's part-stale rather than just part-pruned: `salesRepTaskFields.graphql` selects 8 fields and `SalesRepTaskFieldsFragment` matches — but `CreateSalesRepTaskMutation`, `UpdateSalesRepTaskMutation`, `ChangeSalesRepTaskStatusMutation` and `SalesRepTasksQuery` all declare `createdDate: any, modifiedDate?: any`, which no document requests. Typed as present, `undefined` at runtime. Nothing reads them today.

**Suggested fix:** re-run `yarn generate:graphql-types` against a backend with the full X-module set loaded (a schema missing XCatalog/XOrder types is the likely cause of the pruning), then confirm `yarn validate:types` exits 0.

**2. `calendarRoute.beforeEnter` calls `next()` twice** — `routes.ts:117-121`

```ts
if (guardSalesRep(next) && isSalesRepTasksEnabled()) {
  next();
} else {
  next({ name: DASHBOARD_ROUTE_NAME });
}
```

`guardSalesRep` already calls `next({ name: "Dashboard" })` itself before returning `false` (`routes.ts:38-45`). For a non-rep hitting `/company/calendar` the `else` fires a second `next()` — Vue Router warns *"The `next` callback was called more than once in one navigation guard"* and ignores it. The branches also aim at different routes (storefront `"Dashboard"` vs `DASHBOARD_ROUTE_NAME` = `"SalesRepDashboard"`), so only the first wins today. Every sibling in the file early-returns instead — `documentsRoute`, `guardRepRoute`, `guardCustomerRoute`:

```ts
beforeEnter(_to, _from, next) {
  if (!guardSalesRep(next)) {
    return;
  }
  if (isSalesRepTasksEnabled()) {
    next();
  } else {
    next({ name: DASHBOARD_ROUTE_NAME });
  }
}
```

### 🟡 Small and local

**3. `useSalesRepTaskCalendar.ts:33` hardcodes `sort: "due-date"`** instead of `TASKS_SORT_RULE`. `constants.ts:88-91` exists exactly to prevent this ("The one order every task surface asks for… keep it aligned with a real rule"). Rename the rule server-side and two of three surfaces follow; the dots query silently doesn't.

**4. Dead wiring: `task:sort`.** `salesRepTaskSortRulesQuery.graphql` and the `"task:sort"` entry in `useSalesRepRules.ts:71` are never consumed — only `useSalesRepRules("task", "filter")` is called, and `constants.ts` says so ("no surface offers a sort control, so the rule list itself is not fetched"). Drop both, or note why they're staged ahead of a consumer.

**5. A cancelled task's checkbox silently converts it to completed.** `taskStatus` → `"canceled"` for `isActive: false, completed: false|null`. In `sales-rep-task-list.vue` the checkbox is `:model-value="item.status === 'completed'"` → unchecked and enabled; clicking runs `setCompleted(task.id, true)`. Reachable: the "All" tab passes no filter, so the day's cancelled rows are listed with a neutral "Canceled" chip. Disabling the checkbox for `canceled` would match the chip's read-only intent.

### 🔵 Judgment calls

**6. Calendar dots truncate silently, biased toward the end of the month.** `TASKS_CALENDAR_MAX = 200` with `sort: due-date` ascending over a window starting 7 days *before* the month — past the cap it's the later grid days that lose their dots, with nothing saying so. The comment frames the cap as protection against a pathological book of business, which is fair, but a rep who crosses it sees a *wrong* calendar rather than a truncated one. Reading `totalCount` off the same response and surfacing a marker when it exceeds the cap would close it cheaply.

**7. The widget's picked day doesn't travel to the page.** `sales-rep-tasks.vue` owns its own `selectedDay`, but both exits — "Full calendar" and the overdue notice — are bare `{ name: CALENDAR_ROUTE_NAME }`, so picking Oct 20 in the widget and clicking through lands on today. A day in the route would make the two surfaces continuous.

**8. AC gap worth recording on the ticket.** VCST-5732 lists five groupings (*My Today's tasks, Active, Upcoming, Overdue, Completed*); shipped is **All (selected day) / Upcoming / Overdue / Completed**. `SalesRepTaskFilterRuleResolver` only offers three rules, so this is upstream of the frontend — and server-driven tabs are the right call — but "Active" has no equivalent anywhere in the UI and shouldn't be discovered in E2E.

OlegoO (approved):

PR #2467, opened 2026-09-03
feat(VCST-5378): support authenticated UCP handoffs and storefront OAuth

ToxaKZ:
## Description

Restores anonymous and authenticated UCP checkout handoffs through the existing checkout route and supports storefront OAuth login when Platform is private.

Handoff restoration runs after checkout access checks. The single-use token stays in tab-local sessionStorage; only an opaque reference travels through sign-in. Authentication failures preserve that continuation, and wrong-account or expired-link failures show a local message. Signed-in buyers receive anonymous handoff carts through the existing mergeCart mutation without replacing their identity. Concurrent restores share one request, successful restores are cached, and failed requests remain retryable.

The `/oauth/authorize` page preserves the OAuth request, obtains Platform's antiforgery token, establishes the browser session through `/connect/session`, and continues to consent. It handles failures locally without broadcasting them to other tabs. OpenIddict owns PKCE, consent and token issuance. The development proxy uses standard forwarded headers; deployment PR #6459 supplies production routes and the opt-in Platform configuration.

### Validation

- 2415 frontend tests passed across 161 files, including real-route login continuation, wrong-account recovery, identity preservation and OAuth error handling.
- Production build and TypeScript validation passed. Changed-file lint: 0 errors, 10 warnings. Dependency validation: 0 errors, 19 warnings.
- Live browser checks passed for guest and signed-in handoffs, closed-store login, 403 recovery with the correct account, used-link errors, and OAuth consent continuation.
- 31 live OAuth API checks passed, including code exchange, refresh, logout, CSRF, resource isolation and replay rejection.
- External identity providers were not exercised.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5378

### Related PRs:
https://github.com/VirtoCommerce/vc-platform/pull/3108
https://github.com/VirtoCommerce/vc-module-ucp/pull/7
https://github.com/VirtoCommerce/vc-deploy-dev/pull/6459

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.59.0-pr-2467-1951-195127f5.zip


client-app/shared/checkout/ucp/handoff.ts
```diff
@@ -1,64 +1,118 @@
...
+
+export function applyUcpHandoffBuyer(anonymousBuyerId?: string): void {
+  if (!anonymousBuyerId) {
     return;
   }
 
-  localStorage.setItem(USER_ID_LOCAL_STORAGE, buyerId);
-  setGlobals({ userId: buyerId });
+  localStorage.setItem(USER_ID_LOCAL_STORAGE, anonymousBuyerId);
+  setGlobals({ userId: anonymousBuyerId });
```
ivan-kalachikov, line 29: **An anonymous buyer id overwrites a signed-in user's identity.**

The anonymous probe sends no bearer, so the backend returns `anonymous_buyer_id` even when the browser holds a Platform token (`AnonymousBuyerId = buyerContext.IsAuthenticated ? null : PublicBuyerId`). This writes it unconditionally.

Order on one page load of `/checkout?ucp_session=X` while signed in:

1. `app-runner.ts:64` writes the anon id
2. `setUser` and `app-runner.ts:221` heal it back to the real user
3. `checkout.ts:79` writes the anon id **again** on the cache hit, and nothing heals it after that

`globals.userId` feeds `getFullCart`, `getShortCart`, `getSavedForLater`, `getOrders`, quotes, purchase-requests, whiteLabeling and `validateCoupon`. x-cart then denies the cart: `CanAccessCartAuthorizationRequirement` checks `cart.CustomerId == GetUserId(context)` for an authenticated principal.

Suggestion: return early when `useAuth().headers.value.Authorization` is set. `isAuthenticated` is not populated yet at `app-runner.ts:64`, so it can't be used here.

client-app/router/routes/checkout.ts
```diff
@@ -75,11 +75,15 @@ export const checkoutRoutes: RouteRecordRaw[] = [
...
+          const { cartId, anonymousBuyerId } = await restoreUcpHandoffCart(to.query.ucp_session);
+          applyUcpHandoffBuyer(anonymousBuyerId);
           next({ name: ROUTES.CART_ID.NAME, params: { cartId }, query: { ucp_handoff: "1" }, replace: true });
         } catch (error) {
           console.error("Unable to restore UCP handoff session.", error);
+          if (error instanceof UcpHandoffRestoreError && error.status === 401) {
+            next({ name: ROUTES.SIGN_IN.NAME, query: { returnUrl: to.fullPath }, replace: true });
+            return;
+          }
           next({ name: ROUTES.CART.NAME, replace: true });
```
ivan-kalachikov: **403 and 400 land on `/cart` with no message.**

Backend statuses for restore:

| Status | Cause | Token consumed |
|---|---|---|
| 401 | authentication required | no |
| 403 | signed in as the wrong user | no |
| 400 | token already used or expired | n/a |
| 5xx | infra | no |

Only 401 is handled. 403 is the one case the user can fix themselves by signing in with the right account, and they get a silent bounce to their own cart instead. 400 happens whenever the handoff link is reopened or reloaded.

Suggestion: give 403 its own message, and surface a notification for the rest rather than only `console.error`.

ivan-kalachikov (commented):
Reviewed at `97b4dac`. Verified locally: `vue-tsc --build --force` clean, `eslint` on the changed files 0 errors (9 warnings, all pre-existing per blame), `vitest run --dir client-app` green.

No blocker in the code. Comments are ordered roughly by impact: identity handling on the anonymous-handoff path, then silent failure paths, then test gaps found by mutation testing.

Two things worth recording as clean, since they are the obvious suspicions:

- the continuation check in `authorize.vue` holds against ~18 hostile `returnUrl` inputs (backslashes, `%2F` escapes, userinfo tricks, protocol-relative, tabs/newlines, `javascript:`, case variants);
- OAuth `state` / PKCE round-trips byte-for-byte through the storefront login hop, including the Platform's own `QueryString.Create` recompute in `Authorize()`. Tested against the real `vue-router` in this repo.

Small note: the description says 23 targeted tests, the diff contains 19.

client-app/router/routes/checkout.ts
```diff
@@ -75,11 +75,15 @@ export const checkoutRoutes: RouteRecordRaw[] = [
...
         try {
-          const { cartId, buyerId } = await restoreUcpHandoffCart(to.query.ucp_session);
-          applyUcpHandoffBuyer(buyerId);
+          const { cartId, anonymousBuyerId } = await restoreUcpHandoffCart(to.query.ucp_session);
+          applyUcpHandoffBuyer(anonymousBuyerId);
           next({ name: ROUTES.CART_ID.NAME, params: { cartId }, query: { ucp_handoff: "1" }, replace: true });
         } catch (error) {
           console.error("Unable to restore UCP handoff session.", error);
+          if (error instanceof UcpHandoffRestoreError && error.status === 401) {
+            next({ name: ROUTES.SIGN_IN.NAME, query: { returnUrl: to.fullPath }, replace: true });
```
ivan-kalachikov: **This redirect is undone when the user is already signed in.**

`router/index.ts:52-64` runs for the new navigation, sees `isAuthenticated && to.name === SignIn`, and calls `next(getReturnUrlValue() || { name: Catalog })`. `getReturnUrlValue()` reads `location.href`, which is still `/checkout?ucp_session=X` and carries no `returnUrl`, so the user lands on `/catalog` and the handoff URL is gone.

Reproduced against the real route table: the hops were `/checkout?ucp_session=s4` then `Catalog`.

Reachable when the storefront still holds a user but Platform rejects its bearer, e.g. the refresh token expires between boot and this guard. No test covers it.

client-app/pages/auth/authorize.vue
```diff
@@ -0,0 +1,40 @@
...
+import { useFetch } from "@/core/api/common";
+import { toSameOriginPath } from "@/core/utilities";
+
+const route = useRoute();
+const router = useRouter();
+
+onMounted(async () => {
+  const returnUrl = typeof route.query.returnUrl === "string" ? toSameOriginPath(route.query.returnUrl) : "";
+  if (!/^\/connect\/authorize\?[^#]*$/.test(returnUrl)) {
+    await router.replace("/400");
```
ivan-kalachikov: **`/400` tells a valid user they are banned.**

`locales/en.json:1093` reads *"The user is no longer allowed to sign in. Please contact the administrator."* That copy was written for the only other consumer, `callback.vue` on `SIGN_IN_NOT_ALLOWED`. It is wrong for all three exits here:

| Cause | Platform response |
|---|---|
| `OAuthLoginPath` unset, which is the default | 404 |
| antiforgery cookie lost or `Origin` mismatch | 400 |
| operator in impersonation mode | 401 |

`CreateSession` has eight reasons to return 401 and only one of them (`!CanSignInAsync`) matches that message.

Two suggestions:

- a dedicated i18n string for a failed continuation;
- send **401** to `/sign-in?returnUrl=route.fullPath`, the same thing `router/index.ts:35-38` does. `statusCode` is on the vueuse shell, so it is available. This has to be handled here: the global 401 listener can never rescue this route, because `buildRedirectUrl` returns `null` for any route whose query already holds a `returnUrl`, and this one always does.

client-app/pages/auth/authorize.vue
```diff
@@ -0,0 +1,40 @@
...
+    return;
+  }
+
+  const { data, error: tokenError } = await useFetch("/connect/session", { headers: {} })
```
ivan-kalachikov: **These failures broadcast to every open tab.**

`useFetch` from `@/core/api/common` installs a global `onFetchError` that calls `errorHandler(toServerError(...))`, and that maps everything except 401/403 to `Unhandled`:

- 404 / 400 / 5xx → `unhandledErrorEvent` with `TabsType.ALL` → error toast in every open storefront tab
- 401 → `unauthorizedErrorEvent` with `TabsType.ALL` → other tabs run `fetchUser()` and get pushed to `/sign-in`
- 403 → `forbiddenEvent` → `router.push(NoAccess)` races this page's own `router.replace("/400")`

The page already handles its own errors, so the broadcast is pure side effect. `ucp-handoff.ts` uses bare `fetch` and avoids this.

Suggestion: pass `onFetchError: (context) => context` in the options argument. `combination: "overwrite"` makes it replace the global handler, same pattern as `useAuth.ts:22-27` for `/connect/token`.

client-app/app-runner.ts
```diff
@@ -60,9 +60,9 @@ async function getUcpHandoffUserId(): Promise<string | undefined> {
   }
 
   try {
-    const { buyerId } = await restoreUcpHandoffCart(ucpSession);
-    applyUcpHandoffBuyer(buyerId);
-    return buyerId;
+    const { anonymousBuyerId } = await restoreUcpHandoffCart(ucpSession);
```
ivan-kalachikov: **Boot consumes the single-use token before the app knows anything.**

This runs from `app-runner.ts:148`, before store settings arrive (`:157`) and before the router exists (`:214`). The backend now deletes the session on success (`RemoveAsync` in `RestoreHandoffCore`), and the result lives only in a module-level `Map` that every full document navigation wipes.

Two reachable outcomes:

- reopening the handoff link or pressing F5 → 400 → silent `/cart`;
- closed store: boot consumes the token, the guard redirects to sign-in, login is a full reload → 400 → silent `/cart`. Items survive through `useSignMeIn`'s `mergeCart`, the checkout context does not.

`line 56` also reads `ucp_session` from any path, so `/?ucp_session=X` burns the token as well.

Suggestion: restore only on the checkout route, or persist the outcome across the reload.

client-app/shared/checkout/ucp/handoff.ts
```diff
@@ -1,64 +1,118 @@
...
+async function requestUcpHandoffRestoreWithFallback(ucpSession: string): Promise<UcpHandoffRestoreType> {
+  try {
+    return await requestUcpHandoffRestore(ucpSession, false);
+  } catch (error) {
+    if (!(error instanceof UcpHandoffRestoreError) || error.status !== 401) {
+      throw error;
+    }
+  }
+
+  return requestUcpHandoffRestore(ucpSession, true);
```
ivan-kalachikov, line 75: **The authenticated retry runs even when there is no token.**

`useAuth().headers` is `{}` without an access token, so this second request differs from the probe only by `credentials: "include"` and can only 401 again.

Rejections are not memoised either, so the route guard repeats the whole pair after boot already failed. An authenticated handoff opened while logged out costs **4** POSTs before the sign-in redirect.

Suggestion: skip the retry when `authHeaders.value.Authorization` is absent, and remember the rejection for the life of the document. Worth also not logging `status === 401` as an error, here and at `checkout.ts:82` — it is normal control flow.

client-app/router/index.ts
```diff
@@ -32,7 +32,10 @@ export function createRouter(options: { base: string }) {
 
     if (unauthorizedAccessIsDenied) {
       // save current location to return to it after sign in
-      const query = buildRedirectUrl(to) || {};
+      const isUcpHandoff =
+        to.matched.some((route) => route.name === "Checkout") && typeof to.query.ucp_session === "string";
+      const query =
+        isUcpHandoff || to.name === "OAuthAuthorize" ? { returnUrl: to.fullPath } : buildRedirectUrl(to) || {};
```
ivan-kalachikov, line 50: **The single-use token now travels into sign-in URLs and telemetry.**

Bypassing `buildRedirectUrl` also bypasses `meta.redirectable: false`, which used to drop `?ucp_session=` at this guard. It now lands in the `/sign-in` URL, browser history, App Insights page-view telemetry, and the external-IdP round trip (`callbackUrl=…?returnUrl=…` through `useIdentityProvider`).

The token is single-use and server-expiring, but for an anonymous handoff it alone grants access to the cart's PII.

Not blocking. Worth a conscious decision, or move the token into `sessionStorage` before redirecting.

vite.config.ts
```diff
@@ -25,6 +26,12 @@ function getProxy(target: ProxyOptions["target"], options: Omit<ProxyOptions, "t
     target,
     changeOrigin: true,
     secure: dontTrustSelfSignedCertificate,
+    configure(proxy) {
```
ivan-kalachikov: **`xfwd: true` already does this.**

Vite's `ProxyOptions` exposes `xfwd` (`vite/dist/node/index.d.ts:355`), and the bundled http-proxy sets `x-forwarded-for`, `-port`, `-proto` and `-host` — using `conn instanceof TLSSocket` instead of a cast, and appending to existing values rather than overwriting them.

Two gaps in the hand-rolled version:

- no `proxyReqWs`, so HTTP `/graphql` gets the headers but the live WS upgrade does not;
- `...options` sits **after** `configure`, so any future per-route `configure` would silently drop the forwarding.

Switching to `xfwd: true` also removes the `node:tls` import, which exists only for that cast.

client-app/router/routes/ucp-handoff.ts
```diff
@@ -1,64 +1,118 @@
...
+      `Unable to restore UCP handoff session. Status: ${response.status}. ${details}`,
+      response.status,
+    );
   }
 
   const payload = (await response.json()) as {
+    anonymous_buyer_id?: string;
     checkout?: {
       cart_id?: string;
       buyer?: { id?: string };
```
ivan-kalachikov: **These two are no longer read.**

After this PR only `cart_id`, `cart.id` and `anonymous_buyer_id` are used. Dropping `buyer` and `cart.buyer_id` from the type makes the "identity comes only from `anonymous_buyer_id`" rule visible in the shape itself.

Minor: with `VirtoCommerce:IncludeOutputNullValues` enabled the backend sends `anonymous_buyer_id: null` instead of omitting it, so `?: string` is slightly off (`?: string | null`). The runtime `!anonymousBuyerId` check handles both.

client-app/shared/checkout/ucp/handoff.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    vi.stubGlobal("fetch", fetchMock);
+
+    await restoreUcpHandoffCart("expired-token-session");
+
+    expect(auth.refresh).toHaveBeenCalledOnce();
+    const request = fetchMock.mock.calls[1][1] as RequestInit;
+    expect((request.headers as Headers).get("Authorization")).toBe("Bearer refreshed-token");
+  });
+
+  test("deduplicates concurrent restore calls for a single-use handoff session", async () => {
```
ivan-kalachikov, line 154: **Mutation testing found five survivors in this file.**

| Mutation | Result |
|---|---|
| remove the success-cache read | all 19 tests stay green |
| retry authenticated on **any** status | green |
| hard-code the error's `status` to `401` | green |
| remove the pending-map `finally` cleanup | green |

The cache one matters most. Boot and the route guard both restore the same single-use session, and only the cache stops the second call from re-POSTing a consumed token — that is the normal happy path, and nothing covers it. The dedup test here covers the *concurrent* case only, not sequential reuse.

The `status === 401` narrowing is unpinned on both sides because no fixture ever returns a non-401 failure.

`line 51` too: `toEqual({ cartId: "authenticated-cart" })` passes with `anonymousBuyerId: undefined` present, and the authenticated fixture carries no `checkout.buyer.id` — so the headline claim that authenticated ids never reach anonymous storage isn't actually pinned. Adding `buyer: { id: "contact-1" }` to that fixture and asserting `anonymousBuyerId` is undefined would close it.

client-app/router/ucp-handoff-navigation.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+vi.mock("@/core/composables", () => ({ useThemeContext: () => ({ themeContext: state.theme }) }));
+vi.mock("@/shared/account", () => ({
+  useUser: () => ({ isAuthenticated: state.authenticated, organization: { value: undefined } }),
+}));
+vi.mock("@/core/utilities", async () => {
+  const { buildRedirectUrl } = await import("@/core/utilities/common");
+  return { buildRedirectUrl, getReturnUrlValue: () => undefined };
+});
+vi.mock("./routes", () => ({
```
ivan-kalachikov: **This table pins the route names against themselves.**

The real records never load, so both of these stay green:

- deleting `meta: { requiresAuth: true }` from the real `OAuthAuthorize` in `main.ts`
- deleting `meta: { redirectable: false }` from the real `Checkout`

Renaming the real route also stays green while production breaks: `router/index.ts:38` would fall through to `buildRedirectUrl`, which returns `null` because the query already holds a `returnUrl`, so the continuation is lost after login.

The sibling `client-app/router/index.test.ts` does it the other way — real `mainRoutes`, `importOriginal` for the two composables — and runs in 77 ms. Worth following, plus a `describe` block: these two new files are the only 2 of 161 without one.

client-app/shared/checkout/ucp/handoff.ts
ToxaKZ: Fixed

client-app/router/routes/checkout.ts
ToxaKZ: Fixed

client-app/router/routes/checkout.ts
ToxaKZ: Fixed

client-app/app-runner.ts
ToxaKZ: Fixed

client-app/shared/checkout/ucp/handoff.ts
ToxaKZ: Fixed

client-app/router/routes/ucp-handoff.ts
ToxaKZ: Fixed

client-app/shared/checkout/ucp/handoff.test.ts
ToxaKZ: Fixed

client-app/router/ucp-handoff-navigation.test.ts
ToxaKZ: Fixed

client-app/pages/auth/authorize.vue
ToxaKZ: Fixed

client-app/pages/auth/authorize.vue
ToxaKZ: Fixed

client-app/router/index.ts
ToxaKZ: Fixed

vite.config.ts
ToxaKZ: Fixed

client-app/shared/checkout/ucp/continuation.ts
```diff
@@ -0,0 +1,16 @@
+const STORAGE_PREFIX = "ucp-handoff:";
```
ivan-kalachikov, line 1: **`router/routes/` should hold routes; these two don't.**

Every other file in that folder exports a route table or route constants:

| File | Exports |
|---|---|
| `account.ts`, `cart.ts`, `checkout.ts`, `company.ts`, `main.ts` | `*Routes` (`RouteRecordRaw[]`) |
| `constants.ts` | `ROUTES` |
| `index.ts` | barrel |
| **`ucp-continuation.ts`** | three `sessionStorage` helpers |
| **`ucp-handoff.ts`** | REST client, error class, identity writer |

The two UCP files export no route records at all.

`ucp-handoff.ts` predates this PR (#2343), so that part is inherited. But this PR grew it from 13 to 122 lines and added a sibling beside it, which makes this the natural moment to move both.

Suggested home, matching how the repo keeps domain code under `shared/<domain>/`:

```
client-app/shared/checkout/ucp/
  handoff.ts       <- routes/ucp-handoff.ts
  continuation.ts  <- routes/ucp-continuation.ts
```

One caveat: import by exact subpath, not through the `@/shared/checkout` barrel. `shared/checkout/composables/useCheckout.ts` already pulls in `useRouter`, so a barrel import from `router/routes/checkout.ts` risks a Rollup chunk cycle.

Mechanical move, six import sites:

```
client-app/router/index.ts:7
client-app/router/routes/checkout.ts:7,8
client-app/router/routes/checkout.test.ts:4,5
client-app/router/routes/ucp-handoff.test.ts:4
client-app/router/ucp-handoff-navigation.test.ts:3
```

Non-blocking.

ToxaKZ: Fixed

ivan-kalachikov (dismissed):

vas11yev1work (approved):

PR #2476, opened 2026-09-09
feat(VCST-5707): rework list sharing

ivan-kalachikov:
## Description

A Sales Rep can share one list with several customers at once. The combined "List settings" dialog is split into **Rename** and **Share**, and the list-actions menu becomes **Rename / Share / Remove list**.

**Draft until VCST-5925 lands.** The generated types here came from `vcptcore-dev`. Re-run codegen and `generate:backend-packages` against a backend carrying the merged schema before this is merged.

### Dialogs

- **Rename** (`AddOrUpdateWishlistModal`) — name and description; creates and renames. A new list starts private.
- **Share** (`ShareWishlistModal`) — scope tabs, the sharing link, and the selected scope's own controls. Primary action reads **Share** for a scope with recipients, **Save** otherwise.
- List details gains a Share button.
- Each dialog sends only the fields it owns, so a rename cannot overwrite sharing and a share cannot overwrite the name.
- Sharing is offered to the list's owner only.

### Scopes

- `VcTabSwitch` tabs replace the select: Private, My organization, Anyone with link, Specific customers. The registry gained `icon` and `order`.
- The selected scope's controls are kept alive, so switching tabs preserves the draft. The cache dies with the dialog.
- The sales-rep Customer scope plugs in through the existing registry.

### Recipients

- Reads use `sharingSetting.targets`; `sharedWithId` is gone. The module owning the scope resolves each target's name, city and logo server-side.
- Writes are deltas — `addSharedWithIds` / `removeSharedWithIds` — never a replace-set.
- Multi-select picker over one page of 100 served customers, two-line options, client-side filter.
- Recipients list: per-row remove, `Clear all` with undo, and a `Show all N` / `Show less` toggle past three rows.
- Emptying the list does not stop sharing — Save stays disabled and the field says so. Changing the scope is the way out.

### Message

- Saved with the share (`sharingSetting.message`), offered whether or not anyone new is added, capped at 250 characters (the backend allows 1024).
- One `sendCustomerCommunication` per save reaches every newly added organization; a member of several receives it once.
- The success toast counts the audience the server persisted.

### Stopping a share

- Changing the scope of a shared list asks for confirmation first.
- It asks only where an audience is lost: not on a first share out of Private, not when recipients change inside one scope, and not when the scope being left has no recipients.
- The confirmation renders inside the share dialog's own tree rather than on the modal stack.

### Changed surface

- `WishlistDropdownMenu`: new `shareable` prop, new `share` emit, and `edit` / `remove` now carry the trigger element so a modal opened from the menu can return the focus to something still on screen.
- `WishlistCard`: same three emits, each forwarding that element.
- `SalesRepCommunicationInputType.organizationId` → `organizationIds`; one send now covers the whole newly added audience.
- Locale keys removed: `list_card.list_edit_button`, `add_or_update_wishlist_modal.sharing_scope_label`, `add_or_update_wishlist_modal.sharing_scope_placeholder`, `sales_rep.list_sharing.share_replace_hint`. `list_details.list_settings_button` and `add_or_update_wishlist_modal.edit_mode_title` became linked messages pointing at the new `rename_list_button`.

### Not in this PR

Select-all with its row counter, and picker paging — both belong to VCST-5923.

### Notes for review

- Design spec: `client-app/modules/sales-rep/specs/VCST-5707-lists-sharing/2026-09-14-lists-sharing-design.md`.
- `core/api/graphql/types.ts` carries two fields unrelated to this ticket from the dev environment (`isLockedForCurrentUser`, Loyalty's `storeId`). They go on the regeneration above.
- The core contract is regenerated and `@vc-frontend/core` goes 0.1.0 → **0.2.0**, the breaking level the README asks for on the 0.x line. `IWishlistSharingScopeControlsType.payload` changed shape and the scope element's props were renamed, but no export was removed — and the auto-bump only compares exported names, so it settled on a patch. That matters: a plugin built against 0.1.x declares `requiredHostVersion` "0.1.0", which `^0.1.0` still satisfies on 0.1.1, so it would load and then read props that no longer arrive. At 0.2.0 the manifest gate and the shared-singleton range (raised to match in `federation.mjs`) both refuse it with a reason. `vc-module-sales-rep`'s StorefrontApp is that plugin and needs a release of its own.
- The contract also picks up what landed on dev since it was last built: `isLockedForCurrentUser` (VCST-5317) and the `VcTable` header slot scope (VCST-5472).
- New test ids `wishlist-sharing-*`. The e2e flows in vc-testing-module that went through the old combined dialog need updating.
- "Remove list" follows the Acceptance's Action-wheel screenshot and the Figma menu; the Acceptance text says "Delete". Its icons deliberately do not: the Figma draws a pencil and a trash can, while `address-dropdown-menu` and `members-dropdown-menu` both use `edit` and `delete-2` and colour only the icon. Matching the Figma in one menu of three is the worse outcome, so this one keeps its siblings' pair and the Figma's belongs to a pass over all three.
- `viewItemList` now fires once per list id rather than on every reassignment of `list`. Saving from a dialog would otherwise send a second impression for the same items with no navigation in between; the same change also stops it re-firing on an item removal or a refetch of the open list.
- The dialog leaves out the "Notify via" checkboxes the Figma carries, and its recipient rows show the customer's address where the Figma shows an email — both settled with the requester. Every share goes out on email and push, so the checkboxes would never be anything but ticked and disabled, and there is no email search behind the picker.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5707
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.58.0-pr-2476-7045-7045037a.zip


client-app/pages/account/list-details.vue
```diff
@@ -59,15 +59,29 @@
...
         </div>
       </div>
 
       <div ref="listElement" class="mt-5 w-full">
-        <!-- Skeletons -->
-        <WishlistProductsSkeleton v-if="listLoading" :itemsCount="actualPageRowsCount" />
+        <!-- Skeletons: only while the list on screen is not the one being loaded. `listLoading` is shared by every
+             `useWishlists` caller, so a save from the Rename or Share dialog would otherwise swap the whole table for
+             a skeleton; `list` survives navigation, so `!list` alone would leave the previous list's frame up. -->
+        <WishlistProductsSkeleton v-if="listLoading && list?.id !== listId" :itemsCount="actualPageRowsCount" />
```
goldenmaya: Revisiting the same list now renders an empty widget instead of the skeleton.

Measured on a running app against a stubbed backend, sampling the DOM and `listLoading` every 200 ms while re-entering a list that had already been open once:

| t | listLoading | skeletons | rows |
|---|---|---|---|
| 1000 ms | true | 0 | 0 |
| 2000 ms | true | 0 | 0 |
| 3000 ms | false | 0 | 3 |

The ids match on a revisit, so the new `&& list?.id !== listId` suppresses the skeleton, while the next branch is already true because the module-scoped `list` survived navigation. The rows themselves come from the page-local `wishlistItems` buffer, refilled only after `fetchWishList` resolves, so the widget frame stands empty for the length of that request. At the merge base the branch read `v-if="listLoading"` and the skeleton covered this.

I forced a 2.5 s response to observe it comfortably; in production the gap equals the round trip.

Would you gate on the page buffer as well? Something like

    v-if="listLoading && (list?.id !== listId || !wishlistItems.length)"

keeps the case you fixed, since the buffer is non-empty when a dialog saves. I have not run that variant, and it does swap the empty view for a skeleton right after the last item is removed, so seeding the buffer from `list` at mount may suit the page better.

client-app/shared/wishlists/components/wishlist-dropdown-menu.vue
```diff
@@ -14,29 +15,82 @@
...
+        @click="
+          close();
+          $emit('remove', triggerElement());
+        "
+      >
+        <template #prepend>
+          <VcIcon name="trash-2" />
+        </template>
+
+        <span class="text-danger">{{ $t("shared.wishlists.list_card.remove_list_button") }}</span>
```
goldenmaya: The red label falls below AA once the row itself is coloured `danger`.

Read off the live page in the coffee preset, dark mode: label `#c00c0f` on the menu background `#110f0e` gives 3.00, against the 4.5 required for 14px regular text. Computed from the preset tokens, the danger hover fill takes it to 2.43. The base version kept the label at the kit's own colour and coloured only the icon, so the red text arrives with this PR.

Dropping the class leaves the red icon and the danger hover fill intact, and the kit's label colour measures between 14.17 and 18.12 on that fill across coffee and red in both modes.

```suggestion
        <span>{{ $t("shared.wishlists.list_card.remove_list_button") }}</span>
```

client-app/modules/sales-rep/components/wishlist-customer-sharing.vue
```diff
@@ -51,106 +82,208 @@ const props = defineProps<IProps>();
...
+  &__option-name {
+    // The leading is spelled out so the two lines total the avatar's 32 and the row stays balanced; the menu item
+    // hands down 14/14.
+    @apply truncate text-sm/[18px] font-bold text-neutral-950;
+  }
+
+  &__option-location {
+    // `font-normal` against the kit: `VcMenuItem` bolds the whole row of a selected option, and the second line is
+    // Regular in every state.
+    @apply truncate text-xs font-normal text-neutral-500;
```
goldenmaya: The second line of a picker option falls below AA on exactly the rows that matter most.

Measured live in coffee dark: `#848484` on the selected row fill `#342721` gives 3.85. In coffee light the same pair gives 3.75. Every already-shared customer renders as a selected row, so this is the field's default state rather than an edge case.

One step darker clears all four measured combinations, coffee and red in both modes, with 5.18 as the worst case.

```suggestion
    @apply truncate text-xs font-normal text-neutral-600;
```

client-app/core-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@vc-frontend/core",
-  "version": "0.1.0",
+  "version": "0.1.1",
```
goldenmaya: Should this be a minor bump rather than a patch?

The facade's own README says a breaking change on the 0.x line ships as `minor`, and `contract-versioning.mjs` only ever compares exported names, so it could not notice that `IWishlistSharingScopeControlsType.payload` changed shape or that the scope element's props were renamed from `sharedWithId` and `sharingLink` to `targets` and `message`. `PORT_TO_MF.md` lists that exposed shape as contract surface.

A plugin pinned to `^0.1.0` satisfies a host on 0.1.1, so it would load and then read props that no longer arrive.

The in-repo module moves in lockstep, so this only reaches an external plugin, and you may well know there is none yet. If that is the case, a line in the PR description would close the question; otherwise `yarn bump:core minor` looks like what the README asks for.

goldenmaya (commented):
Reviewed at `3f0ed6d5`. Gates I ran locally: eslint on the changed files is clean apart from two warnings that predate the diff, `vue-tsc` and `build-only` pass, `vitest` over the touched areas is 659 green, locale parity holds across all 13 files and the new plural keys render correctly at 0, 1, 2, 4, 5, 11, 21 and 101. SonarCloud reports no new issues on this PR. The five failing wishlist e2e jobs are the backend, not the code: only `vcptcore-dev` carries `targets`, exactly as the description says.

The scope registry seam, the delta write and the two-dialog split all hold up under probing. One thing worth stating plainly: I expected `onSaved` to see the already-updated `targets` and notify nobody, and that turned out to be wrong. A probe with the real modal and the real `useWishlists` showed the modal keeps the open-time snapshot at both call sites, so the counts and the notification are right. The counterfactual, merging into the same object, does break it, which is the one ask in the list below.

Four separate comments carry the substantive findings. The rest are minor, grouped here so they do not clutter the diff.

**Tests**

- `client-app/modules/sales-rep/components/wishlist-customer-sharing.test.ts:441-445` asserts that `wishlist-share-email-checkbox` and `wishlist-share-push-checkbox` are absent, but neither id exists in any component at base or head, so the test passes for any component. The positive case next to it already covers the behaviour.
- `client-app/shared/wishlists/components/share-wishlist-modal.vue:259` replacing the minted key with the server's is untested: neutering that line keeps all 44 tests green, because every fixture resolves the same key the list started with.
- `client-app/shared/wishlists/composables/useWishlistSharingScopes.ts:98-103` the duplicate-scope guard is untested: dropping the core half of `isTaken` keeps 51 tests green.
- `client-app/modules/sales-rep/components/wishlist-sharing-recipients.vue:142-148` the focus handoff from Clear all to Undo is untested: removing the focus call keeps 66 tests green.
- `client-app/pages/account/list-details.vue` has no test mounting it, and this PR adds `canShare`, the skeleton gate and the analytics dedup there.
- `client-app/shared/wishlists/components/share-wishlist-modal.vue:263` together with `client-app/shared/wishlists/composables/useWishlists.ts:33-41`: the notification is correct only because `applySavedList` builds new objects. One assertion that the pre-save object is left untouched would pin it.

**Accessibility**

- `client-app/modules/sales-rep/components/wishlist-sharing-recipients.vue:68` after Undo the focus lands on `body`, because the button that had it unmounts and nothing takes over. `requestClear` does the mirror move and could have a counterpart.
- `client-app/shared/wishlists/components/share-wishlist-modal.vue:12` the `legend` wraps a `VcLabel`, which renders a `div` when no `for-id` is passed, and `legend` takes phrasing content only. The test stub renders a `label`, so the suite cannot see it.

**Undeclared in the description**

- `client-app/pages/account/list-details.vue:450` and `458-459` the analytics dedup is not mentioned anywhere, and it reaches further than saving from a dialog: `viewItemList` now stops re-firing on any same-id reassignment of `list`, including item removal and every refetch of the open list.
- `client-app/shared/wishlists/components/wishlist-dropdown-menu.vue:29` and `62` swap `edit` and `delete-2` for `pencil` and `trash-2`, which leaves the three shared dropdown menus disagreeing, since `address-dropdown-menu.vue` and `members-dropdown-menu.vue` keep the old pair.
- The description has no list of renamed or removed public surface. The emits of `WishlistCard` and `WishlistDropdownMenu` changed, `SalesRepCommunicationInputType.organizationId` became `organizationIds`, and four locale keys were removed while two more became linked messages.

**Code and copy**

- `client-app/modules/sales-rep/components/wishlist-customer-sharing.vue:181` clearing `clearedRecipients` inside `deselect` cannot have an effect: every path that leaves a row selected has already emptied it.
- `client-app/modules/sales-rep/components/wishlist-customer-sharing.vue:144` the fallback to the drafted count masks a save that persisted nobody once VCST-5925 lands. It may belong in the Open section of the spec next to the regeneration.
- `client-app/shared/wishlists/components/wishlist-status.vue:55` and `client-app/modules/sales-rep/components/wishlist-customer-sharing.vue:150,160` pass a third argument that vue-i18n does not need: a named `count` already drives the form. I checked this against the real locale files, where `t(key, { count: 0 })` yields "Shared with 0 customers" in English and the correct form in Russian. The comment above the call is right about bare `t(key)` but attributes the save to the wrong argument.
- `client-app/shared/wishlists/components/share-wishlist-modal.vue:158` re-sorts a list the registry has already sorted, and `UNORDERED_SCOPE_POSITION` is exported only to make that copy possible.
- `client-app/modules/sales-rep/types/index.ts:142` duplicates `SalesRepCustomerOptionType` at `client-app/modules/sales-rep/composables/useSalesRepCustomerOptions.ts:12` field for field.
- `client-app/shared/wishlists/composables/useWishlistSharingScopes.ts:12` still describes the "Sharing options" select that this PR removes.
- `client-app/modules/sales-rep/components/wishlist-sharing-recipients.vue:173` uses `rounded-lg` where the app layer uses the radius token in 37 places, including every sales-rep sibling.
- `client-app/modules/sales-rep/locales/en.json:262,264` introduce a middle dot that the same files avoid elsewhere, and all 12 translations copied it. Japanese then mixes it with fullwidth parentheses in the neighbouring key.

Accessibility numbers were measured on the coffee and red presets only, in both modes, per the contrast policy for this repo.

client-app/pages/account/list-details.vue
ivan-kalachikov: Fixed, though not with the buffer check — the page now keys the skeleton off its own fetch rather than the shared `listLoading`, which covers the revisit and still keeps a dialog save from swapping the table. `list-details.test.ts` pins it in both directions: it fails against this PR's gate and against dev's.

client-app/shared/wishlists/components/wishlist-dropdown-menu.vue
ivan-kalachikov: Taken, and widened: the menu now follows `address-dropdown-menu` and `members-dropdown-menu` outright — `edit` / `delete-2`, `color="secondary"`, red on the icon only. The Figma does draw a red label with a pencil and a trash can, but matching it in one menu of three was the worse outcome, so that belongs to a pass over all three.

client-app/modules/sales-rep/components/wishlist-customer-sharing.vue
ivan-kalachikov: Taken as written — `neutral-600`.

client-app/core-api/package.json
ivan-kalachikov: Bumped: `yarn bump:core minor` to 0.2.0, with the `@vc-frontend/core` range in `federation.mjs` raised to `^0.2.0` alongside it. The plugin is not hypothetical — `vc-module-sales-rep`'s StorefrontApp registers this scope with the old shape. It pins the 0.1.0 tarball rather than a range, but both gates read `^0.1.0`, so on 0.1.1 it would have loaded and then read props that no longer arrive; at 0.2.0 they refuse it with a reason.

ivan-kalachikov:
On the grouped findings, at `a75c5851`.

**Taken**
- Undo now hands the focus back to Clear all, the mirror of what Clear all does.
- `legend` no longer wraps a `VcLabel`; the label's size-sm styling moved onto the `legend` itself.
- `deselect` no longer clears the undo buffer it could never hold.
- The scope tabs filter the registry's own order instead of re-sorting a copy, so `UNORDERED_SCOPE_POSITION` is no longer exported.
- Dropped the plural argument the named `count` already supplies, in all three places, and fixed the comment that credited it to the wrong one.
- `rounded-[--vc-radius]` on the recipients card.
- Stale "Sharing options" comment on `IWishlistSharingScopeType`.
- `SalesRepCustomerOptionType` and `WishlistSharingRecipientType` now both name one `OrganizationRowType`. Both names survive: one is filled from the rep's directory and the other from `sharingSetting.targets`, and a field only one of them grows belongs on that name rather than on the shape they share.
- The toast's draft fallback is now in the spec's Open section, to go with the regeneration.

**Tests added** — each verified to fail against the unfixed code
- The server's sharing key replacing the dialog's own; a customer notified with the wrong one gets a dead link.
- `applySavedList` leaving the pre-save object alone, which is the only reason the notification reaches anybody.
- The registry's duplicate guard, core half included (new `useWishlistSharingScopes.test.ts`).
- Clear all → Undo and Undo → Clear all focus handoffs.
- The skeleton gate on `list-details`, in a new page test.
- The two channel-checkbox assertions are gone: neither id exists in any component, and "always requests both channels" already covers the behaviour.

**Declined**
- The icons: the Figma draws a pencil and a trash can, so this menu was the one following the design and the other two had fallen behind. Matching the design in one menu of three is worse than matching the siblings, so it now uses `edit` / `delete-2` and the Figma's pair is left for a pass over all three.
- The middle dot: `Recipients · 30` and `Clear all` are in the Figma (nodes `483-9790`, `483-16521`), so the separator stays. The Japanese file did take the wrong character though — U+00B7 next to full-width kana, where U+30FB is the separator that sits in the same square. Fixed there only.

**Description** now carries the `viewItemList` dedup and how far it reaches, the renamed and removed surface, the 0.2.0 rationale, and the two places the dialog departs from the Figma on purpose ("Notify via", and the address where the Figma shows an email — both settled with the requester).

goldenmaya (dismissed):

client-app/shared/wishlists/components/wishlist-card.vue
```diff
@@ -32,8 +32,10 @@
     <div class="wishlist-card__dropdown">
       <WishlistDropdownMenu
         v-if="list.sharingSetting?.access === WishlistAccessType.Write"
-        @edit="$emit('settings')"
-        @remove="$emit('remove')"
+        :shareable="isCorporateMember && !!list.sharingSetting?.isOwner"
```
vas11yev1work, line 35: `canShare` in `list-details.vue:248` checks `isCorporateMember && isOwner && access === Write`, but here shareable only checks `isCorporateMember && isOwner` — the `Write` check is missing.

client-app/shared/wishlists/components/share-wishlist-modal.vue
```diff
@@ -0,0 +1,325 @@
...
+        <template #append>
+          <VcButton
+            v-if="isClipboardSupported"
+            data-test-id="wishlist-sharing-copy-link-button"
+            color="secondary"
+            variant="soft"
+            icon="document-duplicate"
+            icon-size="1.25rem"
+            @click="copySharingLink"
+          />
```
vas11yev1work, line 51: Can we reuse `VcCopyText` component from ui-kit?

client-app/shared/wishlists/components/share-wishlist-modal.vue
```diff
@@ -0,0 +1,325 @@
...
+      });
+    } catch (e) {
+      Logger.error("ShareWishlistModal: sharing scope onSaved failed", e);
+    }
+
+    try {
+      await fetchWishlists();
+    } catch (e) {
+      Logger.error("ShareWishlistModal: refreshing the lists after save failed", e);
+    }
```
vas11yev1work, line 277: `onSaved` and `fetchWishlists` are independent (separate `try/catch`, no shared data). Could we run them as `Promise.allSettled([...])` instead of sequential awaits to save latency?

client-app/shared/wishlists/components/wishlist-card.vue
ivan-kalachikov: The `Write` check is the `v-if` one line above — `WishlistDropdownMenu` is not created at all without it, so `shareable` is only ever evaluated when access is already `Write`. `list-details.vue` needs all three conjuncts because its Share button has no such gate, and this card is the menu's only consumer.

client-app/shared/wishlists/components/share-wishlist-modal.vue
ivan-kalachikov: Kept the field composition — `VcCopyText` is a text-plus-button row, so it would land the button outside the input at `xxs outline` and drop `data-test-id="wishlist-sharing-copy-link-button"` onto its wrapper div, which the e2e flows read. Took the part that mattered though: the button had no accessible name at all (`VcButton` renders `:aria-label="ariaLabel || title"` and neither was passed — inherited from the dialog this was split out of), so it now carries `ui_kit.buttons.copy`, with a test.

client-app/shared/wishlists/components/share-wishlist-modal.vue
ivan-kalachikov: Taken — one `Promise.allSettled`, with each rejection keeping its own log line. Added a test that pins the overlap: it fails against the sequential version.

vas11yev1work (approved):

PR #2477, opened 2026-09-10
VCST-5748: One-Time Password (OTP) Sign-In via Email

DmitryGrishinVirtoworks:
## Description

Adds a passwordless email OTP as an alternative sign-in method, alongside the existing password form.

- **`useEmailOtpAuthentication`** — reads the `VirtoCommerce.OTP` module's `OtpLogin.Enabled` store setting to decide whether the OTP option is shown at all; if the backend module isn't installed or enabled, the storefront falls back to password-only, unchanged.
- **`useOtpSignIn`** — talks to the module's `/api/otp/request` and `/api/otp/verify` endpoints, and completes sign-in on a successful code verification via the platform's `native_sign_in` token grant (`useAuth.nativeSignIn("OTP", { storeId, email, code })` + `signIn`) — the same email/code pair is re-verified by the platform itself, no external-login cookie involved.
- **`EmailOtpSignInForm`** (+ its `request`/`verify` sub-forms) — a 3-step flow (request code → enter 6-digit code → locked/generic terminal states), wired into both the dedicated sign-in page and the homepage banner's login section, next to the existing password form and identity providers.
- Full localization for all 13 supported locales.

No new dependencies. No breaking changes — purely additive UI, hidden entirely when the backend module is absent or the setting is off.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5748
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.58.0-pr-2477-a2f4-a2f4a647.zip


client-app/shared/sign-in/components/email-otp-verify-form.vue
```diff
@@ -0,0 +1,296 @@
...
+      await nextTick();
+      codeInputRef.value?.focus();
+      codeInputRef.value?.select();
+      return;
+    default:
+      errorMessage.value = t("shared.sign_in.email_otp_sign_in_form.verify.errors.generic");
+  }
+}
+
+async function onResend() {
```
ivan-kalachikov, line 196: **Bug:** a failed resend is reported as success.

`requestCode` resolves `undefined` on any HTTP/network failure: `useFetch` never throws, VueUse `execute()` returns `null`. This handler only checks `Disabled`, so on a 500 it clears the code and announces "New code sent".

Repro: SMTP down → click Resend → empty cells + "New code sent".

Fix: treat only `outcome === "Sent"` as success; otherwise keep the code and show `errors.generic` (same as `email-otp-request-form.vue` does with `!result`).

locales/en.json
```diff
@@ -607,6 +607,46 @@
...
+        },
+        "verify": {
+          "header": "Check your email",
+          "subtitle": "We sent a code to {email}.",
+          "code_label": "Code",
+          "paste_hint": "Paste is supported. On phones the code is offered above the keyboard.",
+          "submit_button": "Sign in",
+          "resend_button": "Resend code",
+          "use_different_email_link": "Use a different email",
+          "live_resent": "New code sent. The previous one no longer works.",
```
ivan-kalachikov: `live_resent` here and `invalid_code` (line 632) say earlier codes stop working. They don't.

- Backend uses Identity's TOTP `Email` provider: 3-minute step, ±2 window, codes are not consumed. A resend inside the same step returns the same code.
- WI test scenario F4 expects the 1st code to still verify after a 2nd request.

Same strings in the other 12 locales. Fix: drop the invalidation claim, e.g. "New code sent." and "That code isn't right. Check the latest email."

client-app/shared/sign-in/components/email-otp-sign-in-form.vue
```diff
@@ -0,0 +1,201 @@
...
+    <EmailOtpVerifyForm
+      v-else-if="step === 'verify' && pending"
+      :email="pending.email"
+      :masked-email="pending.maskedEmail"
+      @use-different-email="onUseDifferentEmail"
+      @disabled="step = 'generic'"
+      @locked="onLocked"
+    />
+
+    <div v-else-if="step === 'locked'" class="email-otp-sign-in-form__terminal">
```
ivan-kalachikov, line 14: Copy says "sign-in with a code is temporarily locked", but AC5 locks the whole account (shared Identity counter). "Sign in with a password instead" leads to the same lockout error one submit later.

Suggest: "Your account is temporarily locked" + `ContactAdministratorLink`, like `sign-in-form.vue` does on lockout.

Also (line 114): the countdown decrements per tick and drifts in throttled tabs, and an admin-locked account sends `2147483647` s → a 68-year timer. Compute from a deadline (`Date.now()`) and hide the timer for absurd values.

client-app/shared/sign-in/composables/useOtpSignIn.ts
```diff
@@ -0,0 +1,72 @@
...
+    loading.value = true;
+
+    try {
+      const { data } = await useFetch("/api/otp/verify")
+        .post({ storeId: globals.storeId, email, code })
+        .json<IOtpVerifyResponse>();
+
+      const result = data.value ?? undefined;
+
+      if (result?.outcome === "Success") {
```
ivan-kalachikov, line 54: No `analytics("login", …)` on this path. `sign-in-form.vue` and `pages/auth/callback.vue` both report success and failure; with OTP as the default method, these logins disappear from analytics.

Fix: extract the "complete external sign-in" step from `callback.vue` (`externalSignInCallback` + `signIn` + analytics + `SIGN_IN_NOT_ALLOWED` handling) and reuse it here.

client-app/pages/sign-in.vue
```diff
@@ -43,6 +58,29 @@ const route = useRoute();
 
 const returnUrl = computed<string>(() => getReturnUrl(route.fullPath));
 
+const { hasEmailOtpAuthentication } = useEmailOtpAuthentication();
+
+const signInMode = ref<"password" | "otp">(hasEmailOtpAuthentication.value ? "otp" : "password");
```
ivan-kalachikov: Layout bug for "Password off + identity provider + OTP on": `hasOnlyIdentityProviders` (line 25) is still `true`, so the providers render in the `--only` block under the OTP form and the `#side` divider block (line 31) is skipped.

Don't change `useIdentityProviders`: `impersonate.vue` relies on its current meaning and has no OTP form. Options:

1. Here: `const hasSignInForm = hasPasswordAuthentication || hasEmailOtpAuthentication`, then `hasIdentityProviders && !hasSignInForm` for `--only` and `hasIdentityProviders && hasSignInForm` for `#side`.
2. A `useSignInMethods()` composable that also absorbs the block duplicated in `login-form-section.vue`.

Either way, add the case to `sign-in.test.ts`.

client-app/shared/sign-in/components/email-otp-verify-form.vue
```diff
@@ -0,0 +1,296 @@
...
+  if (!isCodeComplete.value || loading.value) {
+    return;
+  }
+
+  try {
+    const result = await verifyCode(props.email, code.value);
+    await handleOutcome(result);
+  } catch (err) {
+    Logger.error("EmailOtpVerifyForm", err);
+    errorMessage.value = t("shared.sign_in.email_otp_sign_in_form.verify.errors.generic");
```
ivan-kalachikov, line 171: Identity errors from `/connect/token` after a successful verify collapse into the generic message. `callback.vue` redirects `SIGN_IN_NOT_ALLOWED` to `/400`; here the user just sees "Something went wrong".

Side effect: the failed grant leaves `errors` in `useAuth().data`, and `useSignMeIn` copies them via `syncRefs` (immediate). After "Sign in with a password instead" the password form opens with that stale alert.

Fix: read `useAuth().errors` here (map `SIGN_IN_NOT_ALLOWED`, translate the rest) and reset them when leaving the OTP flow.

client-app/shared/sign-in/components/email-otp-verify-form.vue
```diff
@@ -0,0 +1,296 @@
...
+
+    <div
+      class="email-otp-verify-form__field"
+      :class="{
+        'email-otp-verify-form__field--focused': isFocused,
+        'email-otp-verify-form__field--error': !!errorMessage,
+        'email-otp-verify-form__field--busy': loading,
+      }"
+    >
+      <input
```
ivan-kalachikov, line 38: The input never gets focus when this step appears. On phones the `one-time-code` suggestion shows only for a focused field; the prototype focuses it.

Fix: `onMounted(() => codeInputRef.value?.focus())`.

Also: this step is a `<div>` with a click button, so Enter does nothing. Wrap it in `<form @submit.prevent="onSubmit">`.

client-app/shared/sign-in/components/email-otp-verify-form.vue
```diff
@@ -0,0 +1,296 @@
...
+      />
+
+      <div class="email-otp-verify-form__cells" aria-hidden="true">
+        <div
+          v-for="index in CODE_LENGTH"
+          :key="index"
+          class="email-otp-verify-form__cell"
+          :class="{
+            'email-otp-verify-form__cell--filled': !!code[index - 1],
+            'email-otp-verify-form__cell--active': isFocused && index - 1 === code.length && code.length < CODE_LENGTH,
```
ivan-kalachikov, line 61: With 6 digits `cell--active` is never applied (`code.length < CODE_LENGTH`), and the input hides caret, outline and selection. After `InvalidCode` the field is focused and selected but nothing shows it (WCAG 2.4.7).

Fix: keep a visible focus style on the field itself (e.g. `&__field:focus-within` ring) regardless of length.

client-app/shared/sign-in/components/email-otp-verify-form.vue
```diff
@@ -0,0 +1,296 @@
...
+        type="button"
+        class="email-otp-verify-form__link"
+        :disabled="loading"
+        data-test-id="email-otp-resend-button"
+        @click="onResend"
+      >
+        {{ $t("shared.sign_in.email_otp_sign_in_form.verify.resend_button") }}
+      </button>
+
+      <button type="button" class="email-otp-verify-form__link" @click="emit('useDifferentEmail')">
```
ivan-kalachikov, line 99: Enabled while `loading`. A slow verify keeps running after the user leaves this step: token grant + `signIn` complete for the abandoned email.

Fix: `:disabled="loading"`, same as the resend link.

client-app/shared/sign-in/components/email-otp-verify-form.vue
```diff
@@ -0,0 +1,296 @@
...
+      <input
+        id="email-otp-code"
+        ref="codeInputRef"
+        v-model="code"
+        class="email-otp-verify-form__input"
+        type="text"
+        inputmode="numeric"
+        autocomplete="one-time-code"
+        pattern="[0-9]*"
+        :maxlength="CODE_LENGTH"
```
ivan-kalachikov: `maxlength` truncates a pasted value before `onInput` strips non-digits: `" 418302"` (leading space from an email) becomes `41830`.

`onInput` already slices to `CODE_LENGTH`; drop `:maxlength`.

client-app/shared/sign-in/components/email-otp-verify-form.vue
```diff
@@ -0,0 +1,296 @@
...
+  try {
+    const result = await requestCode(props.email);
+
+    if (result?.outcome === "Disabled") {
+      emit("disabled");
+      return;
+    }
+
+    code.value = "";
+    liveMessage.value = t("shared.sign_in.email_otp_sign_in_form.verify.live_resent");
```
ivan-kalachikov, line 215: Same string on every resend → the live-region text doesn't change → the second resend isn't announced.

Fix: clear `liveMessage` first, set it on `nextTick`.

client-app/pages/sign-in.vue
```diff
@@ -2,10 +2,23 @@
...
+        v-if="showEmailOtpForm"
+        :has-password-authentication="hasPasswordAuthentication"
+        @switch-to-password="switchToPassword"
+        @step-changed="otpStep = $event"
+      />
+
+      <template v-else>
+        <SignInForm v-if="hasPasswordAuthentication" />
+
+        <button v-if="hasEmailOtpAuthentication" type="button" class="sign-in__switch-link" @click="switchToOtp">
```
ivan-kalachikov: With OTP as the default, the Register button (inside `SignInForm`) disappears from this page and from the home section; only the header link remains. Worth a product check.

Also: this switch link and the ones in `email-otp-sign-in-form.vue` have no `data-test-id`. QA page objects that click `login-button` break once OTP is enabled on QA.

client-app/shared/sign-in/composables/useOtpSignIn.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    expect(postedBody).toEqual({ storeId: "store-1", email: "buyer@acme.com" });
+    expect(result).toEqual({ outcome: "Sent", maskedEmail: "b•••r@acme.com" });
+    expect(loading.value).toBe(false);
+  });
+
+  it("requestCode logs and rethrows when the request fails", async () => {
+    const fetchState = await getFetchState();
+    fetchState.useFetch.mockImplementation(() => ({
+      post: () => ({
+        json: () => Promise.reject(new Error("network error")),
```
ivan-kalachikov: These mocks reject from `json()`, but the real wrapper never rejects: `createFetch` resolves with `data = null` and `error` set. The path that matters (`requestCode` → `undefined`, see the resend bug) is untested.

Add a case with `fetchResult.data.value = null` and `error.value` set, plus component tests for the three forms. `sign-in.test.ts:19` pins OTP off, so the page's OTP branch is untested too.

client-app/shared/sign-in/composables/useEmailOtpAuthentication.ts
```diff
@@ -0,0 +1,19 @@
+import { computed } from "vue";
+import { useThemeContext } from "@/core/composables";
+
+const OTP_MODULE_ID = "VirtoCommerce.Otp";
+const OTP_ENABLED_SETTING_NAME = "OtpLogin.Enabled";
+
+export function useEmailOtpAuthentication() {
+  const { modulesSettings } = useThemeContext();
+
+  const hasEmailOtpAuthentication = computed(() => {
```
ivan-kalachikov: Same as `useModuleSettings(OTP_MODULE_ID).isEnabled(OTP_ENABLED_SETTING_NAME)` (12 composables use it; `useSalesRepsConfig` is the one-liner). Boolean settings arrive as booleans, so the `"true"` branch isn't needed. Ids belong in `core/constants/modules.ts`.

client-app/shared/layout/components/login-form-section/login-form-section.vue
```diff
@@ -43,19 +61,45 @@
...
 const { isAuthenticated } = useUser();
+const { hasEmailOtpAuthentication } = useEmailOtpAuthentication();
 
 const bgImage = computed(() =>
   themeContext.value.settings.homepage_background_image
     ? getImageUrl(themeContext.value.settings.homepage_background_image)
     : "none",
 );
+
+const signInMode = ref<"password" | "otp">(hasEmailOtpAuthentication.value ? "otp" : "password");
```
ivan-kalachikov: Lines 83-101 are a verbatim copy of `sign-in.vue:63-88` (`signInMode`, `otpStep`, `switchToOtp/Password`, title). One `useSignInMode()` composable for both.

`has-password-authentication` is also hard-coded `true` here (line 11) while the page derives it from the store.

client-app/shared/sign-in/components/email-otp-sign-in-form.vue
```diff
@@ -0,0 +1,201 @@
...
+const DEFAULT_LOCKOUT_SECONDS = 900;
+
+const step = ref<StepType>("request");
+const terminalHeadingRef = ref<HTMLElement>();
+const pending = ref<{
+  email: string;
+  maskedEmail: string;
+}>();
+
+function createCountdown() {
```
ivan-kalachikov, line 119: `createCountdown` re-implements vueuse `useCountdown` (available in `@vueuse/core` 14.3).

Nearby: `onUseDifferentEmail` and `onStartOver` are identical (one `reset()`), the two `watch(step)` can be one, and `DEFAULT_LOCKOUT_SECONDS` is never used by the contract (`Locked` always carries `lockoutSecondsRemaining`).

client-app/shared/sign-in/composables/useOtpSignIn.ts
```diff
@@ -0,0 +1,72 @@
...
+  async function requestCode(email: string): Promise<IOtpRequestResponse | undefined> {
+    loading.value = true;
+
+    try {
+      const { data } = await useFetch("/api/otp/request")
+        .post({ storeId: globals.storeId, email })
+        .json<IOtpRequestResponse>();
+
+      return data.value ?? undefined;
+    } catch (e) {
```
ivan-kalachikov: Unreachable: `useFetch(...).post().json()` resolves on HTTP and network errors, so this `catch` (and the matching one in `email-otp-request-form.vue`) never runs. Either check `error.value` and throw here, or drop the `try/catch`.

client-app/shared/sign-in/components/email-otp-verify-form.vue
```diff
@@ -0,0 +1,296 @@
...
+    &:hover {
+      @apply text-[--link-hover-color];
+    }
+
+    &:disabled {
+      @apply cursor-not-allowed opacity-50;
+    }
+  }
+
+  &__sr-only {
```
ivan-kalachikov: Tailwind `sr-only` already exists (11 uses in the repo). Also `&--filled` (line 239) repeats the base border, so the modifier changes nothing.

client-app/pages/sign-in.vue
```diff
@@ -43,6 +58,29 @@ const route = useRoute();
...
+  otpStep.value = "request";
+  signInMode.value = "otp";
+}
+
+function switchToPassword() {
+  otpStep.value = "request";
+  signInMode.value = "password";
+}
+
+const pageTitle = computed(() =>
```
ivan-kalachikov, line 75: `t` is declared at line 84, after this computed. It works only because `computed` is lazy. Move `const { t } = useI18n()` up.

client-app/shared/sign-in/components/email-otp-verify-form.vue
DmitryGrishinVirtoworks: Fixed. `onResend` now only treats `outcome === "Sent"` as success; anything else (including the silent `undefined` a failed `useFetch` call returns) shows the generic error and keeps the code intact. Added a regression test for this exact case (`email-otp-verify-form.test.ts`).

locales/en.json
DmitryGrishinVirtoworks: Fixed in all 13 locales — dropped the invalidation claim (`live_resent`, `invalid_code`). Matches what we established earlier in this project: the backend TOTP code within a step is not consumed/invalidated by use or by a resend.

client-app/shared/sign-in/components/email-otp-sign-in-form.vue
DmitryGrishinVirtoworks: Fixed the copy (now says the account is locked, not "sign-in with a code") and the drift/overflow issue: the countdown is now computed from a fixed deadline (`Date.now()` based) instead of decrementing per tick, and an absurdly large `lockoutSecondsRemaining` (e.g. from an admin-imposed permanent lock — confirmed `SecurityController.SetLockoutEndDateAsync(user, DateTime.MaxValue)` in vc-platform) is now treated as indefinite: no countdown, "contact administrator" instead.

One deliberate deviation: I did *not* add `ContactAdministratorLink` to the ordinary timed lockout, only to the indefinite one. The timed case already tells the user exactly when they can retry (self-service); `sign-in-form.vue` itself only shows the link for the permanent-style lockout codes, not `user_is_temporary_locked_out`. Happy to add it to both if you feel strongly.

client-app/shared/sign-in/composables/useOtpSignIn.ts
DmitryGrishinVirtoworks: Fixed — `useOtpSignIn`'s new `completeSignIn()` now calls `analytics("login", "otp", ...)` on both success and failure, mirroring `sign-in-form.vue`/`callback.vue`.

client-app/pages/sign-in.vue
DmitryGrishinVirtoworks: Fixed — `sign-in.vue` now derives its own `hasSignInForm = hasPasswordAuthentication || hasEmailOtpAuthentication` instead of relying on `hasOnlyIdentityProviders`, which doesn't know about OTP. Left `useIdentityProviders` itself untouched since `impersonate.vue` depends on its current meaning. Added a test case for this exact scenario.

client-app/shared/sign-in/components/email-otp-verify-form.vue
DmitryGrishinVirtoworks: Fixed the `SIGN_IN_NOT_ALLOWED` handling and added a visible error alert (same `ContactAdministratorLink`-for-lockout pattern as `sign-in-form.vue`), plus a `resetSignInErrors()` call before every new verify attempt.

Left one part open: the deeper issue that a stale error can leak from `useAuth`'s shared `errors` into a *different* form (e.g. switching to the password form right after) isn't fully fixed — `useAuth.ts` has no way to clear its underlying `data` ref, and it's a shared composable used well beyond this feature. Didn't want to change its architecture inside this PR without a separate discussion.

client-app/shared/sign-in/components/email-otp-verify-form.vue
DmitryGrishinVirtoworks: Fixed both — `onMounted` now focuses the code input, and the template is a `<form @submit.prevent="onSubmit">` with the button as `type="submit"`, so Enter submits.

client-app/shared/sign-in/components/email-otp-verify-form.vue
DmitryGrishinVirtoworks: Fixed with a `&__field:focus-within` ring on the whole cell group, so focus stays visible regardless of code length/cursor position, including right after an `InvalidCode` refocus+select.

client-app/shared/sign-in/components/email-otp-verify-form.vue
DmitryGrishinVirtoworks: Checked this one — it's not actually a bug. `VcButton` computes `enabled = !disabled && !loading` internally and binds the native `disabled` attribute from that, so it already can't be clicked while `loading` is true regardless of the `disabled` prop we pass. `onSubmit()` also has its own `if (!isCodeComplete.value || loading.value) return;` guard. No change made.

client-app/shared/sign-in/components/email-otp-verify-form.vue
DmitryGrishinVirtoworks: Fixed — dropped `:maxlength`. `onInput` already slices to `CODE_LENGTH` after stripping non-digits, so the native truncation was the only thing that could cut a pasted code short.

client-app/shared/sign-in/components/email-otp-verify-form.vue
DmitryGrishinVirtoworks: Fixed — `liveMessage` is cleared and re-set on `nextTick()` so a second resend produces an actual text change or the live region for screen readers.

client-app/pages/sign-in.vue
DmitryGrishinVirtoworks: Added `data-test-id` to all three password/OTP switch links so QA selectors don't depend on which form happens to be visible.

Left the Register-button question alone — you flagged it yourself as a product question rather than a bug, and that's not something for me to decide in this PR.

client-app/shared/sign-in/composables/useOtpSignIn.test.ts
DmitryGrishinVirtoworks: Fixed — the fetch mock in `useOtpSignIn.test.ts` now resolves with `data: null` / `error` set instead of rejecting, matching what `useFetch` actually does. Also added dedicated component tests for all three forms (`email-otp-request-form.test.ts`, `email-otp-verify-form.test.ts`, `email-otp-sign-in-form.test.ts`) covering this bug, the lockout/indefinite-lockout branch, and `sign-in.test.ts`'s OTP-on branch.

client-app/shared/sign-in/composables/useEmailOtpAuthentication.ts
DmitryGrishinVirtoworks: Fixed — rewritten around `useModuleSettings(MODULE_ID_OTP).isEnabled(OTP_ENABLED_KEY)`, same pattern as `useXPickup`/`useSalesRepsConfig`. Constants moved to `core/constants/modules.ts`. Dropped the `=== "true"` branch since `isEnabled` already normalizes to boolean.

client-app/shared/layout/components/login-form-section/login-form-section.vue
DmitryGrishinVirtoworks: Fixed — extracted the shared state (`signInMode`/`showEmailOtpForm`/`otpStep`/`switchToOtp`/`switchToPassword`) into `useOtpSignInMode`, used by both. Also fixed `has-password-authentication` being hardcoded `true` here — it now comes from `useIdentityProviders()`, same as the sign-in page.

client-app/shared/sign-in/components/email-otp-sign-in-form.vue
DmitryGrishinVirtoworks: Checked `useCountdown` from the installed `@vueuse/core` (14.3) — it also decrements `remaining - 1` per tick internally, so it has the same drift-in-a-throttled-tab problem this comment (rightly) flagged for our own `createCountdown`. Swapping to it would've reintroduced the bug we just fixed by computing from a deadline instead. Kept the custom implementation for that reason.

Did apply the other two: merged `onUseDifferentEmail`/`onStartOver` into one `resetToRequest`, and merged the two `watch(step)` calls into one. On `DEFAULT_LOCKOUT_SECONDS`: it is actually used, as the fallback when `lockoutSecondsRemaining` is `undefined` — kept as-is.

client-app/shared/sign-in/composables/useOtpSignIn.ts
DmitryGrishinVirtoworks: Fixed here — dropped the dead `catch` in `useOtpSignIn.ts`, kept `try/finally` so `loading` still resets. Deliberately left the component-level catches in `email-otp-request-form.vue`/`email-otp-verify-form.vue` alone: those are a defensive UI boundary (show a generic error if anything ever does throw), not a pointless log-and-rethrow like the one removed here, and they're exercised by the new tests (`verifyCode`/`requestCode` mocked to reject).

client-app/shared/sign-in/components/email-otp-verify-form.vue
DmitryGrishinVirtoworks: Fixed both — dropped the no-op `&--filled` modifier (and the now-unused class binding), and replaced the hand-rolled `&__sr-only` with Tailwind's `sr-only` directly on the element, same as `coupons-section.vue`'s aria-live region.

client-app/pages/sign-in.vue
DmitryGrishinVirtoworks: Fixed — moved `const { t } = useI18n()` above the `pageTitle` computed.

PR #2484, opened 2026-09-11
fix(VCST-5912): fix false invalid state and wrong values on the quantity spinbutton

goldenmaya:
## Description

The quantity spinbutton announced a false state before the user touched it: an untouched `0` was reported as invalid, and the value read out was `min` rather than the `0` on screen.

**Root cause.** `min` is the *stepping floor*, not the range the control accepts — with `allowZero` (the default) `0` stays reachable below it. Publishing `min` as the input's own bound made an untouched `0` a genuine constraint violation, and Chrome clamps a spinbutton's announced value into `[valuemin, valuemax]`, so the screen said `0` while a screen reader said `3`.

Verified on the same Storybook story before and after:

```
before   invalid=true  value=3  valuemin=3  valuemax=10  valuetext=0
after                  value=0  valuemin=0  valuemax=10  valuetext=0
```

**What changed**

- `VcQuantityStepper` now separates the two meanings: `stepMin`/`stepMax` feed the stepping helpers, `boundMin` is what the input and `aria-valuemin` publish. Stepping behaviour is unchanged — 1→3 and 2→0 across the gap still hold, and there are now tests for both.
- `max` is published raw. Omitting it means unbounded and nothing is announced, instead of advertising `9007199254740991`.
- `aria-valuenow` / `aria-valuemax` use `null` rather than `""` when absent — an empty ARIA value is not the same as no value.
- `VcInput` exposes `aria-invalid` from its `error` prop, with a consumer override through `aria["aria-invalid"]`.

**Two decisions worth flagging to reviewers**

*Why `:max="max"` and not `:max="stepMax"`.* `stepMax` is `props.max ?? Number.MAX_SAFE_INTEGER`, an internal fallback needed only because the stepping helpers read `max === 0` as "unbounded". Publishing it would restore the ticket's own complaint — an absurd upper bound read out to the user. That `max === 0` ambiguity is real but pre-existing and now tracked separately.

*Why `aria-invalid` on `VcInput` is in this PR and not in the a11y follow-ups.* It is not an addition — it replaces a signal this fix removes. The native `rangeUnderflow` used to mark the whole band below `min` as invalid, including values that genuinely are (2 with a minimum of 3). Moving the bound to `0` removes that signal for the legitimate case too, so `aria-invalid` plus `aria-describedby` put it back, now with a message that explains it:

```
2 with min 3   min="0"  rangeUnderflow=false  aria-invalid="true"  aria-describedby → message
```

The same treatment for `VcTextarea`, `VcCheckbox` and `VcDatePicker` is deliberately *not* here — nothing in this PR breaks without it.

**Follow-ups filed**

- [VCST-5990](https://virtocommerce.atlassian.net/browse/VCST-5990) — with `allowZero: false` (cart and quote line items) a value below the floor is a truthful constraint violation that nothing explains. Different failure from this ticket: every attribute is correct, the message is missing.
- [VCST-5991](https://virtocommerce.atlassian.net/browse/VCST-5991) — `max={0}` means "unbounded" to the stepping helpers and "ceiling is zero" to the DOM. Measured: the control announces a ceiling of `0`, reports overflow, and still steps 5→6. Unreachable from the shipped app, a trap for forks.
- [VCST-5993](https://virtocommerce.atlassian.net/browse/VCST-5993), [VCST-5994](https://virtocommerce.atlassian.net/browse/VCST-5994) — `VcTextarea` accessible name, and the error state on the remaining controls.

**Verification**

Storybook AX tree before/after (above); unit coverage for the stepper bounds, `QuantityControl` in stepper mode against the mission-modal row, and `VcInput`'s `aria-invalid`; each new behaviour mutation-tested. `yarn lint` clean — 7 warnings, identical at the merge base. `validate:types`, `build-only` and the full suite green.

### Jira-link:

https://virtocommerce.atlassian.net/browse/VCST-5912
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.58.0-pr-2484-ae2f-ae2f564d.zip

---

> [!NOTE]
> **Medium Risk**
> Touches shared UI primitives (`VcInput`, `VcQuantityStepper`) used across cart and product flows; behavior is accessibility-focused but changes published min/max and invalid signaling.
> 
> **Overview**
> Fixes **VCST-5912**: quantity spinbuttons no longer announce a false invalid state or the wrong value when `allowZero` shows `0` below a product minimum.
> 
> **`VcQuantityStepper`** splits the stepping floor (`stepMin` / internal `stepMax`) from what the native input and ARIA expose (`boundMin` is `0` when `allowZero`, otherwise the floor). Omitted `max` stays unbounded in the DOM instead of advertising `Number.MAX_SAFE_INTEGER`; `aria-valuenow` / `aria-valuemax` use `null` when there is nothing to announce. Increment/decrement behavior across the gap below `min` is unchanged and now covered by tests.
> 
> **`VcInput`** maps the visual `error` prop to **`aria-invalid`**, with sensible handling of consumer overrides via `aria["aria-invalid"]`, so validation errors (e.g. quantity between 0 and minimum) are exposed to assistive tech together with `aria-describedby`.
> 
> Adds unit tests for stepper bounds, `QuantityControl` in stepper mode, and `VcInput` `aria-invalid`, plus a Storybook **`AllowZeroBelowMin`** example.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae2f564d6f12c406715f571817e0ca6c02afbeb8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

[VCST-5990]: https://virtocommerce.atlassian.net/browse/VCST-5990?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


client-app/ui-kit/components/organisms/quantity-stepper/vc-quantity-stepper.vue
```diff
@@ -15,15 +15,15 @@
       :name="name"
       :show-empty-details="showEmptyDetails"
       single-line-message
-      :min="min"
+      :min="boundMin"
       :max="max"
```
vas11yev1work, line 19: Let's change `:max="max"` to `:max="stepMax"`. The same thing with `aria-valuemax`

vas11yev1work (dismissed):

client-app/ui-kit/components/organisms/quantity-stepper/vc-quantity-stepper.vue
ivan-kalachikov: Please don't — `stepMax` is the internal fallback for the stepping math, not a bound to publish.

```
:max="max"      → no max attribute        → nothing announced   ✅
:max="stepMax"  → max="9007199254740991"  → read out by AT      ❌
```

It's `props.max ?? Number.MAX_SAFE_INTEGER` (L104), needed only because the helpers read `max === 0` as "unbounded" (`quantity-stepper.ts:62, 76, 81`). Publishing it restores the ticket's own row: *"aria-valuemax = 9007199254740991 — an absurd upper bound announced to the user"*.

Pre-existing and worth its own ticket: `max={0}` means "unbounded" to the helpers but "maximum is zero" to the DOM. No consumer hits it today.

client-app/ui-kit/components/organisms/quantity-stepper/vc-quantity-stepper.vue
```diff
@@ -89,14 +91,18 @@ interface IProps {
...
 });
 
 const lastNonEmptyValue = ref<number | undefined>(undefined);
-const min = computed(() => props.min ?? (props.allowZero ? 0 : 1));
+
+// `allowZero` keeps 0 reachable below `stepMin`, so publishing `stepMin` as the input's own bound
+// makes an untouched 0 a constraint violation.
+const stepMin = computed(() => props.min ?? (props.allowZero ? 0 : 1));
+const stepMax = computed(() => props.max ?? Number.MAX_SAFE_INTEGER);
+const boundMin = computed(() => (props.allowZero ? 0 : stepMin.value));
```
ivan-kalachikov, line 105: `boundMin` only covers `allowZero: true`. With `false` the input's `min` is still the stepping floor, so a value below it reproduces VCST-5912 on untouched render:

```
<input type=number min="1" value="0" aria-valuemin="1" aria-valuenow="0">
  below min       → browser exposes the field as invalid, no message to explain it
  below valuemin  → value announced clamped to "1" while the screen shows "0"
```

Second line is the ticket's row: *"aria-valuenow = 1 — the visible input shows 0"*.

Who lands there: `QuantityControl` defaults to `allowZero: false`; `cart-line-items.vue:27` and `quote-line-items.vue:45` don't pass `allow-zero`; `settings_data.json:16` ships `"stepper"` as the default mode.
Concretely: quote-line-items passes `item.selectedTierPrice?.quantity` with no `min-quantity`, and `selectedTierPrice` is optional in the schema → `min="1" value="0"`.

Fair counterpoint: with `allowZero: false` a 0 really is invalid, so that half is honest — the clamped announcement isn't.

Pre-existing, and I haven't proven the server returns such rows. Follow-up ticket rather than a change here?

client-app/ui-kit/components/molecules/input/vc-input.vue
```diff
@@ -186,6 +189,17 @@ const describedBy = computed(() => {
   return ids || undefined;
 });
 
+// An unrecognised aria-invalid token means "true" per ARIA, and axe rejects it outright.
+const invalid = computed<AriaAttributes["aria-invalid"]>(() => {
```
ivan-kalachikov, line 193: `""` shouldn't collapse to `"true"` — per WAI-ARIA an empty `aria-invalid` means *not* invalid: *"If the attribute is not present, or its value is false, or its value is an empty string, the default value of false applies."*

`"false"` would be wrong too: an explicit `false` suppresses the browser's own constraint-validation state, so a genuinely invalid field stops being announced. Empty doesn't.

Treat it as "no override":

```ts
if (override == null || override === "") {
  return props.error ? "true" : undefined;
}
```

The rest stays correct — unknown token → `"true"`, the number `0` included (serialises to `"0"`).

Nits: axe 4.12 routes an empty ARIA value to *incomplete / needs review* (`aria-valid-attr-value`, messageKey `empty`), not violations — the comment above overstates it. And drop `""` from the `collapses the unusable consumer value` cases.

Nobody in this repo passes `""` — contract hygiene for forks, not a blocker.

client-app/ui-kit/components/organisms/quantity-stepper/vc-quantity-stepper.vue
goldenmaya: Keeping `:max="max"` — agreed with the reasoning above. `stepMax` exists only so the stepping helpers get a number; publishing it puts `9007199254740991` back into the accessibility tree, which is one of the rows the ticket lists.

On the `max === 0` ambiguity flagged in that reply: confirmed by measurement, and worse than described. With `max={0}` and a value of 5 the control is simultaneously over its announced ceiling and freely steppable above it:

```
max attribute      "0"
aria-valuemax      "0"
rangeOverflow      true
increment button   enabled
clicking +         emits 6
```

Filed as VCST-5991 with the root cause (`quantity-stepper.ts:62, 76, 81` use `0` as the sentinel for "unbounded", while the component already has a correct one — `undefined`). Not touched here; pre-existing and unreachable from the shipped app.

client-app/ui-kit/components/organisms/quantity-stepper/vc-quantity-stepper.vue
goldenmaya: Confirmed, including the consumer chain. Filed as VCST-5990 rather than changed here.

Reproduced by mounting `QuantityControl` with each consumer's exact props:

```
quote row, no tier quantity   shown="0" min="1" aria-valuemin="1" aria-valuenow="0"
                              rangeUnderflow=true  aria-invalid=absent  message=null
cart row, below minimum       shown="1" min="3" aria-valuemin="3" aria-valuenow="1"
                              rangeUnderflow=true  aria-invalid=absent  message=null
```

Two things the probe added to your description:

**The value/min half is the only half that reproduces.** `aria-invalid` stays absent on an untouched render — `isValid` in `useQuantityField` starts `true` and only moves on an explicit `validateFields()`, which nothing calls there. So the residual is exactly your counterpoint: the attributes are all truthful, and nothing explains the violation. That makes it 3.3.1, not the 4.1.2 this ticket fixes — which is the main reason it reads better as its own ticket than as a change here.

**The sharper symptom is an asymmetry.** The same `0` in a quote row is explained if the user typed it and silent if it came from data:

```
user clears the field to 0   aria-invalid="true"  message="…add_to_cart.errors.positive_number"
row renders 0 from data      aria-invalid=absent  message=null
```

And the app treats a zero quote quantity as a real state — `edit-quote.vue:241` blocks submit on it.

On "haven't proven the server returns such rows": still unproven for **cart**, and the ticket says so — I derived that shape from the component contract and it needs a QA check. The **quotes** path does not need one: `selectedTierPrice` is optional in the schema and `defineModel` supplies `0`.

Two fixes that look cheap were measured and rejected in the ticket: adding `allow-zero` to `quote-line-items` deletes the message from the typed-in case, and lowering the published floor to the rendered value silently removes `rangeUnderflow` on cart and quotes, so the browser stops blocking submit of an under-minimum quantity.

client-app/ui-kit/components/molecules/input/vc-input.vue
goldenmaya: You're right on both counts. Fixed in f59074c7d.

Empty is now treated as no override, exactly as you wrote it, and the distinction from `"false"` matters for the reason you give — an explicit `false` suppresses the browser's own constraint-validation state, an empty value does not.

The axe correction is also right, and it corrects my own earlier measurement rather than just my wording: an empty ARIA value routes to *incomplete*, and it was `"0"` that produced the violation. The comment overstated it; rewritten.

Also dropped `""` from the `collapses the unusable consumer value` cases and gave it its own test, since it now asserts the opposite outcome. Mutation-checked: restoring the old condition turns that test red.

ivan-kalachikov (dismissed):

ivan-kalachikov (approved):

PR #2488, opened 2026-09-16
feat(VCST-5628): returns module — list, create wizard, details, cancel

basilkot:
## Description

Adds `client-app/modules/returns`: a signed-in customer lists their own returns, starts one from a delivered order, picks lines and quantities, gives a reason, attaches files per line, submits, and can withdraw the request. Requires the `vc-module-return` PR for VCST-5628.

### What is here

- **Module**: routes (list, new from order, edit draft, details), account menu entry, composables over the new `return` xAPI scope, 13 locale files.
- **Attachments** reuse the shared `useFiles` composable and `VcFileUploader`, so the storefront enforces exactly what the upload scope is configured with. Only files that finished uploading carry a URL and can be attached — one still in flight would be claimed as an empty reference.
- **Entry point from the order.** `RequestReturnButton` on the order details page, next to Print and Reorder. It asks `returnableItems` — the same query the wizard opens with, so the wizard costs nothing extra afterwards — and renders only when the order still has a returnable line.
- **Cancelling.** A modal asks for an optional reason. Whether the action is offered at all comes from the server's `availableActions`, and an action that exists but is unavailable renders as a disabled control with the reason in a tooltip. A refused cancel leaves the modal open so the error lands next to what it is about.
- **Errors reach the buyer.** `useReturnErrors` turns the flow service's codes into localized text — `RETURN_QUANTITY_UNAVAILABLE` above all, which is the one failure a correctly filled draft can still hit, since availability is settled at submit.
- Gated on `Return.ReturnEnabled`; reads `Return.AttachmentsRequired` from module settings, so turning the requirement off in admin also unblocks submit here. The upload scope is no longer a setting — see the module PR.
- Statuses render `statusDisplayValue`, resolved by the server from the `Return.Status` dictionary in the current culture.

### Outside the module — please review separately

**`client-app/shared/files/composables/useFiles.ts` (`6eb52b771`)** fixes a hang that is not specific to returns. Upload results are matched to files by name, and a rejected upload can come back without one — `INVALID_SCOPE` names the scope, not the file. Nothing matched, the file stayed in `uploading` forever, and `uploadFiles()` — which retries while any file is new or uploading — recursed until the tab stopped responding. The same happened when the response carried no results at all. Any file the response does not account for is now failed, carrying the server's message where there is one.

Reachable from every consumer of `useFiles`: quotes, company info, product reviews, purchase requests, configurable product options. It only takes one misconfigured scope.

**`client-app/pages/account/order-details.vue`** gains two lines: the module gate and the button. The page only checks whether the module is enabled, the way `product.vue` gates customer-reviews — a shared page should not carry return rules.

### Also

`useReturnDraft.save` and `.submit` now handle a rejected mutation. `save` runs unattended on every keystroke, so its rejection went nowhere at all; and a failed save no longer lets submit hand over a draft missing what the buyer just typed.

### Deliberate

The quantity field commits while the buyer types rather than on blur, because Continue stays disabled until something is selected and a disabled button never gives the input a blur to react to. Typing above the maximum corrects the field on the spot, so the story's "must not accept more than is available" holds literally.

### Known gaps

- Error messages are translated into English and Russian; the other eleven locales carry the English text until a translator reaches them.
- Status and reason labels come from the platform's localization store, which an operator fills in the admin — see the module PR. Until then the storefront shows the raw codes.
- No component tests.

### Verification

`vue-tsc --build` clean, eslint clean, prettier clean, `check-locales` clean. Manually on a 3.1069 stand: the full round trip — start from an order, fill the draft, upload files, submit, cancel, and the returnable quantity restored afterwards.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5628
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.59.0-pr-2488-4c6b-4c6bb181.zip

---

> [!NOTE]
> **Medium Risk**
> Large new order/returns customer flow with mutations, attachments, and shared GraphQL link registration; regressions could affect draft save/submit timing or other modules claiming the same operation names.
> 
> **Overview**
> Introduces a **returns** storefront module so signed-in buyers can list returns, start a request from an order, select quantities, edit a draft (reasons, comments, attachments), submit, view details, and cancel. When `Return.ReturnEnabled` is on, it registers account routes and menu entries, loads locales, and adds a **Request return** action on order details via the extension registry so the host only imports `@/modules/returns`.
> 
> The module ships its own GraphQL operations and composables (policy gating, URL-driven list filters, draft validation/submit, localized error handling). Draft edits autosave through **`UpdateReturn`**, which the module registers at init on the shared queued-mutations controller (debounced, partitioned by `returnId`) instead of hard-coding that mutation in core.
> 
> Core Apollo **queued mutations** now expose **`registerTarget`** with owner/priority rules (host wins), plus tests and a dependency-cruiser rule that blocks host imports past `modules/returns/index.ts`. **`app-runner`** calls `initReturns` alongside other modules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4c6bb181aa91f70e97bb3f12121cdf5d76e6030c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


client-app/modules/returns/composables/useReturnDraft.ts
```diff
@@ -0,0 +1,180 @@
...
+      (line) =>
+        !line.reasonCode ||
+        (requiresComment(line.reasonCode) && !line.reasonComment.trim()) ||
+        (attachmentsRequired.value && line.attachmentUrls.length === 0),
+    ),
+  );
+
+  const canSubmit = computed(() => submitAllowed.value && lines.value.length > 0 && incompleteLines.value.length === 0);
+
+  async function save(): Promise<boolean> {
```
ivan-kalachikov: ### Two autosaves can be in flight at once, and the newer edit loses

`save()` has no in-flight guard, `UpdateReturn` is not a `queuedMutationsLink` target (only the two cart mutations are, `queued-mutations.ts:321-326`), and Submit is not disabled while `saving` is true (`edit-return.vue:86`).

Each call PUTs the whole document, so **the last response to reach the server wins** — not the last edit.

**Repro (no submit needed, slow connection):**

1. Type in "serial number" → blur → 800 ms → request **A** starts
2. Type in "comment" → blur → 800 ms → request **B** starts while A is still in flight
3. A lands after B → server keeps A → **the comment is gone**
4. The status line said "Saved automatically" the whole time

**With submit:** `cancelAutosave()` clears the *timer*, not a request already on the wire. Change the reason, hit Submit inside A's window, and `submitReturn` commits the stale draft.

**Suggested fix** — the link already does exactly this:

```ts
createQueueTarget("UpdateReturn", {
  getPartitionKey: (vars) => vars.command.returnId,
})
```

Then drop the hand-rolled `setTimeout` and call `flushNow("UpdateReturn", returnId)` before submit. It never sends two requests in parallel per queue, and its docstring names this exact case — "on input blur, before navigation" (`queued-mutations/types.ts:8-21`).

Same change also fixes the `onScopeDispose` issue below.

client-app/modules/returns/composables/useReturnDraft.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  function autosave(): void {
+    cancelAutosave();
+    autosaveTimer = setTimeout(() => void save(), AUTOSAVE_DELAY);
+  }
+
+  // Leaving the page within the debounce window would otherwise run save - and report its error -
+  // against a scope that no longer exists.
+  onScopeDispose(cancelAutosave);
```
ivan-kalachikov: ### Navigating away silently drops the last edit

`cancelAutosave()` only does `clearTimeout` — nothing is flushed.

1. Type in a field → blur → 800 ms timer starts
2. Click the "Returns" breadcrumb at ~300 ms
3. Scope disposed → timer killed → **edit lost**

Reopening the draft shows the previous value. The status line read "Saved automatically" while that timer was pending.

Same for a reason just picked or an attachment that just finished, if navigation happens inside the debounce window.

`flushNow("UpdateReturn", returnId)` in a route-leave guard fixes this and the save-ordering issue together.

client-app/modules/returns/composables/useReturnErrors.ts
```diff
@@ -0,0 +1,25 @@
...
+    const graphQLErrors = (error as ApolloError | undefined)?.graphQLErrors;
+    const code = graphQLErrors?.[0]?.extensions?.code;
+
+    return typeof code === "string" && te(`returns.errors.${code}`) ? code : FALLBACK;
+  }
+
+  // The flow service fails with codes a buyer can act on — a line that lost its availability while
+  // the draft sat open, a missing photo. Swallowing them leaves a dead button and no explanation.
+  function report(error: unknown): void {
+    notifications.error({ text: t(`returns.errors.${getCode(error)}`), duration: 15000, single: true });
```
ivan-kalachikov: ### Every handled error fires two competing toasts

The returns mutations don't pass `SUPPRESS_ERROR_NOTIFICATIONS_CONTEXT`, so the global handler reacts as well:

```
RETURN_QUANTITY_UNAVAILABLE
 ├─ errorHandlerLink → toServerError() returns undefined for this code
 │                   → falls into error-handler.ts:33-36
 │                   → toast "Something went wrong" (broadcast.ts:109-124)
 └─ mutate() rejects → report() → your localized text
```

Both land in microtasks, so the order isn't controlled: the buyer can end up with the generic message **instead of** the actionable one — which is the whole point of the error catalogue this PR adds.

**Second issue:** `single: true` calls `clear()` with no group (`useNotifications.ts:21-23`), which wipes the **entire** notification stack, including unrelated toasts from other features.

**Fix:** pass `context: SUPPRESS_ERROR_NOTIFICATIONS_CONTEXT` on the returns mutations (as `sales-rep` and `getStorePlugins` already do) and drop `single: true`.

client-app/modules/returns/components/return-line-attachments.vue
```diff
@@ -0,0 +1,72 @@
...
+  props.attachments.map((attachment) => ({
+    status: "attached",
+    name: attachment.name,
+    url: attachment.url,
+    size: attachment.size,
+    contentType: attachment.mimeType ?? undefined,
+  })),
+);
+
+const { files, options, attachedAndUploadedFiles, addFiles, validateFiles, uploadFiles, removeFiles, fetchOptions } =
```
ivan-kalachikov: ### Submit can drop a file that is still uploading

This destructure takes `attachedAndUploadedFiles` but not `allFilesAttachedOrUploaded` or `hasFailedFiles`, and the only emit is `update:urls`. The parent has no way to know an upload is still in flight or has failed.

**Repro:**

1. A line has photo 1 uploaded and photo 2 still uploading
2. `attachmentUrls = [url_1]` → `incompleteLines` clears → Submit enables
3. Click Submit → `save()` sends only `url_1`, then submits
4. The return goes read-only → **photo 2 sits in the upload scope attached to nothing**, and the agent reviews the return without it

With attachments optional, even the first selected file can be lost the same way.

Existing precedent: `edit-quote.vue:233` gates saving on `allFilesAttachedOrUploaded`, and `company/info.vue:90` renders an alert from `hasFailedFiles`. Today a failed upload is visible only inside the uploader tile.

client-app/modules/returns/pages/return.vue
```diff
@@ -0,0 +1,193 @@
...
+        <template v-if="cancelAction" #footer>
+          <VcTooltip v-if="!cancelAction.isAvailable" placement="top">
+            <template #trigger>
+              <VcButton color="danger" variant="outline" size="sm" disabled>
+                {{ $t("return_details.cancel_button") }}
+              </VcButton>
+            </template>
+
+            <template #content>
+              {{ $t(`returns.action_unavailable.${cancelAction.unavailableReason}`) }}
```
ivan-kalachikov: ### Unknown server codes render as a raw i18n key

`returns.action_unavailable` has exactly one key: `WRONG_STATUS`. Any other code the server sends is printed literally in the tooltip:

> returns.action_unavailable.NOT_OWNER

`unavailableReason` is also `Maybe<String>` (`types.ts:128`), so `isAvailable: false` with no reason renders `returns.action_unavailable.undefined`.

The AC has at least one case that will not be `WRONG_STATUS`: *"A buyer cannot cancel a colleague's return"*.

Same pattern at `select-return-items.vue:55` for `returns.ineligibility` (6 keys). That catalogue lives in the backend repo, so it drifts on any module upgrade.

`useReturnErrors.getCode` already guards exactly this with `te()` before trusting a server code — these two templates don't.

client-app/modules/returns/pages/edit-return.vue
```diff
@@ -0,0 +1,168 @@
...
+          {{ saving ? $t("return_edit.saving") : $t("return_edit.saved_automatically") }}
+        </span>
+
+        <VcButton :disabled="!canSubmit" :loading="submitting" @click="onSubmit">
+          {{ $t("return_edit.submit") }}
+        </VcButton>
+      </div>
+
+      <p v-if="incompleteLines.length" class="mt-2 text-right text-sm text-warning-700">
+        {{ $t("return_edit.incomplete_hint", { count: incompleteLines.length }) }}
```
ivan-kalachikov: ### The hint blames the reason even when the reason is filled

`incompleteLines` flags a line for **three** different causes (`useReturnDraft.ts:74-78`):

```ts
!line.reasonCode                                        // no reason
requiresComment(line.reasonCode) && !line.reasonComment // no comment
attachmentsRequired && line.attachmentUrls.length === 0 // no attachment
```

But the text is fixed: `incomplete_hint` = `"{count} line(s) still need a reason"`.

So with every reason picked and one photo missing: Submit is disabled, and the only hint on screen points at the reason. The buyer has nowhere to look.

client-app/modules/returns/pages/returns.vue
```diff
@@ -0,0 +1,135 @@
...
+      <template #default-container>
+        <VcTable
+          :loading="loading"
+          :columns="columns"
+          :sort="sort"
+          :items="returns"
+          :pages="pages"
+          :page="page"
+          :description="$t('returns.meta.table_description')"
+          @item-click="goToReturn"
```
ivan-kalachikov: ### `@item-click` is not a VcTable event, so this listener never fires

VcTable emits `rowClick` (declared at `vc-table.vue:334`, emitted at 208, 1043, 1077). There is no `itemClick`.

Row navigation works only because this page supplies its own `#desktop-body` and `#mobile-item` slots with an explicit `@click="goToReturn(item)"`. If those slots are ever dropped in favour of VcTable's default body, navigation disappears silently.

Either bind `@row-click`, or remove the dead line.

(The same dead listener is in `quotes.vue:62` and `purchase-requests.vue:27` — copied, not introduced here.)

client-app/modules/returns/components/request-return-button.vue
```diff
@@ -0,0 +1,28 @@
...
+interface IProps {
+  orderId: string;
+}
+
+const props = defineProps<IProps>();
+
+const router = useRouter();
+
+// The same query the wizard opens with, so asking here costs the wizard nothing later.
+const { result } = useGetReturnableItemsQuery(computed(() => ({ orderId: props.orderId })));
```
ivan-kalachikov: ### One extra GraphQL round trip on every order page

This query runs whenever any order is opened — unpaid, cancelled, three years old. The answer is almost always "nothing returnable".

The same page already computes a status gate:

```ts
// order-details.vue:249
const showReorderButton = computed(() => !!order.value && order.value.status === "Completed");
```

Gating the component the same way would skip the request entirely for orders nothing can be returned from.

Separately: `hasReturnableItems` is false until the response lands, so on a returnable order the button pops in after first paint and shifts Print/Reorder sideways.

client-app/modules/returns/pages/return.vue
```diff
@@ -0,0 +1,193 @@
...
+
+function openCancelModal(): void {
+  openModal({
+    component: CancelReturnModal,
+    props: {
+      returnId: props.returnId,
+      returnNumber: orderReturn.value?.number ?? "",
+
+      async onResult(): Promise<void> {
+        await refetch();
```
ivan-kalachikov: ### Cancel refetches GetReturn twice

`useCancelReturnMutation` already declares:

```ts
refetchQueries: [GetReturns, GetReturn, GetReturnableItems]
```

The GetReturn on this page is active, so Apollo refetches it as part of the mutation. Then `onResult` awaits `refetch()` — a second identical network request and a second render pass on every cancel.

Drop the `refetch()` here, or drop `GetReturn` from `refetchQueries`.

client-app/modules/returns/pages/edit-return.vue
```diff
@@ -0,0 +1,168 @@
...
+          <VcButton :disabled="!bulkReason" variant="outline" @click="applyReasonToAll(bulkReason)">
+            {{ $t("return_edit.apply") }}
+          </VcButton>
+        </div>
+      </VcWidget>
+
+      <VcWidget v-for="line in lines" :key="line.orderLineItemId" :title="line.name ?? ''" size="lg" class="mt-5">
+        <p class="mb-4 text-sm text-neutral-400">
+          {{ [line.sku, line.measureUnit].filter(Boolean).join(" · ") }} ·
+          {{ $t("return_edit.returning", { quantity: line.quantity }) }}
```
ivan-kalachikov: ### `RETURN_QUANTITY_UNAVAILABLE` is a dead end for the buyer

Quantity is editable **only in the wizard**. Here it is static text, so the one error a correctly filled draft can still hit has no way out.

**What happens:**

1. 10:00 — wizard says 260 available, buyer picks 240 → `createReturn` bakes 240 into the draft
2. 10:05 — a colleague in the same org submits a return for the same line → available drops to 100
3. 10:08 — buyer fills reasons, uploads photos, clicks Submit
4. Server re-validates in-transaction → `RETURN_QUANTITY_UNAVAILABLE`

**What the buyer sees:** one generic toast — *"Some items are no longer available to return in the quantity you asked for. Adjust the quantities and try again."*

| Buyer asks | Answer on screen |
|---|---|
| Which line? | not shown (a return can have 20) |
| How many are available now? | not shown |
| Where do I change it? | nowhere — no input on this page |

The server sends all of it, and `useReturnErrors` reads only `extensions.code`:

```
extensions: { orderLineItemId, requestedQuantity: 240, availableQuantity: 100 }
```

Only escape today: cancel the return and redo the wizard, losing every reason, comment and uploaded file.

**The AC asks for three things, dead-on:**

> *"The frontend renders this **per line** and lets the buyer **adjust** rather than **losing the draft**"*

None of the three holds right now.

**Smallest fix:** reuse `ReturnQuantityInput` (already written, `components/return-quantity-input.vue`) on this page with `max` from `availableQuantity`, and parse `extensions` in `useReturnErrors` to mark the offending line.

That is a real chunk of work on top of an already large PR — splitting it into its own work item is completely reasonable. Your call: fix here, or track it separately so this doesn't ship claiming an AC it doesn't meet yet.

client-app/modules/returns/pages/select-return-items.vue
```diff
@@ -0,0 +1,161 @@
...
+            <VcCheckbox :model-value="allSelected" :disabled="!returnableItems.length" @change="toggleAll">
+              {{ $t("returns.select_items.select_all") }}
+            </VcCheckbox>
+
+            <span class="text-sm text-neutral-400">
+              {{ $t("returns.select_items.eligible_count", { eligible: returnableItems.length, total: items.length }) }}
+            </span>
+          </div>
+
+          <table class="w-full">
```
ivan-kalachikov: ### Could you check how this table renders on a phone?

Static reading only — I never rendered it, so I may well be wrong here. Two things in the code made me want a screenshot at ~400px.

**1. No mobile layout, and nothing in the chain adds horizontal scroll**

This is a raw `<table class="w-full">`. VcTable's answer to narrow screens is a card layout via the `#mobile-item` slot (`vc-table.vue:9`, `mobileBreakpoint: "md"` by default) — `returns.vue` in this same module uses it.

I grepped for `overflow` across the whole chain and found nothing:

| Where | `overflow` |
|---|---|
| this file + `return.vue` | none |
| `vc-widget.vue`, incl. `__slot-container` | none |
| `vc-table.vue` | none |
| `shared/layout/`, `pages/account/` | none |

At ~400px one row has to fit: product name + SKU/unit line + ineligibility text + Ordered + Returnable + a number input, every cell with `p-5` (20px per side).

My expectation is it either squashes to unreadable or pushes the page into horizontal scroll — but that is a guess from the markup, not something I observed. **If it looks fine on a device, just say so and ignore this.**

Same question for the items table on `return.vue:72`.

**2. No `<caption>` — this one is certain from the code**

VcTable renders a `<caption>` from its `description` prop (`vc-table.vue:58-60`); screen readers use it to announce what the table is. A raw `<table>` has none, so both tables here are unlabelled.

Telling detail: `returns.select_items.table_description` is translated in all 13 locale files and used nowhere — its name matches the VcTable prop exactly, and `returns.vue:19` passes its own one. Looks like this page was meant to use VcTable as well.

client-app/modules/returns/composables/useReturnDraft.ts
basilkot: fixed

client-app/modules/returns/composables/useReturnErrors.ts
basilkot: Fixed, both parts.

client-app/modules/returns/composables/useReturnDraft.ts
basilkot: Fixed

client-app/modules/returns/components/return-line-attachments.vue
basilkot: Fixed. The uploader now reports both flags upward, not just the settled URLs

client-app/modules/returns/pages/return.vue
basilkot: Fixed

client-app/modules/returns/pages/edit-return.vue
basilkot: fixed

client-app/modules/returns/pages/returns.vue
basilkot: Fixed — bound to @row-click, and the dead line is gone.

client-app/modules/returns/composables/useReturnDraft.ts
basilkot: Fixed exactly as suggested — thanks, the repro was spot on.

`UpdateReturn` is now a queue target (`queued-mutations.ts`):

```ts
export const updateReturnConfig: IQueueTargetConfig<{ command?: { returnId?: string } }> = {
  debounceMs: 800,
  getPartitionKey: (vars) => vars.command?.returnId ?? "",
};
```

Typed structurally rather than against `UpdateReturnMutationVariables` so `core` keeps no import from `modules/returns`. The default `mergeQueued` is right here: each save PUTs the whole document, so the later payload should win wholesale.

The hand-rolled `setTimeout` and `cancelAutosave` are gone. `autosave()` is now just `save()`, and `flushAutosave()` is `flushNow("UpdateReturn", returnId)`, called from `onScopeDispose` and from `submit`. That closes the navigation case in the next comment too.

One thing the suggestion didn't cover, which Bugbot then caught on the first version: calling `flushNow` *before* `save()` re-enqueues the pre-submit save and makes it wait out the full debounce. Submit stays clickable for that second because the button keys its loading state off the mutation, which hasn't started yet — so a second click fired a second `submitReturn`. Now the save is enqueued first and the queue drained immediately:

```ts
const saved = save(true);
flushAutosave();
if (!(await saved)) { return false; }
```

plus a flag that flips synchronously on click, since the mutation's own `loading` turns true too late to guard the gap.

Commits: `bd1ece247` (queue target) and `7e9b10086` (submit ordering and the double-click guard). Queued-mutations and boot-order suites pass, 47/47.

client-app/modules/returns/composables/useReturnDraft.ts
basilkot: Fixed. `cancelAutosave` is gone entirely — there is no timer left to clear. What used to be the cancel hook is now a flush:

```ts
function flushAutosave(): void {
  queuedMutationsController.flushNow("UpdateReturn", toValue(returnId));
}

onScopeDispose(flushAutosave);
```

I went with `onScopeDispose` rather than a route-leave guard. It covers the same case and two more: the router reusing this component between two drafts, and any other reason the scope goes away. The mutation isn't tied to the component — `flushNow` hands it to the link synchronously, and the request outlives the unmount.

The in-flight case is handled by the link rather than by us: if a request is already on the wire, `flushNow` sets `flushRequested` and the settle handler drains the pending batch immediately, without a second debounce and without a parallel request.

One consequence worth naming: a save that fails *after* the buyer has left now raises its toast against a scope that no longer exists. That is deliberate and it works — `useNotifications` and `t` are resolved at setup and the store is global — but it does mean a failed last-moment edit is reported on whatever page they landed on. The alternative was keeping the edit silently dropped, which is what you filed.

Not covered, and I don't think it should be: closing the tab inside the debounce window. A route guard wouldn't catch that either, and `beforeunload` can't be relied on to complete a request.

Commit: `bd1ece247`.

client-app/modules/returns/composables/useReturnErrors.ts
basilkot: Fixed, both parts.

All four returns mutations — `createReturn`, `updateReturn`, `submitReturn`, `cancelReturn` — now pass `context: SUPPRESS_ERROR_NOTIFICATIONS_CONTEXT`, so `errorHandlerLink` skips the generic toast and the only message the buyer sees is the one from the catalogue. The race you traced is gone rather than reordered.

`single: true` is dropped. You're right about what it does — `clear()` with no group wipes the whole stack, so a returns failure could take an unrelated toast with it. `report()` is now just:

```ts
notifications.error({ text: t(`returns.errors.${getCode(error)}`), duration: 15000 });
```

Worth noting `updateReturn` is opted out too, not only the three the buyer triggers by hand. Autosave runs unattended, so a rejection there used to produce the generic toast with no context at all; now it goes through `report()` like the rest.

Commit: `4252d0e1e`.

client-app/modules/returns/components/return-line-attachments.vue
basilkot: Fixed. The uploader now reports both flags upward, not just the settled URLs:

```ts
watch(allFilesAttachedOrUploaded, (value) => emit("update:settled", value), { immediate: true });
watch(hasFailedFiles, (value) => emit("update:failed", value), { immediate: true });
```

`update:urls` is unchanged and still carries only settled URLs — that part was right, a URL-less file must not be claimed. What was missing was the parent's ability to tell "no files" from "files not done yet", which is exactly your step 2.

The page keeps the state per line, since the uploader is per line:

```ts
const uploadsPending = computed(() => Object.values(uploadState.value).some((state) => !state.settled));
const uploadsFailed  = computed(() => Object.values(uploadState.value).some((state) => state.failed));
```

Submit is gated on `!canSubmit || uploadsPending`, and both states get a line of text next to the button rather than living only inside the uploader tile — following `edit-quote.vue` for the gate and `company/info.vue` for surfacing the failure. New keys `return_edit.uploads_pending` and `return_edit.uploads_failed` in all 13 locales.

`immediate: true` on both watchers matters: a line with no files must report `settled: true` straight away, otherwise the gate would latch closed on a draft nobody has touched.

Bugbot then found the follow-up on this same map: it was never cleared when `returnId` changed, and the router reuses the page between drafts, so a stale "still uploading" flag kept Submit disabled on the next draft. Cleared on `returnId` now.

Commits: `bd1ece247` and `7e9b10086`.

client-app/modules/returns/pages/return.vue
basilkot: Fixed. `useReturnErrors` grew a sibling to `getCode`, so the guard you pointed at now covers both catalogues instead of just `returns.errors`:

```ts
function codeText(namespace: string, code?: string | null): string {
  const key = code ? `returns.${namespace}.${code}` : "";

  return t(te(key) ? key : `returns.${namespace}.${FALLBACK}`);
}
```

Both templates go through it — the tooltip on `return.vue` and the ineligibility line on `select-return-items.vue` (twice now, the page has a mobile card layout since your other comment).

The `code ? … : ""` handles the `Maybe<String>` case you called out: a null `unavailableReason` no longer produces `returns.action_unavailable.undefined`, it lands on the fallback like any unknown code.

Added `UNHANDLED` to both namespaces in all 13 locales — "This action is not available right now" and "This item cannot be returned". Deliberately vague: when the server sends a code we've never seen, saying less is more honest than guessing at a reason.

On your point about drift — the catalogues living in the module repo is exactly why this is a fallback rather than an exhaustive list. `NOT_OWNER` from the AC case will render the generic line until we add the key; it just won't render the key itself at the buyer.

Commit: `4252d0e1e`.

client-app/modules/returns/pages/edit-return.vue
basilkot: Fixed. The three causes are counted separately now instead of being collapsed into one number:

```ts
const incompleteCounts = computed(() => ({
  reason: missingReason.value.length,
  comment: missingComment.value.length,
  attachment: missingAttachment.value.length,
}));
```

and the page renders one line per cause that actually has a count:

```vue
<template v-for="(count, cause) in incompleteCounts" :key="cause">
  <p v-if="count" …>{{ $t(`return_edit.incomplete.${cause}`, { count }) }}</p>
</template>
```

`incomplete_hint` is replaced by `return_edit.incomplete.{reason,comment,attachment}` in all 13 locales. Your wording is kept verbatim as the `reason` one.

Three separate sentences rather than one with an interpolated list, because joining translated fragments breaks in languages with cases — and the counts genuinely differ per cause, so "3 lines need a reason, 1 needs a photo" is more useful than a single total anyway.

`incompleteLines` still exists and still drives `canSubmit`, but it is now derived from the same three sets through a `Set`, so a line missing two things is counted once for the gate and once per cause for the hint.

Commit: `4252d0e1e`.

client-app/modules/returns/pages/returns.vue
basilkot: Fixed — bound to `@row-click`, and the dead line is gone.

This landed as part of a larger change to the same table, so it's worth saying what else moved: the page was still on the legacy `columns` array plus a hand-rolled `#desktop-body` with raw `<tr>`s. It's on `VcTableColumn` now, which is what made your point load-bearing rather than cosmetic — the explicit `@click` you identified as the only reason navigation worked went away with those rows, so `@row-click` is now doing the actual work.

`#mobile-item` keeps its own `@click`, since a card isn't a row.

The two copies you flagged, `quotes.vue:62` and `purchase-requests.vue:27`, are still there. Left them alone deliberately — both have the same `#desktop-body` structure this page used to have, so they're dead-but-harmless in place, and fixing them properly means the same table migration in two unrelated modules. Happy to file it if you'd rather it didn't sit.

Commit: `95d02f0df`.

client-app/modules/returns/components/request-return-button.vue
basilkot: Fixed, though not quite the way you suggested — and the difference matters.

Gating on `status === "Completed"` would be wrong here: which statuses a return may be raised from is a per-store setting, `Return.AllowedOrderStatuses`. Hardcoding one status would hide the button on any stand configured differently, and silently.

So the module now publishes it. `returnPolicy` gained `allowedOrderStatuses`, and the component reads that:

```ts
const eligible = computed(() => allowsOrderStatus(props.orderStatus));

const { result, loading } = useGetReturnableItemsQuery(
  computed(() => ({ orderId: props.orderId })),
  eligible,
);
```

The query is `enabled`-gated, so on an ineligible order no request goes out at all. `returnPolicy` itself is `cache-first` and already fetched by the module, so the gate costs nothing. `order-details.vue` passes `:order-status="order.status"`.

On the layout shift: the button now holds its place while the answer is in flight — rendered, disabled, `:loading`, with a `min-w` so it doesn't resize when the label replaces the spinner, same trick as the Reorder button next to it. It still disappears if nothing turns out to be returnable, but with the status gate that case is now rare rather than the default.

Commits: `14c3049` in vc-module-return (the policy field), `f4025d84f` here.

client-app/modules/returns/pages/return.vue
basilkot: Fixed — dropped the `refetch()` here rather than `GetReturn` from `refetchQueries`.

The mutation's list is the right place for it: cancel also has to update the returns list and the returnable quantities, and `GetReturn` belongs in the same set. Keeping it there means any other caller of `useCancelReturnMutation` gets the refresh too, whereas the `onResult` hook only ever covered this one page.

`refetch` is no longer destructured from `useReturn` on this page at all, so there's nothing left to call by accident.

Commit: `4252d0e1e`.

client-app/modules/returns/pages/edit-return.vue
basilkot: Done here rather than split out — you're right that it can't ship claiming the AC otherwise.

All three parts of the AC now hold:

**Adjust.** `ReturnQuantityInput` is on the edit page, one per line, with the ceiling read from the server:

```ts
const { result: returnableResult, refetch: refetchReturnable } = useGetReturnableItemsQuery(
  computed(() => ({ orderId: orderReturn.value?.orderId ?? "" })),
  computed(() => !!orderReturn.value?.orderId),
);

function maxQuantity(line: ReturnDraftLineType): number {
  return maxQuantities.value[line.orderLineItemId] ?? line.quantity;
}
```

A draft holds no quantity server-side, so what `returnableItems` reports already includes what this draft is asking for — it is the real ceiling, not the ceiling minus ourselves. Next to the input the page shows `of {quantity} available`.

**Per line.** `useReturnErrors` reads the extensions now, not just the code:

```ts
export function getReturnErrorDetails(error: unknown): ReturnErrorDetailsType
```

`submit` stores `orderLineItemId` from the refusal and the page renders an alert on that widget, so in a twenty-line return the buyer is pointed at the one that failed.

**Without losing the draft.** Nothing is cancelled or navigated away from. On a refusal the composable re-reads `returnableItems` before reporting, so by the time the buyer looks at the input, its max is the number the server will now accept rather than the stale one they were shown ten minutes earlier.

Editing a quantity clears the marker on that line, so the alert doesn't linger after it's been addressed.

Commit: `e49747fb7`.

client-app/modules/returns/pages/select-return-items.vue
basilkot: You read it right, and I didn't argue with it — both tables are on `VcTable` now.

**2 first, since it was the certain one.** `returns.select_items.table_description` is passed as the `description` prop, so the caption exists. `return.vue` needed a new key for its own table — `return_details.table_description`, added in all 13 locales.

**1.** Rather than measure whether the raw table squashed or scrolled, I moved both to `VcTable` with `mobile-breakpoint="lg"` and a `#mobile-item` card, same as `returns.vue` in this module. So below `lg` neither page renders a table row at all: the wizard card stacks name / SKU / ineligibility, then Ordered, Returnable and the quantity input as label-value rows; the details card does the same with Requested and Approved. The shared part of the item cell came out into `ReturnItemSummary` so the card and the row can't drift.

The column-per-`VcTableColumn` rewrite also removed the hand-rolled `<tr>` markup, which is what your `@item-click` comment was really about — the two land together.

To be straight about the limit of this: nobody has rendered either page at 400px yet, so I can't claim the cards look good, only that there is no longer a table to overflow. I'll attach screenshots.

Commit: `e49747fb7`.

basilkot: image
image

ivan-kalachikov (dismissed):

client-app/core/api/graphql/config/links/queued-mutations/queued-mutations.ts
```diff
@@ -318,10 +318,18 @@ export function createQueueTarget<TVars extends Record<string, unknown>>(
...
+export const updateReturnConfig: IQueueTargetConfig<{ command?: { returnId?: string } }> = {
+  debounceMs: 800,
+  getPartitionKey: (vars) => vars.command?.returnId ?? "",
+};
+
 export const queuedMutationsController = createQueuedMutationsController({
   targets: [
     createQueueTarget("UpdateShortCartItemQuantity", updateShortCartItemQuantityConfig),
     createQueueTarget("RemoveCartItems", removeCartItemsConfig),
+    createQueueTarget("UpdateReturn", updateReturnConfig),
```
ivan-kalachikov: ### `core` now names a module's mutation — the first time it names any module

```
grep -rn "modules/" client-app/core   →  0 hits before this PR
```

`"UpdateReturn"` is the first. The structural typing does keep the *import* out, but the dependency is still there — it just moved into a string literal, where no linter can see it and a rename in the module breaks it silently.

**This one is on my earlier comment.** I recommended `queuedMutationsLink` without checking it had an extension point. It doesn't: `targets` is a `new Set(...)` built once at construction (line 43), and the controller exposes only `link` + `flushNow`. Hardcoding the name in core was the only way to use it as written.

**The repo already has the right shape for this**, twice, and both are called from `sales-rep/index.ts`:

```ts
// core/api/graphql/config/registerCacheTypePolicies.ts — exported from core-api/index.ts:69
registerCacheTypePolicies(layoutTypePolicies, { owner: "sales-rep" });

// same idea, different seam
useWishlistSharingScopes().registerSharingScope({ scope: CUSTOMER_SHARING_SCOPE, ... });
```

Core hands out a registration function, the module calls it in `init()`, core stays ignorant. Applied here:

```ts
queuedMutationsController.registerTarget(
  createQueueTarget("UpdateReturn", {
    debounceMs: 800,
    getPartitionKey: (vars) => vars.command?.returnId ?? "",
  }),
  { owner: "returns" },
);
```

Late registration is safe: the link checks `targets.has(operation.operationName)` per request (line 212), not at build time, and `init()` runs at boot while the first `updateReturn` needs a navigation first.

Worth copying one more thing from `registerCacheTypePolicies`: it resolves owner collisions (`owner` + `priority`, with a refused list for debugging). A queue registry needs at least enough to stop a second module silently replacing the config for the same operation name.

If you'd rather not open core in this PR, leaving it with a comment saying it's deliberate is defensible — but it deserves a follow-up either way, because as it stands it reads as precedent for the next module.

client-app/pages/account/order-details.vue
```diff
@@ -181,14 +183,17 @@ import { breakpointsTailwind, useBreakpoints } from "@vueuse/core";
...
 import { OrderStatusCode } from "@/core/constants/order-status.ts";
+import { MODULE_ID as RETURNS_MODULE_ID, ENABLED_KEY as RETURNS_ENABLED_KEY } from "@/modules/returns/constants";
 import { useUserOrder, OrderLineItems, OrderStatus } from "@/shared/account";
 import { getItemsForAddBulkItemsToCartResultsModal, useShortCart } from "@/shared/cart";
 import { AcceptedGifts, OrderCommentSection, OrderSummary } from "@/shared/checkout";
 import { BOPIS_CODE } from "@/shared/checkout/composables/useBopis.ts";
 import { AddressInfo, VendorName } from "@/shared/common";
 import { BackButtonInHeader } from "@/shared/layout";
 import { useModal } from "@/shared/modal";
+import RequestReturnButton from "@/modules/returns/components/request-return-button.vue";
```
ivan-kalachikov: ### The host reaches past the module's `index.ts` — which is an `error` for `sales-rep`

Two imports here go into the module's internals rather than its entry point:

```ts
import { MODULE_ID, ENABLED_KEY } from "@/modules/returns/constants";   // :188
import RequestReturnButton from "@/modules/returns/components/...";     // :196
```

`.dependency-cruiser.cjs:218` already forbids exactly this shape for another module:

```js
{
  name: "no-host-to-sales-rep-internals",
  severity: "error",
  comment: "The host reached past 'modules/sales-rep/index.ts'. That module is being
            extracted into its own repository, so its entry point is the only surface
            the host may depend on.
            Widen this rule to the next module as it is prepared for extraction.",
  from: { pathNot: "^client-app/modules/sales-rep/" },
  to:   { path: "^client-app/modules/sales-rep/",
          pathNot: "^client-app/modules/sales-rep/index[.]ts$" },
}
```

`returns` has no such rule yet, so this passes CI in silence.

**How sales-rep does it.** Its `index.ts` is 160 lines and exports exactly one thing, `init`. Outside the module there are only two references in the whole repo — `app-runner.ts:41` and a test mock. It puts its own UI into host pages from the inside:

```ts
const { register } = useExtensionRegistry();
register("accountMenu", MY_CUSTOMERS_NAV_LINK_ID, {
  component: defineAsyncComponent(() => import("./components/link-my-customers.vue")),
});
```

The arrow is reversed: the module knows the host, the host knows nothing.

**Suggested shape here:**

```vue
<!-- order-details.vue -->
<ExtensionPoint category="orderDetails" name="actions" :order="order" />
```

```ts
// returns/index.ts, inside init()
register("orderDetails", EXTENSION_NAMES.orderDetails.actions, {
  component: defineAsyncComponent(() => import("./components/request-return-button.vue")),
});
```

What that buys beyond the boundary:

- **The `returnsEnabled` gate disappears.** `init()` already runs only when the module is on, so this page stops carrying return rules at all.
- **`defineAsyncComponent` is the idiom there**, which also closes the static-import point from my earlier review — today the button and its query composable land in the order-details chunk on every storefront, enabled or not.
- `condition` is supported, and two categories already type it against `CustomerOrderType` (`paymentPage`, `orderPaymentPage`), so a status check can live in the entry.
- Props reach the component: `ExtensionPoint.vue` does `v-bind="{ ...getProps(category, name), ...$attrs }"`.

**Cost, stated plainly:** there is no `orderDetails` category yet and no extension point on this page, so it needs the four steps in `shared/common/composables/extensionRegistry/README.md` — category type, initial registry entry, `EXTENSION_NAMES`, and the tag itself. That is core work, but reusable infrastructure rather than core learning about `returns`.

I first considered re-exporting the component from `returns/index.ts` instead. That is worse: the dependency arrow is unchanged and only the path becomes one the rule allows — the rule goes green while the coupling stays.

Once the imports are gone, copying the `sales-rep` rule with the path swapped keeps it that way.

client-app/core/api/graphql/config/links/queued-mutations/queued-mutations.ts
basilkot: Agreed, and thanks for owning the earlier suggestion — but the string literal in core was mine to notice, not yours to pre-empt.

The controller hands out a registration function now, shaped like `registerCacheTypePolicies`:

```ts
registerTarget: (target: IQueueTarget, owner: IQueueTargetOwner) => boolean;
debug: IQueueTargetsDebug;
```

and the module claims its own name in `init()`:

```ts
queuedMutationsController.registerTarget(
  createQueueTarget<{ command?: { returnId?: string } }>("UpdateReturn", {
    debounceMs: 800,
    getPartitionKey: (vars) => vars.command?.returnId ?? "",
  }),
  { owner: "returns" },
);
```

`grep -rn "modules/" client-app/core` is back to 0 hits.

I took the collision handling you pointed at rather than a simpler first-wins: `owner` + optional `priority`, the host seeded at 100, plugin claims capped at 99 so escalation stays available plugin-vs-plugin but never reaches the host, and a `rejected` list for debugging. A refused claim is logged and returns `false` rather than throwing — a losing module should degrade to unqueued mutations, not fail to boot.

Six tests: late registration queues, an unclaimed operation is left alone, the host cannot be displaced even with `priority: 9999`, a second plugin on the same name is refused and the holder recorded, and a higher-priority plugin does take over.

Commit: `27d9d5d62`.

client-app/pages/account/order-details.vue
basilkot: Done as suggested, including the four setup steps.

- `extensionRegistryMap.ts` — `orderDetails: ExtensionEntryType<{ order?: CustomerOrderType }, never, (order?: CustomerOrderType) => boolean>`, so `condition` is typed against the order like `paymentPage` is.
- `initialExtensionRegistry.ts` and `EXTENSION_NAMES.orderDetails.actions`.
- `order-details.vue` renders `<ExtensionPoint category="orderDetails" :name="EXTENSION_NAMES.orderDetails.actions" :order="order" />`.
- `returns/index.ts` registers into it with `defineAsyncComponent`.

Both imports are gone, and so is the `returnsEnabled` gate and the `useModuleSettings` call behind it — `init()` already runs only when the module is on, exactly as you said. The button now takes `order` rather than `orderId` + `orderStatus`, since that is what the point passes.

`no-host-to-returns-internals` added with the path swapped. `depcruise` reports **0 errors** (21 pre-existing warnings elsewhere).

Two things worth confirming back to you:

You were right that re-exporting from `index.ts` would have been worse — I did not try it. The arrow is genuinely reversed now: nothing outside the module references it except `app-runner.ts`.

And the eager-import point from the earlier round closes with this: the button and `useGetReturnableItemsQuery` no longer land in the order-details chunk on storefronts that have returns switched off.

What I have not done is render it — the extension point is verified by types and the dependency rule, not by eyes. I'll check it on the stand.

Commit: `5f8b3b6db`.

ivan-kalachikov (dismissed):

client-app/modules/returns/composables/useReturnDraft.ts
```diff
@@ -0,0 +1,247 @@
...
+    reason: missingReason.value.length,
+    comment: missingComment.value.length,
+    attachment: missingAttachment.value.length,
+  }));
+
+  const incompleteLines = computed(() => [
+    ...new Set([...missingReason.value, ...missingComment.value, ...missingAttachment.value]),
+  ]);
+
+  const canSubmit = computed(() => submitAllowed.value && lines.value.length > 0 && incompleteLines.value.length === 0);
```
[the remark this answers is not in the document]
basilkot: Already fixed in 3106e1caa, which landed two commits after the d1364ef8 this review ran against.

`missingQuantity` filters lines with `quantity < 1` and feeds both `incompleteCounts` and `incompleteLines`, so `canSubmit` is false while any line sits at zero — including the transient zero the field commits while a buyer clears it to retype.

The diagnosis was right and the sequence you described is the one that reaches the server, so thanks for it.

client-app/modules/returns/pages/edit-return.vue
```diff
@@ -0,0 +1,233 @@
...
+
+watch(
+  () => props.returnId,
+  () => {
+    uploadState.value = {};
+  },
+);
+
+const uploadsPending = computed(() => Object.values(uploadState.value).some((state) => !state.settled));
+const uploadsFailed = computed(() => Object.values(uploadState.value).some((state) => state.failed));
```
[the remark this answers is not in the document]
basilkot: Already fixed in 3106e1caa, after the reviewed commit.

`uploadsPending` is now `!state.settled && !state.failed`, with `uploadsFailed` tracked separately. A failed upload shows only the failure warning, and Submit stays disabled through `uploadsFailed` rather than by pretending a transfer is still running.

client-app/modules/returns/composables/useReturnableItems.ts
```diff
@@ -0,0 +1,96 @@
...
+      return undefined;
+    }
+
+    try {
+      return await createDraftInternal();
+    } catch (error) {
+      report(error);
+
+      return undefined;
+    }
```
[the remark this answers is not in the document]
basilkot: Already fixed in 3106e1caa, after the reviewed commit.

`createDraft` wraps `createDraftInternal` and awaits `refetch()` before reporting, so the ceilings are re-read whenever the create is rejected — the same thing submit already did.

client-app/modules/returns/pages/edit-return.vue
```diff
@@ -0,0 +1,238 @@
...
+            :label="$t('return_edit.apply_to_all', { count: lines.length })"
+          />
+
+          <VcButton :disabled="!bulkReason" variant="outline" @click="applyReasonToAll(bulkReason)">
+            {{ $t("return_edit.apply") }}
+          </VcButton>
+        </div>
+      </VcWidget>
+
+      <VcWidget v-for="line in lines" :key="line.orderLineItemId" :title="line.name ?? ''" size="lg" class="mt-5">
```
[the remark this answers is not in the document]
basilkot: Real, and the mechanism is exactly as described. Fixed in 0b7fbeca.

Verified both links rather than taking the chain on trust. `useFiles` seeds its list at `useFiles.ts:81` under `files.value.length === 0`, and the pass after it only refreshes entries it can match by `name`, so nothing removes the previous draft's files. And this page is the routed component for `:returnId/edit`, so vue-router reuses the instance across drafts — which is why the `uploadState` reset existed in the first place.

Worth adding that the two halves interact the way you say. Because the uploader is reused rather than remounted, its `immediate` watchers never re-run, so clearing `uploadState` on `returnId` did not just lose a stale flag — it lost the real one, and Submit unblocked with a transfer still in flight on that line.

Fix is to key each line widget by draft and line rather than line alone, so the subtree is rebuilt and the children re-report themselves. The `uploadState` reset stays, since a line the next draft does not carry would otherwise leave an entry behind.

One note on blast radius, for the record rather than as mitigation: the module PR now refuses a file already owned by another return, so the autosave you describe would be rejected with `ATTACHMENT_NOT_AVAILABLE` rather than silently attaching another draft's photo. That turned a data bug into a confusing error, and this is the fix for the actual cause.

Not covered by a test — the returns module has no frontend unit tests yet, which is a gap I have flagged separately.

PR #2501, opened 2026-09-22
feat(VCST-2945): barcode search setup

yuskithedeveloper:
## Description

Adds the storefront half of **VCST-2945 — barcode scanner search setup**. The existing barcode scanner (VCST-2622) now follows two store settings owned by the Catalog module: whether the scanner button is shown at all, and how a scanned code is matched — the old **full-text search**, or an **exact match** on the fields the store configured (SKU, GTIN, MPN or short-text properties), which also finds variations and opens the product directly on a single hit.

Requires the backend PRs:
- https://github.com/VirtoCommerce/vc-module-catalog/pull/909 — the settings `Catalog.Search.BarcodeScannerEnabled` / `Catalog.Search.BarcodeSearchFields` and the admin blade that edits them
- https://github.com/VirtoCommerce/vc-module-x-catalog/pull/113 — expands the `barcode:"<value>"` filter term into the configured fields

On an older backend both settings are absent: the scanner stays enabled and behaves exactly as today.

## Behaviour

- **Settings** (`shared/layout/composables/useBarcodeSearch.ts`, keys in `core/constants/modules.ts`): read from module `VirtoCommerce.Catalog` through `useModuleSettings`. `BarcodeScannerEnabled === false` hides the scanner in both search bars (missing ⇒ enabled). `BarcodeSearchFields` is the JSON array the blade saves; a non-empty array switches scans to exact mode, anything else (missing, `[]`, unparsable) keeps full text.

- **Exact mode**: a scan navigates to the search page with **`?barcode=<value>`**; `category.vue` turns it into the filter term `barcode:"<escaped value>"` (the existing `escapeFilterSyntaxValue`, so QR payloads with quotes or colons stay one term) and leaves the keyword empty — the backend expands the term into the configured fields. **Full-text mode is unchanged**: the value is submitted as the search phrase (`?q=`).

- **A scan is an identity lookup, not browsing**: a barcode request ignores `?facets=` and the in-stock, purchased-before and branch preferences (the shopper is holding the item; the sample store's variation is out of stock and would otherwise be unfindable), so the sidebar, browsing controls and control chips stay hidden while a barcode is active — no control claims a filter that is not applied. Paging and sorting still work.

- **A single hit opens the product** (`shouldOpenSingleBarcodeHit`, a pure function with its own tests): the redirect is bound to the request that produced the result — the code captured when the request was issued must still be the current one, the request must not have been narrowed by facets/page/sort, exactly one product with total 1, and at most once per code — so a slower earlier response or a later narrowing of a multi-hit page can never redirect. `router.replace`, so Back does not loop. A variation hit opens the variation's own page (`/product/<id>`; variations have no slug).

- **Headings and empty state**: *Your search for barcode {barcode} returned the following* / *No products found for barcode {barcode}*, with the search-only empty view and a Reset that clears `barcode` (and `q`), so a miss is never a dead end.

- **Generated filters**: the `filter` fragment now selects `isGenerated` and `useProducts.prepareFilters` drops generated filters — the backend reports the barcode expansion that way, like `is:product`. `barcode` is additionally listed in `EXCLUDED_FILTER_NAMES` so an older backend that echoes it as a user filter still shows no chip. `types.ts` was patched by hand for the new field (codegen is not runnable against a source-linked backend) and checked against the schema.

- **Both bars, one handler**: `search-bar.vue` and `mobile-search-bar.vue` delegate to the composable; the mobile overlay/dropdown is hidden before the push through the same mechanism the full-text path uses. Scans fire the `search` analytics event, are deliberately **not** written to search history, and stay **global** (no category scope).

- **Tests**: `useBarcodeSearch.test.ts` (17 — settings parsing, route building, hide-before-push order, analytics, the redirect decision), scanner gating in `search-bar.test.ts` and the new `mobile-search-bar.test.ts`, escaping in `facets.test.ts`, the generated-filter drop in `useProducts.test.ts`. Full vitest run green (3123); eslint 0 errors in touched files; vue-tsc clean.

- **i18n**: `pages.search.header_barcode` / `header_barcode_empty` in all 13 locales.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-2945
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/vc-theme-b2b-vue-2.59.0-pr-2501-fe26-fe26e3c6.zip


client-app/shared/catalog/composables/useProducts.ts
```diff
@@ -553,7 +553,9 @@ export function useProducts(
   }
 
   function prepareFilters(filters: SearchProductFilterResult[]) {
-    return filters.filter((filter) => !isZeroPriceFilter(filter) && !isExcludedFilter(filter));
+    // Server-generated filters (`is:product`, the barcode expansion, …) are not the user's choice:
+    // they must not become chips, count as applied filters, or be echoed back as facet filters.
+    return filters.filter((filter) => !filter.isGenerated && !isZeroPriceFilter(filter) && !isExcludedFilter(filter));
```
ivan-kalachikov, line 558: **This hides intent-search filters in every search, not only barcode lookups.**

`vc-module-intent-search` marks the filters it infers with `isGenerated: true` (facets, exact IDs, category outline; see `SearchIntentAnalysisMiddleware.cs:122,133,141`).

Example: the shopper searches "red nike sneakers" and the AI adds `color=red` and `brand=Nike`.
- Before: both show as chips, and removing one turns on `preserveUserQuery`.
- After: the results are narrowed with no chip and no Reset. The first facet click then drops those filters without telling the shopper.

Also, `is:product` isn't generated. x-catalog adds it without the flag, and `EXCLUDED_FILTER_NAMES` already hides it by name.

Suggestion: drop generated filters only while a barcode lookup is active.

client-app/shared/catalog/components/category.vue
```diff
@@ -421,24 +440,38 @@ const categoryListProperties = computed(() => ({
   related_type: "category",
 }));
 
+// A scanned code arrives in its own query param and is matched by the API against the product index
+// fields configured for the store, so it filters instead of becoming a search keyword.
+const barcode = useRouteQueryParam<string>(QueryParamName.Barcode, { defaultValue: "" });
```
ivan-kalachikov, line 445: **Load-more mode: a second scan reuses the old page number.**

1. Scan code A (many hits) and load more up to page 3.
2. Scan code B. The URL becomes `/search?barcode=B` with no `page`.
3. The component is reused, so `currentPage` is still 3.
4. `useProducts.fetchProducts` requests B at page 3. Its single hit is on page 1, so the response has `totalCount: 1` and no items. The shopper sees an empty result and no redirect.

Only `searchQueryParam` has a watcher that resets the page. The "Reset page" chip is hidden in barcode mode, so the shopper can't recover.

Suggestion: reset the page (and `pageHistory`) when `barcode` changes.

client-app/shared/catalog/components/category.vue
```diff
@@ -421,24 +440,38 @@ const categoryListProperties = computed(() => ({
...
 const filteredOnlyBySearch = computed(() => {
-  return !hasSelectedFilters.value && !!searchQueryParam.value;
+  return !hasSelectedFilters.value && (!!searchQueryParam.value || isBarcodeLookup.value);
 });
 const emptyViewSearchOnly = computed(() => {
   return filteredOnlyBySearch.value && products.value.length === 0 && !fetchingProducts.value;
 });
 const hideAllControls = computed(() => {
-  return emptyViewSearchOnly.value;
+  return emptyViewSearchOnly.value || isBarcodeLookup.value;
```
ivan-kalachikov, line 459: **Sorting and the grid/list switch are hidden during a barcode lookup.**

`hideAllControls` hides the whole controls block. When a code matches several products (a shared GTIN, or several fields matched with OR), the shopper can't sort or change the view.

The PR description says "paging and sorting still work". Sorting still works through the URL, but not from the UI.

client-app/shared/catalog/components/category.vue
```diff
@@ -512,13 +556,19 @@ const searchParams = computed<ProductsSearchParamsType>(() => ({
   itemsPerPage: props.fixedProductsCount || itemsPerPage.value,
   sort: sortQueryParam.value,
   keyword: searchQueryParam.value || props.keyword,
```
ivan-kalachikov, line 558: **`?barcode=` and `?q=` together: the keyword still narrows an exact lookup.**

Example: `/search?barcode=123&q=hat`
- A valid code can return nothing.
- Or `q` cuts several hits down to one, and the page redirects as if the code were unique. `q` isn't in `NARROWING_QUERY_PARAMS`.

Only a hand-written or shared URL can do this, but it's an easy fix: send no keyword while `isBarcodeLookup` is true.

client-app/core/utilities/search/facets.ts
```diff
@@ -89,6 +89,16 @@ export function getFilterExpressionForBrand(brandName?: MaybeRef<string>): strin
...
 }
 
+/**
+ * `barcode` is a virtual filter name: the API expands it into the product index fields configured
+ * for the store (`Catalog.Search.BarcodeSearchFields`).
+ * A scanned payload is arbitrary text, so it is escaped to keep quotes and backslashes inside the term.
+ */
+export function getFilterExpressionForBarcode(value?: MaybeRef<string>): string {
+  const barcode = unref(value);
+  return barcode ? `barcode:"${escapeFilterSyntaxValue(barcode)}"` : "";
```
ivan-kalachikov, line 99: **`?barcode=a&barcode=b` crashes the search page.**

vue-router returns a repeated param as an array. `escapeFilterSyntaxValue` then throws `value.replaceAll is not a function` inside the `searchParams` computed.

Suggestion: normalize it where it's read (take the first value, or ignore anything that isn't a string).

client-app/shared/catalog/components/category.vue
```diff
@@ -565,6 +623,29 @@ async function fetchProducts(): Promise<void> {
...
+    redirectedBarcodes,
+    totalCount: totalProductsCount.value,
+    itemCount: products.value.length,
+  });
+
+  if (!canOpen) {
+    return false;
+  }
+
+  redirectedBarcodes.add(requestedBarcode);
```
ivan-kalachikov, line 640: **The redirect reads shared refs, not the response of its own request.**

`products` and `totalProductsCount` are shared, and `useProducts.fetchProducts` doesn't discard stale responses.
- Redirecting to the wrong product would need two responses to resolve almost together. That's unlikely.
- A late older response overwriting the list is a real race, but dev already has it.

So "bound to the request that produced the result" is only partly true. Low risk.

client-app/shared/layout/composables/useBarcodeSearch.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  function onBarcodeScanned(value: string): void {
+    if (!value) {
+      return;
+    }
+
+    if (hasBarcodeSearchFields(getSettingValue(CATALOG_BARCODE_SEARCH_FIELDS_KEY))) {
+      // Parity with the full-text path's `search` event. By design the code is NOT saved to the search
+      // history (it is not a keyword) and the lookup stays global (no category scope).
+      analytics("search", value);
```
ivan-kalachikov, line 81: **Barcode lookups miss part of the search analytics.**

- `viewSearchResults` never fires for barcode results, because both call sites in `category.vue` check `searchQueryParam`.
- This `search` event doesn't include the results count that the full-text path sends.

client-app/shared/catalog/components/category.vue
```diff
@@ -565,6 +623,29 @@ async function fetchProducts(): Promise<void> {
   }
 }
 
+function openSingleBarcodeHit(requestedBarcode: string, wasNarrowed: boolean): boolean {
```
ivan-kalachikov, line 626: **No test covers the page-level barcode behavior.**

Only `shouldOpenSingleBarcodeHit` is tested. Nothing tests:
- the single-hit redirect,
- that facets and controls are ignored,
- Reset clearing `barcode`,
- the page reset on a second scan (see the comment on line 445),
- `barcode` + `q` (see the comment on line 558).

client-app/shared/layout/composables/useBarcodeSearch.ts
```diff
@@ -0,0 +1,95 @@
...
+  return { name: ROUTES.SEARCH.NAME, query: { [QueryParamName.Barcode]: value } };
+}
+
+/** The setting holds a JSON array of index fields; an unreadable one must not switch to exact match. */
+function hasBarcodeSearchFields(rawValue: unknown): boolean {
+  return !!parseJsonStringArray(rawValue)?.length;
+}
+
+/**
+ * D4: a scanned code identifies one product, so a single hit opens it instead of a one-item list.
```
ivan-kalachikov, line 30: nit: `D4:` refers to a decision ID from the spec. Please drop it; the sentence reads fine without it.

client-app/core/api/graphql/fragments/filter.graphql
```diff
@@ -2,6 +2,7 @@ fragment filter on SearchProductFilterResult {
   name
   label
   filterType
+  isGenerated
```
ivan-kalachikov, line 5: nit: `types.ts` was edited by hand. The result looks right, and `isGenerated` is already in the schema, but the repo rule is codegen only (`yarn generate:graphql-types`). Please regenerate it against a backend.

